PR #2, opened 2022-05-12
Project create external

DavidFair:


DavidFair:
TODO:
- [ ] Port Router add Interface

actions/src/subnet_actions.py
```diff
@@ -0,0 +1,56 @@
...
+        return func(**kwargs)
+
+    def subnet_show(self, subnet):
+        """
+        Show subnet properties
+        :param subnet: Name or ID
+        :return: (status (Bool), reason/output (String/Dict))
+        """
+        try:
+            subnet = self.conn.network.find_subnet(subnet, ignore_missing=False)
```
Reillyhewitson, line 30: I don't see a conn variable being created here. How is it being accessed?

actions/router.delete.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 description: Delete Router
 enabled: true
```
Reillyhewitson, line 3: As this isn't implemented you might want to set this and router.update to false.

actions/src/subnet_actions.py
DavidFair: This is the legacy way of using the API, I'll add a `raise NotImplementError` to make it clear :+1:

Reillyhewitson (approved):

PR #3, opened 2022-05-26
Openstack check actions

Reillyhewitson:
These are the new check actions for stackstorm.
There are 4:
-  Security Group check - identifies security groups by ip-prefix and port ranges.
- Deleting Machines check - identifies machines that have been stuck in a deleting state for >10 mins.
- Loadbalancer check - identifies loadbalancers and amphorae that are inaccessible or errored.
- Old volume snapshots check - identifies snapshot volumes that were last updated in the previous month or earlier

Additionally there is a script to generate issues in any atlassian queue.


actions/checks.create.tickets.yaml
```diff
@@ -0,0 +1,30 @@
+description: Creates tickets on atlassian
+enabled: true
+entry_point: src/openstack_check_actions.py
+name: checks.create.tickets
```
DavidFair, line 4: Can we group these into `atlassian.create.tickets` instead of checks, as I can see other Atlassian based actions in the future

actions/checks.deleted.machines.yaml
```diff
@@ -0,0 +1,25 @@
...
+    default: ""
+    required: true
+    type: string
+  project_id:
+    description: ID of the project to be scanned for the above security group details (*Requires one of)
+    default: ""
+    required: false
+    type: string
+  all_projects:
+    description: Toggle to scan all projects in cloud (*Requires one of)
```
DavidFair, line 21: We can drop `Requires one of` since the GUI/API will enforce the bool type

actions/checks.loadbalancers.yaml
```diff
@@ -0,0 +1,15 @@
+description: Check for amphorae with an error status and find out if it is an issue with the amphora or loadbalancer
```
DavidFair, line 1: So this should be a sensor, since we will likely trigger some sort of reactive action when Amphorae / the Load balancer falls over:
https://docs.stackstorm.com/sensors.html

actions/checks.security.groups.yaml
```diff
@@ -0,0 +1,40 @@
...
+    default: security_groups_check
+    immutable: true
+    type: string
+  cloud_account:
+    description: The clouds.yaml account to use whilst performing this action
+    default: ""
+    required: true
+    type: string
+  ip_prefix:
+    description: The IPs that are allowed to access the port range specified below
```
DavidFair, line 16: Are these IPs comma separated?

actions/src/openstack_check_actions.py
```diff
@@ -0,0 +1,447 @@
...
+import logging
+import sys
+from datetime import datetime
+import requests
+from openstack_api.openstack_connection import OpenstackConnection
+from openstack_action import OpenstackAction
+from requests.auth import HTTPBasicAuth
+
+
+class CheckActions(OpenstackAction):
```
DavidFair, line 13: You want to inherit from `Action` rather than `OpenstackAction` (which is the legacy API)
This means we're using Stackstorm as-is and we can simply use their docs in the future too

actions/src/openstack_check_actions.py
```diff
@@ -0,0 +1,447 @@
...
+
+    # pylint: disable=useless-super-delegation
```
DavidFair, line 15: This `__init__` block can be deleted all-together. Python will automagically generate this so it's something to potentially go wrong in the future at the moment

actions/src/openstack_check_actions.py
```diff
@@ -0,0 +1,447 @@
...
+                if (
+                    rules["remote_ip_prefix"] == ip_prefix
+                    and rules["port_range_max"] == max_port
+                    and rules["port_range_min"] == min_port
+                ):
+                    ruledict = {}
+                    print("Issue with rule " + rules["security_group_id"])
+                    ruledict["id"] = rules["security_group_id"]
+                    ruledict["server_list"] = []
+
```
DavidFair, line 52: Can we further chunk this down into sub methods e.g.:
```
def check_project():
   with OpenstackConnection(cloud_name=cloud) as conn:
            sec_group = conn.list_security_groups(filters={"project_id": project})
   bad_rules = self.filter_rules_with_issues(rules)
   rules_with_issues = self.find_rules_in_use(bad_rules)
```

Rather than it all being in a single method? A nice way method to tell your method does to much is describe the body without using the words `and` or `or`. At the moment I would say
"This checks projects by getting security groups and filters bad rules and those used in projects"

Instead it would be good to get it to
"This checks projects for invalid security rules" (then the child method do the filtering / checking)

actions/src/openstack_check_actions.py
```diff
@@ -0,0 +1,447 @@
...
+    # pylint: disable=arguments-differ
+    def run(self, submodule: str, **kwargs):
+        """
+        Dynamically dispatches to the method wanted
+        """
+        func: Callable = getattr(self, submodule)
+        return func(**kwargs)
+
+    # pylint: disable=too-many-arguments, too-many-locals, no-self-use
+    def check_project(
```
DavidFair, line 29: If this is an internal method the Python notation is:
```
def _check_project
```

That indicates to future readers only this file should call it. Also `check_project` is a tad too generic, what specifically does it check for?

actions/src/openstack_check_actions.py
```diff
@@ -0,0 +1,447 @@
...
+        self, project: str, cloud: str, max_port: int, min_port: int, ip_prefix: str
```
DavidFair, line 30: Pylint is warning since `self` can be dropped. Instead we can simply do:
```
@static_method
def _check_project(cloud:str, max_port:int, ...):
```

actions/src/openstack_check_actions.py
```diff
@@ -0,0 +1,447 @@
...
+                output.append(
+                    {
+                        "dataTitle": {"id": str(i.id), "action": str(i["status"])},
+                        "dataBody": {"id": i["id"], "data": json.dumps(i)},
+                    }
+                )
+
+        return output
+
+    def loadbalancer_check(self, cloud_account: str):
```
DavidFair, line 189: Ditto on this method doing too much in a single body, can we break it down into sub methods?

actions/src/openstack_check_actions.py
```diff
@@ -0,0 +1,447 @@
...
+            if cloud_account == "dev":
+                amphorae = requests.get(
+                    "https://dev-openstack.nubes.rl.ac.uk:9876/v2/octavia/amphorae",
+                    headers={"X-Auth-Token": conn.auth_token},
+                )
+            else:
+                amphorae = requests.get(
+                    "https://openstack.nubes.rl.ac.uk:9876/v2/octavia/amphorae",
+                    headers={"X-Auth-Token": conn.auth_token()},
+                )
```
DavidFair, line 205: Can we move Amphorae handling into a separate file as we can reuse these chunks in future code (and it helps keep everything tidy). It could go into the `lib/` folder in a new `amphorae` module

Also this approach of checking for `dev` is quite brittle, but I can't think of a better approach.
I wonder if there's a way to get the Openstack API to effectively handle this from the URL in the clouds.yaml file :thinking:

DavidFair (changes requested):
(Partial Review, I haven't checked the workflow definitions yet)

Some general maintainability / reusability / general Python fixes and improvements
Overall though it looks like a really good chunk of work

actions/checks.deleted.machines.yaml
Reillyhewitson: This is more related too needing one of either project ID or a True for all projects. Is there a better way I could enforce that?

actions/checks.security.groups.yaml
Reillyhewitson: No, these are given as a CIDR e.g. 0.0.0.0/0 for all internet.
I'll make this clearer.

actions/src/openstack_check_actions.py
Reillyhewitson: I've taken out a little bit of this. This method ends up looking pretty unwieldy but most of it is just generating the different metadata for tickets for each variation, so I don't think I can split up any more than I have in the commit I just did.

actions/src/openstack_check_actions.py
Reillyhewitson: Missed this in the latest commit. I'll move it around now.

actions/checks.deleted.machines.yaml
DavidFair: Ah okay, I assumed it was a partial sentence. Now I understand it I think this is fine.

As long as we have a test, something like 

```
if bool(project_id) == all_projects:
    raise RuntimeError("A project ID or all projects must be enabled. Either both or neither were entered")
```

actions/src/openstack_check_actions.py
```diff
@@ -0,0 +1,463 @@
...
+        }
+
+        if all_projects:
+            with OpenstackConnection(cloud_name=cloud_account) as conn:
+                projects = conn.list_projects()
+            for project in projects:
+                deleted = self._check_deleted(
+                    project=project["id"], cloud=cloud_account
+                )
+            output["server_list"].extend(deleted)
```
DavidFair, line 167: There's a bug here, only the final iteration of the loop will be appended as it's not indented enough

actions/src/openstack_check_actions.py
```diff
@@ -0,0 +1,463 @@
...
+            sys.exit(1)
+        return output
+
+    @staticmethod
+    def _check_amphora_status(amphorae):
+        try:
+            amph_json = amphorae.json()
+            return amph_json
+        except requests.exceptions.JSONDecodeError:
+            logging.critical(
```
DavidFair, line 300: Do this as a string then return that:

```
msg = f"There was no JSON response \nThe status code was: {amphorae.status_code}. \n The body was: {amphorae.content}"
logging.critical(msg)
return msg
```

actions/src/openstack_check_actions.py
```diff
@@ -0,0 +1,463 @@
...
+
+        return ["error", status]
+
+    # pylint: disable=invalid-name
+    @staticmethod
+    def _ping_lb(ip):
+        # Runs the ping command to check that loadbalancer is up
+        response = os.system(
+            "ping -c 1 " + ip
+        )  # Might want to update to subprocess as this has been deprecated
```
DavidFair, line 330: Use https://pypi.org/project/pythonping/ instead of hand rolling our own

actions/src/openstack_check_actions.py
```diff
@@ -0,0 +1,463 @@
...
+                    )
+                )
+        else:
+            snapshots = self.check_snapshots(
+                project=project_id, cloud_account=cloud_account
+            )
+
+        for snapshot in snapshots:
+            with OpenstackConnection(cloud_name=cloud_account) as conn:
+                proj = conn.get_project(name_or_id=snapshot["project_id"])
```
DavidFair, line 366: Could we swap these, open the connection once then do a loop within it.
This way we don't hammer the auth server constantly opening new connections each time

lib/amphorae.py
```diff
@@ -0,0 +1,20 @@
+from openstack_api.openstack_connection import OpenstackConnection
+import requests
+
+
+def get_amphorae(cloud_account: str):
+    """
+    Method to get list of amphorae in openstack project
+    """
+    with OpenstackConnection(cloud_name=cloud_account) as conn:
+        if cloud_account == "dev":
```
DavidFair, line 10: This block can be simplified to
```python

DEV_URL = "https://dev-openstack.nubes.rl.ac.uk:9876/v2/octavia/amphorae"
PROD_URL = "https://openstack.nubes.rl.ac.uk:9876/v2/octavia/amphorae"

def get_amphorae(cloud_account: str):
    """
    Method to get list of amphorae in openstack project
    """
    url = DEV_URL if "dev".casefold() in cloud_account.casefold() else PROD_URL
    with OpenstackConnection(cloud_name=cloud_account) as conn:
         return requests.get(url, headers={"X-Auth-Token": conn.auth_token})

File: actions/src/openstack_check_actions.py [hunk 17]
```diff
@@ -0,0 +1,463 @@
...
+        output = {
+            "title": "{p[title_text]}",
+            "body": "The loadbalance ping test result was: {p[lb_status]}\nThe status of the amphora was: {p[amp_status]}\nThe amphora id is: {p[amp_id]}\nThe loadbalancer id is: {p[lb_id]}",
+            "server_list": [],
+        }
+        if amphorae.status_code == 403:
+            # Checks user has permissions to access api
+            logging.critical("Please run this script using a cloud admin account.")
+            return "Please run this script using a cloud admin account."
+        if amphorae.status_code == 200:
```
DavidFair, line 225: Simplify this by checking:
```
if amphorae.status_code != 200:
            # Notes problem with accessing api if anything other than 403 or 200 returned
            logging.critical("We encountered a problem accessing the API")
            logging.critical("The status code was: %s ", str(amphorae.status_code))
            logging.critical("The JSON response was: \n %s", str(amph_json))
            return False  # Note I've got rid of the sys.exit

for i in amph_json["amphorae"]:
# This does not need to be in an else block, as we have to have a 200 to get this far
```

actions/atlassian.create.tickets.yaml
```diff
@@ -0,0 +1,30 @@
...
+enabled: true
+entry_point: src/openstack_check_actions.py
+name: atlassian.create.tickets
+parameters:
+  submodule:
+    default: create_ticket
+    immutable: true
+    type: string
+  tickets_info:
+    description: Dictionary from which the script will generate tickets. Should be provided by an upstream action.
```
DavidFair, line 11: Dictionary of what? Name and ticket description I'm going to speculate

actions/src/openstack_check_actions.py
```diff
@@ -0,0 +1,277 @@
...
+            for serv_groups in applied:
+                if serv_groups["id"] == rules["security_group_id"]:
+                    print("Rule is applied")
+                    rules_with_issues.append(
+                        {
+                            "dataTitle": {"id": server["id"]},
+                            "dataBody": {
+                                "proj_id": project,
+                                "sec_id": serv_groups["id"],
+                                "applied": "Yes",
```
DavidFair, line 74: Can we pack this into a dataclass, since they all use the same layout in multiple places?

actions/src/openstack_check_actions.py
```diff
@@ -0,0 +1,277 @@
...
+        # print(servers.get_all_servers())
+        rules_with_issues = {
+            "title": "Server {p[id]} has an incorrectly configured server group",
+            "body": "Project id: {p[proj_id]}\nSecurity Group ID: {p[sec_id]}\n Is applied: {p[applied]}",
+            "server_list": [],
+        }
+
+        if all_projects:
+            with OpenstackConnection(cloud_name=cloud_account) as conn:
+                projects = conn.list_projects()
```
DavidFair, line 123: This can be simplified by treating a list as 1 or more objects (so a single project would be handled by the same code):
```python
if all_projects:
   with OpenstackConnection(...)
        projects = conn.list_projects()
else:
    projects = [project]

for project in projects:
  output = self._check_project_loadbalancers(
  project=project["id"],
  cloud=cloud_account,
  max_port=max_port,
  min_port=min_port,
  ip_prefix=ip_prefix,
                )
  rules_with_issues["server_list"].extend(output)

File: actions/src/openstack_check_actions.py [hunk 21]
```diff
@@ -0,0 +1,277 @@
...
+                            p=i["dataTitle"]
+                        ),
+                        "description": actual_tickets_info["body"].format(
+                            p=i["dataBody"]
+                        ),
+                    },
+                    "serviceDeskId": servicedesk_id,  # Point this at relevant service desk
+                    "requestTypeId": requesttype_id,
+                },
+            )
```
DavidFair, line 272: Can we move this into a new module in `lib` I can see a `post_ticket` method getting use in other actions too so it would be good to move it before we duplicate it after forgetting we have it

rules/openstack.deletedmachines.rule.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+trigger:
+  type: "stackstorm_openstack.openstack.deletingmachines"
+action:
+  ref: "stackstorm_openstack.atlassian.create.tickets"
+  parameters:
+    tickets_info: "{{ trigger }}"
+    email: "{{ st2kv.system.jsm_email }}"
+    api_key: "{{ st2kv.system.jsm_key }}"
+    servicedesk_id: "12"
```
DavidFair, line 14: Could we put a comment on 12 and 94 what they map into Atlassian in this rule and the next one

sensors/src/deleting_machines_sensor.py
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        with OpenstackConnection(cloud_name=cloud_account) as conn:
+            projects = conn.list_projects()
+        for project in projects:
+            deleted = self._check_deleted(project=project["id"], cloud=cloud_account)
+            print(output["server_list"])
+            output["server_list"].extend(deleted)
+        print(output["server_list"])
+        if len(output["server_list"]) > 0:
```
DavidFair, line 50: A list with elements is "truthy" in Python so this is the equivalent and preferred in Python for future reference:
```python
if (output["server_list"]):
    self.sensor_service.dispatch_with_context(
````

File: sensors/src/loadbalancer_sensor.py [hunk 24]
```diff
@@ -0,0 +1,167 @@
...
+            "title": "{p[title_text]}",
+            "body": "The loadbalance ping test result was: {p[lb_status]}\nThe status of the amphora was: {p[amp_status]}\nThe amphora id is: {p[amp_id]}\nThe loadbalancer id is: {p[lb_id]}",
+            "server_list": [],
+        }
+        if amphorae.status_code != 200:
+            # Notes problem with accessing api if anything other than 403 or 200 returned
+            logging.critical("We encountered a problem accessing the API")
+            logging.critical("The status code was: %s ", str(amphorae.status_code))
+            logging.critical("The JSON response was: \n %s", str(amph_json))
+            return False
```
DavidFair, line 43: `# pylint: disable=inconsistent-return-statements`
Can be removed if this line simply is
`return`
and....

sensors/src/loadbalancer_sensor.py
```diff
@@ -0,0 +1,167 @@
...
+                        }
+                    )
+            else:
+                logging.info("%s is fine.", i["id"])
+        if len(output["server_list"]) > 0:
+
+            self.sensor_service.dispatch(
+                trigger="stackstorm_openstack.openstack.loadbalancer", payload=output
+            )
+            return output
```
DavidFair, line 109: ...This `return output` can be dropped as the sensors does not expect us to return anything from `run`

sensors/src/loadbalancer_sensor.py
```diff
@@ -0,0 +1,167 @@
...
+            logging.critical("The JSON response was: \n %s", str(amph_json))
+            return False
+
+        # Gets list of amphorae and iterates through it to check the loadbalancer and amphora status.
+        for i in amph_json["amphorae"]:
+            status = self._check_status(i)
+            ping_result = self._ping_lb(i["lb_network_ip"])
+            # This section builds out the ticket for each one with an error
+            if status[0] == "error" or ping_result == "error":
+                if status[0].lower() == "error" and ping_result.lower() == "error":
```
DavidFair, line 51: I would simplify these conditions since the latter implies the former and humans struggle with multiple conditions in a row (it tends to be bug-prone). 
Simply use a chain of ifs (not elifs as it's not Pythonic, since people have to figure out if the previous condition affected the future one)

```python
if status[0].lower() == "error" and ping_result.lower() == "error":
   ...
if status[0].lower() != "error" and ping_result.lower() == "error":
  ...
if status[0].lower() == "error" and ping_result.lower() != "error":
  ...
if status[0].lower() != "error" and ping_result.lower() != "error":
   print("Everything is fine")
```

It also makes it easier to see we've handled 2^2 = 4 combos as there's 4 if branches.

sensors/src/loadbalancer_sensor.py
```diff
@@ -0,0 +1,167 @@
...
+                + "\nThe body was: \n"
+                + str(amphorae.content)
+            )
+
+    # pylint: disable=missing-function-docstring
+    @staticmethod
+    def _check_status(amphora):
+        # Extracts the status of the amphora and returns relevant info
+        status = amphora["status"]
+        if status in ("ALLOCATED", "BOOTING", "READY"):
```
DavidFair, line 136: `status.upper() in`

DavidFair (changes requested):
Some tidy up and nits, otherwise it looks really close to completion

rules/openstack.deletedmachines.rule.yaml
```diff
@@ -0,0 +1,15 @@
...
+enabled: true
+
+trigger:
+  type: "stackstorm_openstack.openstack.deletingmachines"
+action:
+  ref: "stackstorm_openstack.atlassian.create.tickets"
+  parameters:
+    tickets_info: "{{ trigger }}"
+    email: "{{ st2kv.system.jsm_email }}"
+    api_key: "{{ st2kv.system.jsm_key }}"
```
DavidFair, line 13: These KV entries need adding to the `readme` similar to https://github.com/stfc/st2-cloud-pack#setup-openstack

actions/src/openstack_check_actions.py
Reillyhewitson: Stackstorm requires that payloads be dictionaries. I pass these as part of the payload so this would not work.

DavidFair (approved):
I'm happy to merge these changes, we're just re-testing the functionality then I'll merge
Nice work, looking forward to these checks being enabled :+1:

Reillyhewitson:
Action run failed. So investigated and found a major logic issue where the script would return after finding the first matching rule in a group. Solved so that the script will continue to through the group and return all matching rules

DavidFair (approved):

PR #7, opened 2022-07-26
Add old VM emails

joelvdavies:
Adds in and finishes some work Anish had done, refactored to work with clouds.yaml.
New actions:

- send.email 
- email.server.users

Also updates and fixes:

- server.list

Removes (these were broken anyway):
- send.server.info.email
- server.list.per.user


actions/send.email.yaml
```diff
@@ -0,0 +1,57 @@
...
+enabled: true
+entry_point: src/send_email.py
+name: send.email
+parameters:
+    send_as_html:
+        description: "If True Send Email as HTML, else as plain text"
+        required: true
+        type: boolean
+        default: true
+    admin_override:
```
DavidFair, line 12: This should be named like, "dry-run" or "test-run" rather than admin override

actions/send.email.yaml
```diff
@@ -0,0 +1,57 @@
...
+        description: "Override sending emails to instead send to one test email address"
+        required: true
+        type: boolean
+        default: false
+    admin_override_email:
```
DavidFair, line 17: I think this should be dropped
If we have "test-run" (or equiv) enabled we should use the email address (we enter the admin one instead) and send a single or random subset (rather than all) of the emails to this address

actions/send.email.yaml
```diff
@@ -0,0 +1,57 @@
...
+    smtp_account:
+        description: "SMTP Account to use. Must be configured in email.yaml"
+        required: true
+        type: string
+    email_to:
+        description: "Recipient Email Addresses"
+        required: true
+        type: array
+    email_from:
+        description: "Sender Email Address"
```
DavidFair, line 30: I'm not keen on this parameter, it completely opens us to spoofing. I'd rather hardcode something like `cloudsupport` or `stackstorm` as the sender

actions/send.server.info.email.yaml
```diff
@@ -1,98 +1,116 @@
 ---
-description: Email notifications for error emails
+description: Sends email notifications to VM owners based on a given query.
 enabled: true
 entry_point: workflows/send.server.info.email.yaml
```
DavidFair: I'm going to bikeshed here a little

Send server info implies it's just a daily report. Instead it would be better if we called it something like `report.server.problems`. Then leave the fact it's by email as an implementation detail, since its equally valid to use Jira / Slack / Letters in the post....etc.

actions/send.server.info.email.yaml
```diff
@@ -1,98 +1,116 @@
...
+      - "name_in"
+      - "name_not_in"
+      - "name_contains"
+      - "name_not_contains"
+      - "server_shutoff"
+      - "server_error_and_shutoff"
+      - "server_error"
+      - "server_older_than"
+      - "server_younger_than"
     required: true
```
DavidFair: So we should have an optional property `project` which limits this workflow to a project.

I'd also like to put in guards that any of these cannot be ran without a project specified:
```
      - "id_in"
      - "id_not_in"
      - "name_in"
      - "name_not_in"
      - "name_contains"
      - "name_not_contains"
```

Especially the not_contains could result in 49,995 emails being sent because 5 VMs didn't match a name....

actions/src/queryopenstack/classes/list_ips.py
```diff
@@ -34,10 +34,16 @@ def __init__(self, conn):
             "not_project_id": lambda func_dict, args: func_dict["project_id"]
             not in args,
             "older_than": lambda func_dict, args: self.is_older_than_x_days(
-                func_dict, days=args
+                func_dict["created_at"], days=args[0]
```
DavidFair, line 37: This is the old API and will not work for you, you'll need to write them using the Openstack API layer similar to https://github.com/DavidFair/st2-openstack-pack/blob/master/actions/src/project_action.py

actions/src/queryopenstack/classes/list_items.py
```diff
@@ -159,37 +159,45 @@ def get_property_func(self, key):
         """
         return self.property_func_dict.get(key, None)
 
-    def is_older_than_x_days(self, created_at, days):
+    def is_older_than_x_days(
```
DavidFair, line 162: Ditto ^

DavidFair (commented):
Partial review

actions/src/search_openstack.py
```diff
@@ -66,7 +77,8 @@ def get_preset(
 
         # search openstack_resource ID presets (works for all openstack objects)
         func = {
-            "id_in": lambda properties_list, criteria_list, sort_by_list, **kwargs: query(
+            "id_in": lambda cloud_account, properties_list, criteria_list, sort_by_list, **kwargs: query(
```
DavidFair, line 80: Are we sticking with this, or is the the legacy way?

As lambdas and methods are equivalent can we change these to methods and simply pass the method in instead. It's a lot more maintainable and allows us to later wrap each in tests

lib/email_api/email_api.py
```diff
@@ -0,0 +1,178 @@
...
+from email.header import Header
+from email.mime.application import MIMEApplication
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
+from email.utils import formatdate
+from typing import Dict
+
+
+class EmailApi:
+    def load_template(self, prop, **kwargs):
```
DavidFair, line 13: Use named params instead of kwargs

actions/src/queryopenstack/query.py
```diff
@@ -45,7 +47,7 @@ def query(
 
     # if no user defined openstack connection given - create default one
     if not openstack_conn:
-        openstack_conn = create_openstack_connection()
+        openstack_conn = create_openstack_connection(cloud_name=cloud_account)
```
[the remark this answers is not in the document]
joelvdavies: This is the old code along.

actions/src/search_openstack.py
joelvdavies: This is the old code.

actions/src/email_actions.py
```diff
@@ -0,0 +1,85 @@
...
+        :param properties_to_select: The list of properties to select and output from the found servers
+        :param get_html: When True tables returned are in html format
+        :return:
+        """
+        servers = self._server_api[f"search_{query_preset}"](
+            cloud_account, project_identifier, **kwargs
+        )
+
+        if not "user_email" in properties_to_select:
+            raise ValueError("properties_to_select must contain 'user_email'")
```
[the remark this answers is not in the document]
joelvdavies: But the user_email is something selected in the servers action before it is grouped by it. Since the action is the only place it assumes this property exists I don't think it is suited for the api.

lib/email_api/email_api.py
joelvdavies: @DavidFair Because of the way I am selecting the query to run in OpenstackServer, would it be okay to leave the kwargs in there? - they are not used but must be present e.g. search_all_servers doesn't need the day parameter and search_servers_older_than does so the former will cause an error if passed days. The only alternative I can think of is using a massive if statement to switch between the different functions which is a lot more messy.

actions/send.server.email.yaml
```diff
@@ -1,61 +1,129 @@
 ---
-description: Email notifications for error emails
+description: Sends email notifications to VM owners based on a given query.
 enabled: true
-entry_point: workflows/send.server.email.yaml
+entry_point: src/email_actions.py
 name: send.server.email
```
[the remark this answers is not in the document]
joelvdavies: How about email.server.users?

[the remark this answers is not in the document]
DavidFair: Yeah that sounds loads better and I can tell what it does just from the name :+1:

actions/src/email_actions.py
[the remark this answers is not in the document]
DavidFair: Ah okay, so this is to select grabbing emails from the Openstack query :+1:

lib/email_api/email_api.py
DavidFair: Can we document the supported ones in a doc string then or put the URL to where a future dev can see the supported ones? 

In this case that's fine since we do genuinely have dynamic kwargs

actions/src/queryopenstack/query.py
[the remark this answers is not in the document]
DavidFair: Can we revert this using `git checkout main actions/src/queryopenstack/query.py` so it doesn't appear in the PR?

actions/src/email_actions.py
```diff
@@ -26,6 +26,7 @@ def run(self, submodule: str, **kwargs):
     def send_email(
         self,
         subject: str,
+        email_to: str,
```
DavidFair, line 29: This type doesn't match the description: `str != List[str]`

DavidFair:
I think we just need some unit tests and I'll be happy to get this in
We'll also need to test the various queries on Dev. My plan is to get this into `main` then put together a test matrix for the various query types we support to ensure they all work as expected.

If they all work, we'll install the pack onto prod, if not we'll do follow up PR(s) to either fix/disable the functionality that aren't working as expected

lib/email_api/email_api.py
```diff
@@ -0,0 +1,219 @@
...
+from smtplib import SMTP_SSL
+from email.header import Header
+from email.mime.application import MIMEApplication
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
+from email.utils import formatdate
+from typing import Dict, List
+
+
+class EmailApi:
```
DavidFair, line 12: Nit:

To make testing this easier, I was wondering if it's worth moving the pack config to `__init__` as a member (i.e. `pack_config`), as I want to port the Jupyter stuff to use the same pattern too?

It would be nice to come up with an established pattern to get keys / creds code-wise between us that works across all of them

lib/email_api/email_api.py
joelvdavies: Where would be the best place to put this? The supported ones are defined in the enum so it produces a drop down list, otherwise they come from the functions defined in server_action.

lib/email_api/email_api.py
DavidFair: Edit: I think this worth a discussion on call, as it would be nice to either adopt the AWS pack pattern or pass through the pack config entirely through

tests/lib/email/test_email_api.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def setUp(self) -> None:
+        self.api = EmailApi()
+
+    @raises(FileNotFoundError)
+    def test_load_template_file_not_found(self):
+        """
+        Tests that load_template can raise a FileNotFoundError
+        """
+        self.assertRaises(FileNotFoundError, self.api.load_template("test"))
```
DavidFair, line 39: I think this is duplicated, this should just be `self.api.load_template("test")` since we already have the `@raises` decorator?

tests/lib/email/test_email_api.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def test_send_email_insecure(self):
+        """
+        Tests that send_email functions as expected
+        """
+        # Replace to avoid attempting to load files
+        self.api.load_template = MagicMock(return_value="")
+
+        # Avoid actually attempting to send any emails
+        with mock.patch("email_api.email_api.SMTP_SSL") as smtp_mock:
```
DavidFair, line 75: Can we move this into `setUp` so all methods by default have this patched out?

DavidFair (changes requested):

lib/openstack_api/openstack_server.py
```diff
@@ -0,0 +1,286 @@
...
+        Returns a list of Servers matching a given query
+        :param cloud_account: The associated clouds.yaml account
+        :param project_identifier: The project to get all associated servers with, can be empty for all projects
+        :return: A list of all servers
+        """
+        selected_servers = []
+        if project_identifier == "":
+            projects = self._identity_api.list_projects(cloud_account)
+        else:
+            projects = [
```
DavidFair, line 57: I think you can only search a project at a time looking at the params, any reason why we pass the projects around as a list?

tests/lib/test_openstack_identity.py
```diff
@@ -197,3 +197,25 @@ def test_find_user_returns_result(self):
...
+
+    def test_find_user_all_domains_returns_result(self):
+        """
+        Checks that find_user_all_domains returns the correct result
+        """
+        cloud, user = NonCallableMock(), NonCallableMock()
+        returned = self.instance.find_user_all_domains(cloud, user)
+
+        self.mocked_connection.assert_called_once_with(cloud)
+        self.identity_api.find_user.assert_called_once_with(
```
DavidFair, line 216: Do we know what will happen if a duplicate record is found? Will it throw a `MultipleRecords` exception or return both?

tests/lib/test_openstack_server.py
```diff
@@ -0,0 +1,320 @@
...
+
+        self.api.list_servers.assert_called_once_with(
+            filters={
+                "all_tenants": True,
+                "project_id": self.identity_module.find_mandatory_project.return_value.id,
+                "limit": 10000,
+            }
+        )
+
+    @mock.patch("openstack_api.openstack_query.datetime", wraps=datetime)
```
DavidFair, line 108: Why are we patching instead of doing dependency injection here?
Patching is alright for legacy / existing code, but it adds a burden to writing all tests in the future since we have to patch the dep each time rather than do something like `self.mocked_query`

DavidFair (changes requested):

lib/openstack_api/openstack_server.py
```diff
@@ -0,0 +1,286 @@
...
+        Returns a list of servers with ids that aren't in the list given
+        :param cloud_account: The associated clouds.yaml account
+        :param project_identifier: The project to get all associated servers with, can be empty for all projects
+        :return: A list of servers matching the query
+        """
+        selected_servers = self.search_all_servers(cloud_account, project_identifier)
+
+        return self._query_api.apply_query(
+            selected_servers,
+            lambda a: all(x in a["status"] for x in ["SHUTOFF", "ERROR"]),
```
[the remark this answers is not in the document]
joelvdavies: Its only in two places and I think the solution makes it messier so would rather leave it if possible.

lib/openstack_api/openstack_server.py
joelvdavies: Yes as it will iterate over all of them when searching for servers.

tests/lib/test_openstack_server.py
joelvdavies: This is for datetime not openstack_query

tests/lib/test_openstack_identity.py
joelvdavies: Speaking to Alex & Jacob sounds like it would likely just error as keystone would return a 501 error. Not sure how the API would react to that, but I would think its likely it would just crash there before sending any emails.

DavidFair: Lets apply YAGNI to this then

tests/lib/test_openstack_identity.py
```diff
@@ -172,6 +173,15 @@ def test_find_mandatory_project_raises_for_missing(self):
         self.instance.find_project = Mock(return_value=None)
         self.instance.find_mandatory_project(NonCallableMock(), NonCallableMock())
 
+    def test_list_projects(self):
+        """
+        Tests calling the API wrapper without a domain will throw
```
DavidFair, line 178: The doc string is incorrect for this? As it doesn't throw (or we don't check it does)

DavidFair (approved):
LGTM, lets :shipit:

PR #10, opened 2022-08-01
Switch to using yaml for Jupyter API Keys

DavidFair:
This is part of the work moving from the K-V store into the YAML file.
Ports the Jupyter API to use pack configs instead. The next PR will add support for deleting a number of accounts based on a template name.


joelvdavies (approved):
Looks good.

PR #13, opened 2022-08-08
Tweaks for VM emails

joelvdavies:
- Adds try/except to ignore an error caused by the OpenStack API attempting to fetch details of servers listed in a project when they no longer exist
- Fixes error in email.server.users
- Output "No results" instead of an error when no results are found in server.list

### DIFF
--- a/lib/openstack_api/openstack_query.py
+++ b/lib/openstack_api/openstack_query.py
@@ -180,6 +180,9 @@ class OpenstackQuery(OpenstackWrapperBase):
         property_funcs = self.get_default_property_funcs(object_type, cloud_account)
         output = self.parse_properties(items, properties_to_select, property_funcs)
 
+        if len(output) == 0:
+            return "No results"
+
         if group_by != "":
             output = self.collate_results(output, group_by, get_html)
         else:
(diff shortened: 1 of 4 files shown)


lib/openstack_api/openstack_query.py
```diff
@@ -180,6 +180,9 @@ def parse_and_output_table(
         property_funcs = self.get_default_property_funcs(object_type, cloud_account)
         output = self.parse_properties(items, properties_to_select, property_funcs)
 
+        if len(output) == 0:
+            return "No results"
```
DavidFair, line 184: Return `None` then interpret this in the action as no results. Since the lib layer shouldn't be doing anything user facing (bar exceptions)

lib/openstack_api/openstack_server.py
```diff
@@ -62,15 +63,19 @@ def search_all_servers(
...
             for project in projects:
-                selected_servers.extend(
-                    conn.list_servers(
-                        filters={
-                            "all_tenants": True,
-                            "project_id": project.id,
-                            "limit": 10000,
-                        }
+                try:
+                    selected_servers.extend(
```
DavidFair, line 67: Can we write a unit test for this, that `list_servers` throwing a HttpException is handled since it's something we discovered in prod?

Also the unit test can be used as the template for a future action too

DavidFair (changes requested):

DavidFair (dismissed):

joelvdavies:
Missed `.id` 😅

DavidFair (approved):

PR #15, opened 2022-08-11
Add floating.ip.list

joelvdavies:
Adds a floating.ip.list action for searching floating ips. Largely the same as send.server.list.


tests/lib/test_openstack_floating_ip.py
```diff
@@ -0,0 +1,278 @@
...
+
+    def test_search_all_fips_no_project(self):
+        """
+        Tests calling search_all_fips with no project selected
+        """
+        self.instance.search_all_fips(cloud_account="test", project_identifier="")
+
+        self.mocked_connection.assert_called_once_with("test")
+
+        self.api.list_floating_ips.assert_has_calls(
```
DavidFair, line 58: Could we use `assert_called_once_with` instead?

DavidFair (changes requested):
Looks good, just one question then I'm happy to merge

tests/lib/test_openstack_floating_ip.py
joelvdavies: Yep, done

DavidFair (approved):

PR #17, opened 2022-08-12
Add project emails

joelvdavies:
Adds an `email` parameter to the project.create action and puts it in the project tags. Also allows them to be listed in floating.ip.list.

### DIFF
--- a/lib/openstack_api/openstack_identity.py
+++ b/lib/openstack_api/openstack_identity.py
@@ -1,4 +1,4 @@
-from typing import List, Optional
+from typing import List, Optional, Union
 
 from openstack.exceptions import ConflictException
 from openstack.identity.v3.project import Project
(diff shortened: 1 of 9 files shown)


lib/openstack_api/openstack_identity.py
```diff
@@ -136,3 +141,29 @@ def find_user_all_domains(
...
+
+    def find_project_email(self, cloud_account: str, project_identifier: str):
+        """
+        Returns the contact email of a project
+        :param cloud_account: The clouds entry to use
+        :param project_identifier: The name or Openstack ID for the project
+        :return: The found email or None
+        """
+        found_email = None
+        project = self.find_project(cloud_account, project_identifier)
```
DavidFair, line 166: Should this be find_mandatory_project ? As it should always be a valid project specified

joelvdavies: I will double check there are no bugged FIPs, otherwise yes it should be.

joelvdavies: There are some, so it will have to be find project.

DavidFair (approved):

PR #18, opened 2022-08-12
Update project.list

joelvdavies:
Updates the project.list action so that it works again. Also includes a "projects_without_email" query to find projects that don't have an email assigned.


lib/openstack_api/openstack_query.py
```diff
@@ -36,19 +37,41 @@ def query_prop_not_in(self, prop: str, values: List[str]):
...
+        )
 
     def query_prop_not_contains(self, prop: str, snippets: List[str]):
         """
         Returns a query for checking if a property value does not contain all the snippets
         given in a list
         """
-        return lambda a: all(snippet not in a[prop] for snippet in snippets)
+        return self._query_if_prop_exists(
+            prop, lambda a: all(snippet not in a[prop] for snippet in snippets), True
```
DavidFair, line 73: Could we just do something like, 
```
return self._query_if_prop_exists(
    prop, not self.query_prop_contains(*args)
)
```

DavidFair (commented):
LGTM, just one question:

lib/openstack_api/openstack_query.py
DavidFair: Ah because of the nested bool operations in `query_if_prop_exists` rather than it being a 1:1 true or false

DavidFair (approved):

PR #19, opened 2022-08-15
Add project email to internal & external create workflows

joelvdavies:
Noticed I didn't add the project email to these workflows so they would likely have errored.


DavidFair (approved):
LGTM, would be good to test against dev to be sure but I'm happy to merge after

PR #21, opened 2022-08-15
Add project update action

joelvdavies:
Adds project.update action for updating project properties like name, description, email.

Currently a draft as currently unable to test.


actions/src/project_action.py
```diff
@@ -80,3 +80,38 @@ def project_create(
...
+        :param cloud_account: The account from the clouds configuration to use
+        :param project_identifier: Name or Openstack ID of the project to update
+        :param name: Name of the project
+        :param email: Contact email of the project
+        :param description: Description of the project
+        :param is_enabled: Enable or disable the project (takes values of unchanged, true or false)
+        :return: status, optional project
+        """
+        is_enabled_value = None
+        if is_enabled == "true":
```
DavidFair, line 104: I'd do this using casefold, since I saw Megan fighting a bug yesterday where it randomly changed case on something that should never change case:

```
if is_enabled.casefold() == "true".casefold():
```

lib/openstack_api/openstack_identity.py
```diff
@@ -142,18 +142,64 @@ def find_user_all_domains(
...
+        """
+        Returns the index of a project tag matching a given query
+        :param project_tags: The project tags
+        :param select_func: Function that determines whether a tag should be selected
+        :return: The index of found tag or -1
+        """
+        for i, project_tag in enumerate(project_tags):
+            if select_func(project_tag):
+                return i
+        return -1
```
DavidFair, line 157: So this is C-Style coding so slightly goes against being Pythonic

In this case can we simply:
```python
def _find_project_tag_index( ... ) -> Optional[int]:
    ....
        return i
    return None
```

Callers should be explicitly checking for `result is None` rather than `not result` explicitly with positional indices anyway to prevent `0 == false` bugs

DavidFair (changes requested):
Some minor nits

DavidFair (approved):

PR #22, opened 2022-08-15
Add Create Tickets action documentation

Reillyhewitson:
There's some documentation in the source code for the action but I wanted to make it easier to find. If there's anything about this action you have questions about let me know so that I can add it to the documentation.


DavidFair (approved):
LGTM thanks

PR #27, opened 2022-08-16
Add documentation for openstack.check.*, openstack.*.rule and openstack.projects.sync

Reillyhewitson:
Documentation for the openstack check actions and sensors/rules as well as openstack.projects.sync


joelvdavies (approved):
Looks good to me

PR #28, opened 2022-08-17
Add image list action

joelvdavies:
Adds image.list action.

### DIFF
--- a/actions/image.list.yaml
+++ b/actions/image.list.yaml
@@ -36,6 +36,7 @@ parameters:
       - "images_name_not_in"
       - "images_name_contains"
       - "images_name_not_contains"
+      - "images_non_existent_project"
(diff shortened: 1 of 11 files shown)


actions/image.list.yaml
```diff
@@ -0,0 +1,85 @@
+---
+description: List images
+enabled: true
+entry_point: src/image_actions.py
+name: image.list
+parameters:
+  timeout:
+    default: 99999
```
DavidFair, line 8: The timeout is a tad too long, we should cap it to something sensible like `3600` seconds (1 hour) or so

joelvdavies: 1.5 hours? - I have seen 45 minutes before, it may take longer in the future

PR #31, opened 2022-08-18
Add image emails

joelvdavies:
Adds the email.image.users action and does some clean-up to reduce some code repetition.

Also noticed an error when there are no emails to send, so fixed that.

### DIFF
--- a/actions/src/email_actions.py
+++ b/actions/src/email_actions.py
@@ -1,6 +1,7 @@
 from typing import Callable, Dict, List
 from email_api.email_api import EmailApi
 from openstack_api.openstack_floating_ip import OpenstackFloatingIP
+from openstack_api.openstack_image import OpenstackImage
 from openstack_api.openstack_query import OpenstackQuery
 from openstack_api.openstack_server import OpenstackServer
 
@@ -16,6 +17,9 @@ class EmailActions(Action):
         )
         self._floating_ip_api: OpenstackServer = config.get(
             "openstack_floating_ip_api", OpenstackFloatingIP()
+        )
+        self._image_api: OpenstackImage = config.get(
+            "openstack_image_api", OpenstackImage()
         )
         self._query_api: OpenstackQuery = config.get(
             "openstack_query_api", OpenstackQuery()
@@ -92,7 +96,7 @@ class EmailActions(Action):
         """
         Finds all servers matching a query and then sends emails to their users
         :param: cloud_account: The account from the clouds configuration to use
-        :param: project_identifier: The project this applies to (or empty for all servers)
+        :param: project_identifier: The project this applies to (or empty for all projects)
         :param: query_preset: The query to use when searching for servers
         :param: message: Message to add to the body of emails sent
         :param: properties_to_select: The list of properties to select and output from the found servers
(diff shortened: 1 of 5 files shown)


actions/src/email_actions.py
```diff
@@ -1,13 +1,34 @@
...
 from openstack_api.openstack_floating_ip import OpenstackFloatingIP
+from openstack_api.openstack_image import OpenstackImage
 from openstack_api.openstack_query import OpenstackQuery
 from openstack_api.openstack_server import OpenstackServer
 
 from st2common.runners.base_action import Action
 
 
 class EmailActions(Action):
+    @dataclass
```
DavidFair, line 13: Can we extract this out to `lib/structs` (which should probably be called `lib/dataclasses` to align with Python instead of C)

That way it's clear this is shared between the lib and action layer, rather than being an implementation detail in just the action layer

actions/src/email_actions.py
```diff
@@ -1,13 +1,34 @@
...
+    class EmailActionParams:
+        """
+        Structure containing the information needed to _email_users for a particular OpenstackResource
+        :param: required_email_property: The name of the property that must be obtained to get the email of the
```
DavidFair, line 17: So the dataclass doesn't check if the property is not in the properties to select. This snippet of doc-string should be moved to where that check is actually done.

Ditto on the errors too, since this is a plain ol' dumb data class with no additional checking

actions/src/email_actions.py
```diff
@@ -17,6 +38,9 @@ def __init__(self, *args, config: Dict = None, **kwargs):
         self._floating_ip_api: OpenstackServer = config.get(
             "openstack_floating_ip_api", OpenstackFloatingIP()
         )
+        self._image_api: OpenstackImage = config.get(
```
DavidFair, line 41: Longer-term I can see this splitting across action files, i.e. `openstack.image.email_users` since it's a code smell of this action gradually turning into a god action.

Next time we go to add an email action can we split this down. For the moment it's fine as it is though.

tests/actions/test_email_actions.py
```diff
@@ -359,6 +360,129 @@ def test_email_floating_ip_users_no_project_error(self):
...
+        )
+
+    def test_email_image_users_no_project(self):
+        """
+        Tests the action that sends emails to image users does not give a value error when a project
+        is required for the query type
+        """
+
+        i = 0
+        for query_preset in OpenstackImage.SEARCH_QUERY_PRESETS_NO_PROJECT:
```
DavidFair, line 457:
```
for i, query_present in enumerate(OpenstackImage.SEARCH_QUERY_PRESETS_NO_PROJECT):
            self._email_image_users(query_preset)
            self.assertEqual(self.email_mock.send_emails.call_count, i)
```

DavidFair (changes requested):

actions/src/email_actions.py
joelvdavies: This isn't shared with the lib layer though, its only used in this file.

tests/actions/test_email_actions.py
joelvdavies: Nice, thanks

joelvdavies:
Pylint is giving errors for code I haven't touched due to a recent update adding a new `missing-timeout` check. I will look at fixing them in another PR.

actions/src/email_actions.py
DavidFair: Okay, then can we hoist it out of EmailActions into it's own class prefixed with `_` to `_EmailActionParams` to indicate it's a private implementation detail currently?
That way we don't have nested class structures within each module too.

DavidFair (approved):
LGTM, my outstanding comment will be picked up in a future PR

PR #32, opened 2022-08-19
Add non existence queries

joelvdavies:
Adds functions for finding servers, floating ips and images with non-existent projects, as well as one for finding servers that are listed in projects but do no exist.

Adds the actions:
correctness.find.non.existent.servers
correctness.find.non.existent.floating.ips
correctness.find.non.existent.images
correctness.find.servers.in.non.existent.projects
correctness.find.floating.ips.in.non.existent.projects
correctness.find.images.in.non.existent.projects

### DIFF
--- a/actions/find.non.existent.floating.ip.projects.yaml
+++ b/actions/find.non.existent.floating.ip.projects.yaml
@@ -1,0 +1,18 @@
+---
+description: Find non-existent projects referenced by floating ips
+enabled: true
+entry_point: src/floating_ip_actions.py
+name: find.non.existent.floating.ip.projects
+parameters:
+  timeout:
+    default: 5400
+  submodule:
+    default: find_non_existent_projects
+    type: string
+    immutable: true
+  cloud_account:
+    description: "The clouds.yaml account to use whilst performing this action"
+    default: ""
+    required: true
+    type: string
+runner_type: python-script
--- a/lib/openstack_api/openstack_query.py
+++ b/lib/openstack_api/openstack_query.py
@@ -1,11 +1,12 @@
 import datetime
-from dataclasses import dataclass
 from typing import Any, Callable, Dict, List
 
 import openstack
-from openstack.connection import Connection
-from openstack.identity.v3.project import Project
 from tabulate import tabulate
+from openstack_api.dataclasses import (
+    NonExistentCheckParams,
+    NonExistentProjectCheckParams,
+)
 from openstack_api.openstack_connection import OpenstackConnection
 
 from openstack_api.openstack_identity import OpenstackIdentity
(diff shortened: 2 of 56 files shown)


actions/find.non.existent.floating.ip.projects.yaml
```diff
@@ -0,0 +1,18 @@
+---
+description: Find non-existent projects referenced by floating ips
+enabled: true
+entry_point: src/floating_ip_actions.py
+name: find.non.existent.floating.ip.projects
```
DavidFair, line 5: These should be grouped into floating_ip operations, i.e. `floating.ip.find.non.existent.projects` or alternatively we have some sort of:
`correctness.find.floating.ip.on.non.existent.project`

(Also I'm getting more and more tempted to move away from the pre-existing `.` separator and switch to the `correctness.find_floating_ip_on_non_existent_project` but not for this PR)

lib/openstack_api/openstack_query.py
```diff
@@ -298,3 +304,104 @@ def parse_and_output_table(
             output = self.generate_table(output, get_html)
 
         return output
+
+    @dataclass
+    class NonExistentCheckParams:
```
DavidFair, line 309: This dataclass should go into it's own file, in this case it could sit in the `lib/openstack_api` module since it's an implementation detail not shared with the actions layer

lib/openstack_api/openstack_query.py
```diff
@@ -298,3 +304,104 @@ def parse_and_output_table(
...
+                        check_params.object_get_func(conn, object_id)
+                    except openstack.exceptions.ResourceNotFound:
+                        if project.id in selected_projects:
+                            selected_projects[project.id].append(object_id)
+                        else:
+                            selected_projects.update({project.id: [object_id]})
+        return selected_projects
+
+    @dataclass
+    class NonExistentProjectCheckParams:
```
DavidFair, line 367: Can the above dataclass inherit this one to re-use the fields? Since there's duplication going on between these two

tests/lib/test_openstack_floating_ip.py
```diff
@@ -299,3 +295,71 @@ def test_search_fips_down_before(self, mock_datetime):
         )
 
         self.assertEqual(result, [self.mock_fip_list[0]])
+
+    def test_find_non_existent_fips(self):
+        """
+        Tests calling find_non_existent_fips
+        """
+        # pylint:disable=too-few-public-methods,invalid-name,redefined-builtin
+        class ObjectMock:
```
DavidFair, line 304: So `redefined-builtin` is something we don't want to disable, as it can break in unexpected ways for future devs as you're effectively shadowing builtin objects.

In this case I would simply create a stub dataclass like so:
```
@dataclass
class _StubbedReturn:
    id: str
    project_id: str

def test_foo():
    instance = _StubbedReturn("bar", "baz")
    ...
```

And let the dataclass handle it for you, since we're not actually calling these so a mock is a touch overkill

tests/lib/test_openstack_floating_ip.py
```diff
@@ -299,3 +295,71 @@ def test_search_fips_down_before(self, mock_datetime):
...
+                ]
+            },
+        )
+
+    def test_find_non_existent_projects(self):
+        """
+        Tests calling find_non_existent_projects
+        """
+        # pylint:disable=too-few-public-methods,invalid-name,redefined-builtin
+        class ObjectMock:
```
DavidFair, line 343: ^

DavidFair (changes requested):
Mostly changes around stubbing in the unit testing layer

tests/lib/test_openstack_floating_ip.py
joelvdavies: The redefined-builtin was actually due to id which must be that name since it is from the openstack API, but using a dataclass fixed it anyway since it was only a problem in the constructor.

joelvdavies:
Pylint is giving errors for code I haven't touched due to a recent update adding a new missing-timeout check. I will look at fixing them in another PR.

DavidFair (approved):
LGTM, building on this I'd like to create polling sensors which trigger an email action when these are discovered

That way we can fix all/most of them then be alerted automatically in the future. But that's obviously for the future.

PR #34, opened 2022-08-19
Freeipa create users

DavidFair:
Adds a workflow and supporting action(s) to create users on a given FreeIPA server.


actions/workflow.freeipa.user.create.yaml
```diff
@@ -0,0 +1,30 @@
...
+description: Creates a given list of users
+enabled: true
+entry_point: workflows/freeipa.user.create.yaml
+runner_type: orquesta
+
+parameters:
+  submodule:
+    default: user_delete
+    immutable: true
+    type: string
```
joelvdavies, line 11: This should not be here - its not taken in by the workflow.

actions/workflow.freeipa.user.create.yaml
```diff
@@ -0,0 +1,30 @@
...
+  server_env:
+    description: The FreeIPA environment to create users on
+    required: true
+    type: string
+    enum:
+      - prod
+      - dev
+  user_base_name:
+    description: The users to remove, or base name of the users to remove without the hyphen (e.g. "jupyter")
```
joelvdavies, line 20: This description isn't for creating users.

actions/workflow.freeipa.user.create.yaml
```diff
@@ -0,0 +1,30 @@
...
+    required: true
+    type: string
+  first_index:
+    description: The index of the first user to remove
```
joelvdavies, line 24: ^

actions/workflow.freeipa.user.create.yaml
```diff
@@ -0,0 +1,30 @@
...
+    required: false
+    type: integer
+  last_index:
+    description: The index of the last user to remove (inclusive).
```
joelvdavies, line 28: ^

joelvdavies (changes requested):
A few things that weren't changed after what I assume was a copy & paste.

actions/workflow.freeipa.user.create.yaml
ElliottKasoar: Should also be `name: workflow.freeipa.user.create` to be consistent with other `workflow.[x]` actions?

joelvdavies: @ElliottKasoar Yes I was in two minds about it, on the one hand they are all actions that happen to run workflows and freeipa.user.create would probably be the expected name for this but it is inconsistent, so I think your right it should be workflow.* for now at least.

joelvdavies (approved):
LGTM

PR #35, opened 2022-08-19
Add initial PR template

DavidFair:
This template should help both submitters and reviewers by nudging them
to the various pack standards used within


joelvdavies (approved):
Looks good.

PR #38, opened 2022-09-01
Fix user.list action

joelvdavies:
### Description:

Fixes user.list action bringing it in line with the other list actions.

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:


DavidFair (approved):
LGTM, thanks for bringing this into line with all the other actions you added

PR #39, opened 2022-09-02
Add immutable project flag

### LINKED ISSUE #20: Block deleting of certain projects
Prevent a user from deleting the admin/service projects for example. Suggested changes:

- Tag certain projects as immutable
- Add a "dry run" checkbox enabled by default to the action which returns the deleted project while only deleting when unchecked

The immutable tag may be used after an upgrade to OpenStack Victoria https://docs.openstack.org/python-openstackclient/wallaby/cli/command-objects/project-v3.html.

joelvdavies:
### Description:

Adds an immutable option to project create action/workflows and ensures projects that have the flag in their tags are not deleted by project.delete. Also ensures they cannot be updated by project.update except to remove the immutable flag itself. Finally a 'delete' checkbox is added to project.delete which by default is unchecked so that the action only prints out the project it found to reduce the chances of accidental deletion.

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

Closes #20

### DIFF
--- a/actions/src/project_actions.py
+++ b/actions/src/project_actions.py
@@ -34,20 +34,39 @@ class ProjectAction(Action):
         return func(**kwargs)
 
     def project_delete(
-        self, cloud_account: str, project_identifier: str
+        self, cloud_account: str, project_identifier: str, delete: bool
     ) -> Tuple[bool, str]:
         """
         Deletes a project
         :param cloud_account: The account from the clouds configuration to use
         :param project_identifier: (Either) The project name to delete
+        :param delete: When true will delete the project, when false will only return it to reduce
+                       chances of accidental deletion
         :return: The result of the operation
         """
-        delete_ok = self._identity_api.delete_project(
+        if delete:
+            delete_ok = self._identity_api.delete_project(
+                cloud_account=cloud_account, project_identifier=project_identifier
+            )
+            # Currently, we only handle one error, other throws will propagate upwards
+            err = "" if delete_ok else "The specified project was not found"
+            return delete_ok, err
+
+        project = self._identity_api.find_mandatory_project(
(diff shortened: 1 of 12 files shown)


actions/src/project_actions.py
```diff
@@ -34,20 +34,39 @@ def run(self, submodule: str, **kwargs):
...
         Deletes a project
         :param cloud_account: The account from the clouds configuration to use
         :param project_identifier: (Either) The project name to delete
+        :param delete: When true will delete the project, when false will only return it to reduce
+                       chances of accidental deletion
         :return: The result of the operation
         """
-        delete_ok = self._identity_api.delete_project(
+        if delete:
+            delete_ok = self._identity_api.delete_project(
```
DavidFair, line 48: So this is logic creeping into our actions layer

What would be better is if `delete_project` took the flag, then run all steps except for the real call. This also is trivial to unit test so we can have a dedicated test for it.

That way it truly is a dry run rather than two logic paths dependent on a flag.

actions/src/project_actions.py
```diff
@@ -127,6 +152,7 @@ def project_update(
         email: str,
         description: str,
         is_enabled: str,
+        immutable: str,
```
DavidFair, line 155: I thought Stackstorm passed through bool types rather that us having to do the comparison if we set the correct type in the action layer?

DavidFair (changes requested):

actions/src/project_actions.py
joelvdavies: I made it a drop down of true, false, unchanged along with the enabled property so we can leave them as they are rather than always updating them every time the action is run to update the other properties like email.

DavidFair (approved):
LGTM, great to get this in :+1:

PR #40, opened 2022-09-06
Create workflow to delete FreeIPA users

ElliottKasoar:
### Description:

Adds a workflow to delete users on a given FreeIPA server, based on the equivalent user creation workflow.

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:


joelvdavies (approved):
LGTM

PR #41, opened 2022-09-06
Add action to create Jupyter users

ElliottKasoar:
### Description:

Adds an action and tests for creating a range of Jupyter users, based on the similar user deletion action.

Includes changes based on #45

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/src/jupyter.py
+++ b/actions/src/jupyter.py
@@ -1,6 +1,7 @@
 from typing import Dict, Callable, Optional
 
 from jupyter_api.user_api import UserApi
+from jupyter_api.get_token import get_token
 from st2common.runners.base_action import Action
 from structs.jupyter_users import JupyterUsers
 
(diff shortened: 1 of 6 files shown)


actions/src/jupyter.py
```diff
@@ -29,18 +30,21 @@ def user_delete(
...
-        key_name = None
-        env = jupyter_env.casefold()
-        if env == "prod".casefold():
-            key_name = "prod_token"
-        if env == "training".casefold():
-            key_name = "training_token"
-        if env == "dev".casefold():
-            key_name = "dev_token"
-        if not key_name:
-            raise KeyError("Unknown Jupyter environment")
```
DavidFair, line 42: :+1: Good change, this is a faux-par on my part considering I bang on about not putting logic in the actions layer....

DavidFair (approved):
LGTM, glad to see this completed and it's a good refactor too :+1:

PR #42, opened 2022-09-07
Create workflow to spin up Jupyter users

ElliottKasoar:
### Description:

Adds a workflow to create users on FreeIPA and JupyterHub, and start up their servers, as well as a workflow to stop servers and delete users from JupyterHub and FreeIPA.

Depends on #40 

To do:

- Divide into smaller PRs
- Add unit tests for server start/stop

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:


DavidFair:
> I was about to submit one with server stop and start together, and one with the two complete workflows together, since they're very similar. Is that ok or would you rather each action completely separate?

That's perfect since it makes sense to group them together, I didn't see the other PR so sorry about asking you to do something you've already done :sweat_smile:

PR #46, opened 2022-09-08
Refactor emails

joelvdavies:
### Description:

Refactors the email code so we have functions for emailing users in the lib folder.

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:


lib/openstack_api/openstack_image.py
```diff
@@ -284,11 +287,55 @@ def find_non_existent_projects(self, cloud_account: str) -> Dict[str, List[str]]
...
             cloud_account=cloud_account,
             check_params=NonExistentProjectCheckParams(
                 object_list_func=lambda conn: conn.list_images(),
                 object_id_param_name="id",
                 object_project_param_name="owner",
             ),
         )
+
+    # pylint:disable=too-many-arguments
+    def email_users(
```
DavidFair, line 300: I think we should extract this method into a parent class, something like:
```
from abc import ABC
class EmailAction(ABC):
   def email_users(...):
        ...

   def email_support(...):
        ...
```

Then this class gets mixed into the actions:
```
class OpenstackImage(OpenstackWrapperBase, EmailAction):
```

This would give us a completely consistent set of params and remove the duplication
Thoughts?

lib/openstack_api/openstack_query.py
```diff
@@ -372,3 +379,76 @@ def find_non_existant_object_projects(
                     else:
                         selected_projects.update({project_id: [object_id]})
         return selected_projects
+
+    # pylint:disable=too-many-arguments,too-many-locals
+    def email_users(
```
DavidFair, line 384: So this method fails the "and or or" test
It performs the query *and* triggers the sending of emails.

Instead we should break it down into something like:
```
def notify_users(...):
    found_items = perform_query()
    prepared_emails = prepare_emails()
    send_emails(prepared_emails)
```

It might seem a touch overkill at the moment, but methods like this organically grow over time so it's better to trim them into separate responsibilities early on.

tests/lib/email_api/test_email_helpers.py
```diff
@@ -0,0 +1,55 @@
...
+            "port": 123,
+            "secure": True,
+            "smtp_auth": True,
+        },
+    ]
+
+    TEST_CONFIG = {"smtp_accounts": TEST_ACCOUNTS}
+
+    def setUp(self) -> None:
+        pass
```
DavidFair, line 32: This can be removed

tests/lib/test_openstack_image.py
```diff
@@ -349,3 +351,95 @@ def __getitem__(self, item):
         result = self.instance.find_non_existent_projects(cloud_account="test")
 
         self.assertEqual(result, {"ProjectID1": ["ImageID1", "ImageID2"]})
+
+    @raises(ValueError)
+    def test_email_users_no_email_error(self):
```
DavidFair, line 356: ...

Splitting this into a parent class would also allow us to remove duplication of these tests too

DavidFair (changes requested):

lib/openstack_api/openstack_image.py
joelvdavies: The only complication I can see is needing an OpenstackQuery instance in the new class as well which could be used in OpenstackImage but creates a difference in code between resources that have email actions and those that only have a list one. I do think the other refactor for the querying could do with a base class for handling the search functions however. If I do that over there then I could have new 'EmailAction' inherit from that to get the openstack query instance as well as implement these functions which would solve it?

DavidFair: So I would break up the action code so that the OpenstackQuery only has the email step, and the query happens in the action

Something like:

```
def query_and_email():
    results = query()
    super().email_users(results)
```
If that's possible?

joelvdavies: There is also the additional check of whether the parameters given to the query are valid e.g. if they need a project or not, but this check is only needed for emails and not the list actions. Also I thought the idea was we wanted a single function that could run the query and email in the lib folder and not rely on the actions?

DavidFair: So I'm happy for the actions layer to call two methods from the lib layer, as long as the implementation isn't in the actions layer

It's just separation of concerns is all

DavidFair (approved):
On discussion we're going to follow this up with the next PR(s) since it will develop merge conflict in the ongoing feature branch

PR #48, opened 2022-09-09
Add hypervisor list

joelvdavies:
### Description:

Updates the hypervisor.list action to bring it in line with the others and make it function again.

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:


lib/openstack_api/openstack_query.py
```diff
@@ -272,6 +272,16 @@ def get_user_prop(user_id, prop):
             }
         if object_type == "user":
             return {}
+        if object_type == "hypervisor":
+            return {
+                # Note: These parameters are depreciated, currently the openstack sdk is using a max microversion
+                # 2.88 meaning these will work but the 'uptime' parameter will not. Later on we will need
```
DavidFair, line 278: Since `uptime` is unavailable at our current SDK version does this mean we cannot query HV uptimes?
If not can we raise a `NotImplementedError` when a user tries to use that query

joelvdavies: Correct, it just returns None. Good idea, thanks.

lib/openstack_api/openstack_query.py
```diff
@@ -281,6 +285,11 @@ def get_user_prop(user_id, prop):
                 "vcpu_usage": lambda a: f"{a['vcpus_used']}/{a['vcpus']}",
                 "memory_mb_usage": lambda a: f"{a['memory_mb_used']}/{a['memory_mb']}",
                 "local_gb_usage": lambda a: f"{a['local_gb_used']}/{a['local_gb']}",
+                "uptime": lambda _: _raise(
```
DavidFair, line 288: Do we need the stubbed raise?

I'm pretty sure `"uptime": lambda _: raise NotImplementedError(...)` would work as lambda and functions are first class citizens and therefore equal?

joelvdavies: I tried that initially but pycharm doesn't like it, seems its not valid https://stackoverflow.com/questions/8294618/define-a-lambda-expression-that-raises-an-exception

DavidFair (approved):

PR #51, opened 2022-09-15
Refactor emails & queries

joelvdavies:
### Description:

Refactoring email and query code to place more of it in lib and reduce some code repetition.
Also addresses some outstanding comments on https://github.com/stfc/st2-cloud-pack/pull/46.

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:


lib/openstack_api/dataclasses.py
```diff
@@ -59,5 +59,19 @@ class EmailQueryParams:
...
+    :param query_preset: The query to use when searching
+    :param properties_to_select: The properties to select and display from the found resource
+    :param group_by: Property to group the results by - can be empty for no grouping
+    :param get_html: Whether the result should be in html format or not
+    """
+
+    query_preset: str
+    properties_to_select: List[str]
+    group_by: str
+    get_html: bool
```
DavidFair, line 77: Nit: maybe `return_html` flows better in my head, as `get_html` could be a method

lib/openstack_api/dataclasses.py
```diff
@@ -59,5 +59,19 @@ class EmailQueryParams:
...
+    """
+    Contains the data needed for running OpenstackQuery.search_resource
+    :param query_preset: The query to use when searching
+    :param properties_to_select: The properties to select and display from the found resource
+    :param group_by: Property to group the results by - can be empty for no grouping
+    :param get_html: Whether the result should be in html format or not
+    """
+
+    query_preset: str
+    properties_to_select: List[str]
```
DavidFair, line 75: If we have a list of known properties can we use an enum to constrain the list of available values. Currently we have to look at the implementation each time to see the supported params 
https://docs.python.org/3/library/enum.html#using-automatic-values

lib/openstack_api/openstack_floating_ip.py
```diff
@@ -1,21 +1,18 @@
...
 from openstack_api.openstack_identity import OpenstackIdentity
-from openstack_api.openstack_query import OpenstackQuery
-from openstack_api.openstack_wrapper_base import OpenstackWrapperBase
-from openstack_api.dataclasses import EmailQueryParams
-from structs.email_params import EmailParams
-from structs.smtp_account import SMTPAccount
 
 
-class OpenstackFloatingIP(OpenstackWrapperBase):
+class OpenstackFloatingIP(OpenstackQueryEmailBase):
```
DavidFair, line 15: Inherit from both classes rather instead of a single one

That should it's an action and it handles emails too

lib/openstack_api/openstack_query_base.py
```diff
@@ -0,0 +1,48 @@
+from abc import abstractmethod
+from typing import Dict, Callable, Any
+
+from openstack_api.dataclasses import QueryParams
+from openstack_api.openstack_connection import OpenstackConnection
+from openstack_api.openstack_query import OpenstackQuery
+from openstack_api.openstack_wrapper_base import OpenstackWrapperBase
+
+
+class OpenstackQueryBase(OpenstackWrapperBase):
```
DavidFair, line 10: This should be a plain class
It means we will have to init twice, but it fits the design pattern prefer composition over inheritance which (in the long-term) leads to cleaner code

Else someone changing the WrapperBase may indirectly affect the QueryBase despite them being "unrelated" apart from the connection_cls

tests/lib/test_openstack_hypervisor.py
```diff
@@ -1,10 +1,27 @@
...
 
+from nose.tools import raises
+
 from openstack_api.openstack_hypervisor import OpenstackHypervisor
 
+from tests.lib.test_openstack_query_base import OpenstackQueryBaseTests
+
+
+# pylint:disable=too-few-public-methods
+class _HypervisorMock:
```
DavidFair, line 12: This could be a dataclass AFAIK
As we can still override methods in DataClasses too

DavidFair (changes requested):

lib/openstack_api/dataclasses.py
joelvdavies: This is a bit tricky as this is a general dataclass for all resources. It would be possible to add them in the actions but it would be a long list that's tedious to update.

DavidFair: Ah I thought it was a fairly consistent subset with a few bespoke ones rather than bespoke to each class :+1:

DavidFair (approved):
Nice removal of duplication throughout our Query actions

Just to double check before I merge, has a subset of this been tested since the last commit? Simply testing one or two actions is fine for me.

joelvdavies:
@DavidFair Yes I didn't test every single query, but did check all of the list actions/emails worked this morning.

PR #52, opened 2022-09-15
Jupyter start stop servers

ElliottKasoar:
### Description:

Adds actions to start and stop default servers for Jupyter users, and tests, similar to those for creating and deleting users.

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:


DavidFair (changes requested):
Only two missing unit tests for response code, otherwise LGTM

DavidFair (approved):

PR #54, opened 2022-09-16
Create workflows for Jupyter user setup/deletion

ElliottKasoar:
### Description:

Depends on #52

Adds two workflows to combine Jupyter user creation/starting servers, and user deletion/stopping servers.

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/workflow.jupyter.user.cleanup.yaml
+++ b/actions/workflow.jupyter.user.cleanup.yaml
@@ -1,0 +1,27 @@
+name: workflow.jupyter.user.cleanup
+description: Stops servers and deletes a given list of JupyterHub users
+enabled: true
+entry_point: workflows/jupyter.user.cleanup.yaml
+runner_type: orquesta
+
+parameters:
+  jupyter_env:
+    description: The Jupyter environment to clean up users on
+    required: true
+    type: string
+    enum:
+      - prod
+      - dev
+      - training
+  user_base_name:
+    description: The users to clean up, or base name of the users to clean up without the hyphen (e.g. "jupyter")
+    required: true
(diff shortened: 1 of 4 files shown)


actions/workflows/jupyter.user.delete.yaml
```diff
@@ -0,0 +1,27 @@
+version: 1.0
+description: Deletes users on the FreeIPA server and the associated JupyterHub account
```
DavidFair, line 2: This description still mentions FreeIPA

actions/workflow.jupyter.user.setup.yaml
```diff
@@ -0,0 +1,27 @@
+name: workflow.jupyter.user.setup
```
DavidFair, line 1: Nit: could we call it `jupyter.user.prepare`, as this prepares the platform for a course
Setup is more tweaking the configuration in my mind.

This is a really pedantic nit even for me :sweat_smile:

actions/workflow.jupyter.user.delete.yaml
```diff
@@ -0,0 +1,27 @@
+name: workflow.jupyter.user.delete
```
DavidFair, line 1: Could we call this something like:
`workflow.jupyter.user.cleanup` to not confuse it with our delete action

DavidFair (commented):
Just some [bike-shedding](https://en.wikipedia.org/wiki/Law_of_triviality) points

DavidFair (approved):
LGTM, a much needed workflow :+1:

PR #64, opened 2022-10-21
Automate Hypervisor Updates

Ziggiyzoo:
### Description:

Automation of updating empty hypervisors.

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:


tests/lib/test_openstack_hypervisor.py
```diff
@@ -101,14 +102,38 @@ def test_search_all_hvs(self):
...
 
-        self.mocked_connection.assert_called_once_with("test")
+        result = self.instance.search_all_hvs(self.cloud_account)
 
-        self.api.list_hypervisors.assert_called_once_with(
-            filters={},
+        self.mocked_connection.assert_called_once_with(cloud_name=self.cloud_account)
+        self.assertEqual(result, self.api.list_hypervisors.return_value)
+
+    def test_get_all_empty_hvs_without_disabled_hvs(self):
```
DavidFair, line 111: We need tests that `vcpus_user > 0` and `running_vms > 0` correctly filter out those HVs instead of returning them as eligible

config.schema.yaml
```diff
@@ -17,6 +17,25 @@ jupyter:
       description: "Token for jupyter.stfc.ac.uk"
       required: false
 
+cloud_names:
+  description: Automated Updater Settings and tokens
```
meoflynn, line 21: This should be wrapped in " " as it's a String.

meoflynn: Maybe update the description to "Cloud name to use in automated jobs"

config.schema.yaml
```diff
@@ -17,6 +17,25 @@ jupyter:
...
+  type: "object"
+  required: true
+  properties:
+    dev_cloud:
+      type: "string"
+      description: "Cloud to use when running Automated Update Poll on Dev"
```
meoflynn, line 27: Change from "Automated Update Poll" to "Automated Jobs on Dev"

config.schema.yaml
```diff
@@ -17,6 +17,25 @@ jupyter:
...
+      required: false
+    prod_cloud:
+      type: "string"
+      description: "Cloud to use when running Automated Update Poll on Prod"
```
meoflynn, line 31: ^

lib/openstack_api/openstack_hypervisor.py
```diff
@@ -81,15 +80,34 @@ def _raise(ex):
             ),
         }
 
-    def search_all_hvs(self, cloud_account: str, **_) -> List[Hypervisor]:
+    def search_all_hvs(self, cloud_account: str) -> List[Hypervisor]:
```
meoflynn, line 83: Need to add back in `**_` as the OpenStack Query object may require this

lib/openstack_api/openstack_hypervisor.py
```diff
@@ -81,15 +80,34 @@ def _raise(ex):
...
+        """
+        hvs = self.search_all_hvs(cloud_account)
+        empty_hvs = []
+        for hpv in hvs:
+            hpv = dict(hpv)
+            if (
+                hpv["vcpus_used"] == 0
+                and hpv["running_vms"] == 0
+                and hpv["status"] == "enabled"
+            ):
```
meoflynn, line 107: This check is deprecated, we should replace this with the latest equivalent

sensors/src/hypervisor_polling_sensor.py
```diff
@@ -0,0 +1,69 @@
...
+
+        self._log.info("Hypervisor to update: " + str(empty_hvs))
+        self._log.info("Number of hypervisors possible to update: " + str(list_length))
+
+        # TODO - Check if the hypervisor needs updating (No point rebooting up to date hypervisors)
+        # hvs_to_update = self.api.check_hypervisor_version(self._cloud[self._environment], empty_hvs)
+
+        # Check compare the length of empty hvs list and the out of date hvs list
+        # if (len(hvs_to_update)/list_length) > 0.75:
+        #     hvs_to_update = hvs_to_update[0:int(list_length * (PERCENTAGE_TO_REBOOT / 100))]
```
meoflynn, line 58: Let's add this into a new issue

meoflynn (changes requested):

config.schema.yaml
```diff
@@ -17,6 +17,25 @@ jupyter:
...
+      description: "Cloud to use when running Automated Jobs on Dev"
+      required: false
+    prod_cloud:
+      type: "string"
+      description: "Cloud to use when running Automated Jobs on Prod"
+      required: false
+    selected_env:
+      type: "string"
+      description: "Cloud to use when running Automated Jobs on Dev"
+      required: true
```
meoflynn, line 36: `dev_cloud` and `selected_env` have the same descriptions. Are these duplicate properties?

Maybe update the description for `selected_env` to separate it from the `dev_cloud` property?

meoflynn (changes requested):

DavidFair:
@anish-mudaraddi Are you happy to adopt this?

anish-mudaraddi:
i'm closing this - I've looked over the code and there isn't anything useful here
It's adding a Sensor for polling hypervisors but this should really be an Action that is run periodically by a Cron job. 

A Sensor is for interacting with external services - this pack is built to run Openstack commands so we shouldn't need a sensor to query hypervisors - we should just create an Action the queries for hypervisors which are empty (using the query library) that is run periodically. 

This has the added benefit of allowing us to specify the cloud domain and other parts we can fine tune easier via writing one or more triggers.

This pack doesn't implement any rebooting either - I think the idea was to dispatch a trigger called `openstack.reboot_hypervisors` but it doesn't do that at the moment and there's no action that can handle that yet. There aren't any defined Rules for this.

I'll work on this in the new year

PR #67, opened 2022-11-17
Flavor migration and unit tests

meoflynn:
### Description:

Python Script in the Lib layer for OpenStack Flavors for migrating flavors between Clouds. 

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/openstack_api/openstack_flavor.py
+++ b/lib/openstack_api/openstack_flavor.py
@@ -31,24 +31,18 @@ class OpenstackFlavor(OpenstackWrapperBase):
         :param dest_cloud: Cloud account for destination cloud
         :return: List of names of missing flavors or empty list
         """
-        # list flavors from prod
+        # list flavors from source cloud
         source_flavors = self.list_flavor(source_cloud)
 
-        # list flavors from dev
+        # list flavors from destination cloud
         dest_flavors = self.list_flavor(dest_cloud)
 
-        # Prepare list of flavors that are in prod and dev
-        source_flavor_names = []
-        dest_flavor_names = []
+        # Prepare set of flavors in source and destination clouds
+        source_flavor_names = {source.name for source in source_flavors}
+        dest_flavor_names = {dest.name for dest in dest_flavors}
 
-        for source in source_flavors:
-            source_flavor_names.append(source.name)
-
-        for dest in dest_flavors:
-            dest_flavor_names.append(dest.name)
-
-        # Find set of flavor names which are in prod but are missing from dev
-        missing_flavors = set(source_flavor_names).difference(set(dest_flavor_names))
+        # Find set of flavor names which are in source cloud but missing in the destination cloud
+        missing_flavors = source_flavor_names.difference(dest_flavor_names)
 
         return list(missing_flavors)
 
--- a/lib/openstack_api/openstack_flavor.py
+++ b/lib/openstack_api/openstack_flavor.py
@@ -121,7 +121,7 @@ class OpenstackFlavor(OpenstackWrapperBase):
         missing_flavors = self.get_missing_flavors(source_cloud, dest_cloud)
         # if there are missing flavors in the list i.e. missing_flavors is not an empty list
         if not missing_flavors:
-            return
+            return None
 
         for flavor_name in missing_flavors:
             flavor_data = self.get_flavor(source_cloud, flavor_name)
--- a/lib/openstack_api/openstack_flavor.py
+++ b/lib/openstack_api/openstack_flavor.py
@@ -1,0 +1,132 @@
+from typing import List, Dict, Union, Optional
+from openstack.compute.v2.flavor import Flavor
+from openstack_api.openstack_wrapper_base import OpenstackWrapperBase
+
+
+class OpenstackFlavor(OpenstackWrapperBase):
+    def list_flavor(self, cloud_account: str) -> List[Flavor]:
+        """
+        Get list of flavors from either production or development cloud
+        :param cloud_account: Name of cloud to use
+        :return: List of Flavor Objects
+        """
+        with self._connection_cls(cloud_account) as conn:
+            return conn.list_flavors()
+
+    def get_flavor(self, cloud_account: str, flavor_name: str) -> Flavor:
+        """
+        Get information for a specific flavor
+        :param cloud_account: Name of cloud to use
+        :param flavor_name: name of the flavor to get
+        :return: Flavor object or None if unsuccessful
+        """
+        with self._connection_cls(cloud_account) as conn:
+            return conn.get_flavor(flavor_name)
+
+    def get_missing_flavors(self, source_cloud: str, dest_cloud: str) -> List[str]:
+        """
+        Get list of flavors missing in Dev cloud based on the flavors currently in
+        production
+        :param source_cloud: Cloud account for source cloud
+        :param dest_cloud: Cloud account for destination cloud
+        :return: List of names of missing flavors or empty list
+        """
+        # list flavors from prod
+        source_flavors = self.list_flavor(source_cloud)
+
+        # list flavors from dev
+        dest_flavors = self.list_flavor(dest_cloud)
+
+        # Prepare list of flavors that are in prod and dev
+        source_flavor_names = []
+        dest_flavor_names = []
+
+        for source in source_flavors:
+            source_flavor_names.append(source.name)
+
+        for dest in dest_flavors:
+            dest_flavor_names.append(dest.name)
+
+        # Find set of flavor names which are in prod but are missing from dev
+        missing_flavors = set(source_flavor_names).difference(set(dest_flavor_names))
+
+        return list(missing_flavors)
+
+    def create_flavor(self, cloud_account, flavor_data) -> Flavor:
+        """
+        Creates a new flavor
+        :param cloud_account: connection to cloud - should be dev cloud
+        :param flavor_data: Flavor object from production that we want to create in dev
+        :return: Flavor object or None if unsuccessful
+        """
+
+        if (
+            flavor_data.swap == ""
+        ):  # openstack stores swap=0 as swap='' when getting flavor information
+            flavor_swap = 0
+        else:
+            flavor_swap = flavor_data.swap
+
+        with self._connection_cls(cloud_account) as conn:
+            return conn.create_flavor(
+                name=flavor_data.name,
+                ram=flavor_data.ram,
+                vcpus=flavor_data.vcpus,
+                disk=flavor_data.disk,
+                flavorid="auto",
+                ephemeral=flavor_data.ephemeral,
+                swap=flavor_swap,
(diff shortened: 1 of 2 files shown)


lib/openstack_api/openstack_flavor.py
```diff
@@ -0,0 +1,127 @@
...
+        :returns extra_specs: Dictionary containing flavor metadata
+        """
+
+        if isinstance(flavor_id, str):
+            flavor = self.get_flavor(cloud_account, flavor_id)
+            extra_specs = flavor.extra_specs
+        elif isinstance(flavor_id, Flavor):
+            extra_specs = flavor_id.extra_specs
+        else:
+            extra_specs = {}
```
DavidFair, line 99: Could we change this to:

```
        if isinstance(flavor_id, str):
            flavor = self.get_flavor(cloud_account, flavor_id)
            extra_specs = flavor.extra_specs
        else:
            extra_specs = flavor_id.extra_specs
```

Since we should only get a flavor or string type here, so the `else {}` is a subtle bug waiting to happen

lib/openstack_api/openstack_flavor.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def set_flavor_specs(self, cloud_account: str, flavor_id: str, extra_specs: dict):
+        """
+        Sets extra specs for a flavor
+        :param cloud_account: connection to the cloud
+        :param flavor_id: ID of the flavor to add metadata to
+        :param extra_specs: Dictionary of extra specs for the flavor
+        """
+        with self._connection_cls(cloud_account) as conn:
+            if extra_specs != {}:
```
DavidFair, line 111: We can rely on Python truthy types:
```
if extra_specs:
   conn.set_flavor_specs(...)
```

lib/openstack_api/openstack_flavor.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def migrate_flavors(self, source_cloud: str, dest_cloud: str):
+        """
+        Migrates flavors from a source cloud to a destination cloud
+        :param source_cloud: Cloud account used as the source for flavors
+        :param dest_cloud: Cloud account for the destination for flavors
+        """
+        missing_flavors = self.get_missing_flavors(source_cloud, dest_cloud)
+        # if there are missing flavors in the list i.e. missing_flavors is not an empty list
+        if missing_flavors:
```
DavidFair, line 122: Can we change this so the main branch isn't nested in the if-condition:

```python
if not missing_flavors:
    return

for flavor_name in missing_flavors:
    ....
```

tests/lib/test_openstack_flavor.py
```diff
@@ -0,0 +1,140 @@
...
+        mocked_cloud_name = NonCallableMock()
+        mocked_flavor_name = NonCallableMock()
+        flavor = self.instance.get_flavor(mocked_cloud_name, mocked_flavor_name)
+
+        assert flavor == self.api.get_flavor.return_value
+
+        self.api.get_flavor.assert_called_once_with(mocked_flavor_name)
+        self.mocked_connection.assert_called_once_with(mocked_cloud_name)
+
+    def test_get_missing_flavors_calls_list_flavor(self):
```
DavidFair, line 41: `test_get_missing_flavors_calls_list_flavor_with_correct_clouds`
And the docstring is incorrect, as this test is checking we pass the source and dest cloud in the correct order rather than checking the diff in the next test

tests/lib/test_openstack_flavor.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def test_get_specs(self):
+        """
+        Test getting the extra specs from a flavor
+        """
+        mocked_cloud_name = NonCallableMock()
+        mocked_flavor = NonCallableMock()
+
+        self.instance.get_flavor(mocked_cloud_name, mocked_flavor)
+        self.instance.get_flavor_specs(mocked_cloud_name, mocked_flavor)
```
DavidFair, line 115: This test has no asserts, so it's not testing anything at the moment?

tests/lib/test_openstack_flavor.py
```diff
@@ -0,0 +1,140 @@
...
+    def test_set_specs(self):
+        """
+        Test setting the extra specs for a flavor
+        """
+        mocked_cloud_name = NonCallableMock()
+        flavor_id = NonCallableMock()
+        extra_specs = NonCallableMock()
+
+        self.instance.set_flavor_specs(mocked_cloud_name, flavor_id, extra_specs)
+        self.mocked_connection.assert_called_once_with(mocked_cloud_name)
```
DavidFair, line 126: Ditto, we're checking the connection opens but not the set_specs method is called

tests/lib/test_openstack_flavor.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        Test that the migrate_flavors method behaves as expected
+        """
+        mocked_source_cloud = NonCallableMock()
+        mocked_dest_cloud = NonCallableMock()
+        source_flavor = Mock()
+        dest_flavor = self.instance.create_flavor(mocked_dest_cloud, source_flavor)
+        assert dest_flavor == self.api.create_flavor.return_value
+        self.instance.get_flavor_specs(mocked_source_cloud, dest_flavor.id)
+        self.instance.set_flavor_specs(
```
DavidFair, line 138: Are these missing `.assert_called_once_with`? as it's making the method calls manually 

Also it doesn't look like `migrate_flavor` is actually called in this test?

DavidFair (changes requested):

lib/openstack_api/openstack_flavor.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def get_missing_flavors(self, source_cloud: str, dest_cloud: str) -> List[str]:
+        """
+        Get list of flavors missing in Dev cloud based on the flavors currently in
+        production
+        :param source_cloud: Cloud account for source cloud
+        :param dest_cloud: Cloud account for destination cloud
+        :return: List of names of missing flavors or empty list
+        """
+        # list flavors from prod
```
DavidFair, line 34: Nit: this should say from source cloud

lib/openstack_api/openstack_flavor.py
```diff
@@ -0,0 +1,132 @@
...
+        source_flavors = self.list_flavor(source_cloud)
+
+        # list flavors from dev
```
DavidFair, line 37: ^

lib/openstack_api/openstack_flavor.py
```diff
@@ -0,0 +1,132 @@
...
+        dest_flavors = self.list_flavor(dest_cloud)
+
+        # Prepare list of flavors that are in prod and dev
+        source_flavor_names = []
```
DavidFair, line 41: We can remove the two for loops below and use list comprehension instead 
```
source_flavor_names = [source.name for source in source_flavors]
dest_flavor_names = [dest.name for dest in dest_flavors]
```

Also we can go straight to a set rather than doing list -> set
```
source_flavor_names = {source.name for source in source_flavors}
dest_flavor_names = {dest.name for dest in dest_flavors}
```

This allows us to get rid of the `set(source_flavor_names)` and `set(dest_flavor_names)` casts below

lib/openstack_api/openstack_flavor.py
```diff
@@ -0,0 +1,132 @@
...
+    def create_flavor(self, cloud_account, flavor_data) -> Flavor:
+        """
+        Creates a new flavor
+        :param cloud_account: connection to cloud - should be dev cloud
+        :param flavor_data: Flavor object from production that we want to create in dev
+        :return: Flavor object or None if unsuccessful
+        """
+
+        if (
+            flavor_data.swap == ""
```
DavidFair, line 64: Rely on the falsey type check for this:
```
if (flavor_data.swap):
   flavor_swap = flavor_data.swap
else:
   flavor_swap = 0
```

tests/lib/test_openstack_flavor.py
```diff
@@ -0,0 +1,211 @@
...
+        self.api = self.mocked_connection.return_value.__enter__.return_value
+
+    def test_list_flavor(self):
+        """
+        Test that we can retrieve a list of flavors from a specific cloud
+        """
+        mocked_name = NonCallableMock()
+        flavors = self.instance.list_flavor(mocked_name)
+        assert flavors == self.api.list_flavors.return_value
+        self.mocked_connection.assert_called_once_with(mocked_name)
```
DavidFair, line 26: Nit we don't check `self.api.list_flavors.assert_called_once_with()`

tests/lib/test_openstack_flavor.py
```diff
@@ -0,0 +1,211 @@
...
+        for i, obj in enumerate(mock_dest_names):
+            obj.name = i + 2
+
+        mock_side_effect = [mock_source_names, mock_dest_names]
+
+        self.instance.list_flavor = Mock(side_effect=mock_side_effect)
+        return_diff = self.instance.get_missing_flavors(
+            NonCallableMock(), NonCallableMock()
+        )
+        assert [0, 1] == return_diff
```
DavidFair, line 73: A small comment here would help visualise why 0, 1. Something alone the lines of 
```
# We expect 0, 1 since the first set contains 0:3 and the second set contains 2:5
# so 0, 1 are the elements missing in set two
```

Feel free to reword this

tests/lib/test_openstack_flavor.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def test_empty_dict(self):
+        """
+        Test that method set_flavor_specs not called when empty_dictionary passed
+        """
+        empty_flavor_spec = {}
+        mocked_cloud_name = NonCallableMock()
+        flavor_id = NonCallableMock()
+        self.instance.set_flavor_specs(mocked_cloud_name, flavor_id, empty_flavor_spec)
+        self.api.set_flavor_specs.assert_not_called()
```
DavidFair, line 105: :+1: I didn't think of this case

tests/lib/test_openstack_flavor.py
```diff
@@ -0,0 +1,211 @@
...
+        # mock return value
+        missing_flavors = NonCallableMock()
+        # mock functions
+        self.instance.get_missing_flavors = Mock()
+        self.instance.get_flavor = Mock()
+        # prepare return value
+        self.instance.get_missing_flavors.return_value = missing_flavors
+        # call method
+        self.instance.get_missing_flavors(mocked_source_cloud, mocked_dest_cloud)
+        # assert at least one function that is called by method
```
DavidFair, line 164: We could trim out these comments, since we use the same pattern throughout the code base.
If you'd like to preserve them, maybe a developer section in the README would be a more convenient place to put an example?

DavidFair (changes requested):

DavidFair (approved):

PR #69, opened 2022-11-18
Actions for migrating Flavors

meoflynn:
### Description:
Python script and unit test in action layer for migrating OpenStack Flavors.

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:


lib/openstack_api/openstack_flavor.py
```diff
@@ -114,6 +113,7 @@ def migrate_flavors(
         """
         missing_flavors = self.get_missing_flavors(source_cloud, dest_cloud)
         # if there are missing flavors in the list i.e. missing_flavors is not an empty list
+        print(missing_flavors)
```
DavidFair, line 116: This is leftover from testing and could be removed

DavidFair (changes requested):

DavidFair (approved):

Issue #71, opened 2023-05-23
Improve Openstack Querying 


anish-mudaraddi:
Improve the Querying Mechanism currently provided. 

I recommend Splitting out the Querying functionality into a separate module on the lib layer

It should have the project structure as follows:

```
lib/openstack_query
    query_wrapper.py
    query_utils.py
    
    server
        query_server.py
        
    hv
        query_hv.py
     
    project
        query_project.py
    
    ..etc...
       
```

`query_wrapper.py`: A base class for all `query_<resource>.py` files. 
It should expose the instantiate the following methods that each 

```
# following functions are used to set the query parameters
select ( *properties )
select_all ( )
where ( *tuple(preset, args)  )
sort_by (  tuple(sort_by_preset, args)  )
group_by (  tuple(group_by_preset, args)  )


# used to run the function
run ( cloud_account, **kwargs )

# get outputs of function
to_list()   # get list of outputs matching query
to_str()   # get string table output 
to_html() # get html table output

```

each `query_<resource>` function must implement:

```
from_subset( list(resource_objects) or list(resource_identifiers) )
from_all( )
```

This way a query can be created like so: 
```
query = SeverQuery(cloud_account).select( prop1, prop2, prop3 ).where( preset1 ).from_all() 
query.run()
out = query.to_str()
```

queries can be easily combined like so:
```
query1 = ServerQuery(cloud_account).select('user_ids').where(STATUS_DISABLED).from_all()
query1.run()
selected_users = query1.to_list()

query2 = UserQuery(cloud_account).select('user_ids', 'email').from_subset(selected_users)
query2.run()
selected_hvs = query2.to_list()
```

anish-mudaraddi:
Currently, the query mechanism works by querying for all instances of a given resource and then filtering using Python

We can improve the performance of our queries by providing filter kwargs to Openstack APIs like the compute API 

`conn.compute.list_servers()` can take a number of filtering parameters as shown here https://docs.openstack.org/api-ref/compute/?expanded=list-servers-detail#list-servers

This does not cover all cases so filter functions may still be needed but we should rework our mechanism to make use of these wherever possible.

anish-mudaraddi:
If we want to utilise the filter kwargs, we will have to reconsider how the query mechanism is designed

Currently: 
We perform one query and iteratively select for values that pass each filter given.

For some cases this is not possible using filter kwargs. 
e.g.: Performing a query for servers which are shutoff or errored. 


Proposal:
[...]

PR #72, opened 2023-06-08
Openstack Query API - Base

anish-mudaraddi:
### Description:

This adds Openstack Query API + Server Querying functionality
follows guidelines outlined in #71  

Still TODO
* [x] Docstrings for classes and functions
* [x] Integration tests (deferred for a later PR)

### Submitter:

Have you (where applicable):

* [x] Added unit tests?
* [ ] Checked the latest commit runs on Dev? (Not yet)
* [ ] Updated the example config file(s) and README? (Not yet)

---

Once this gets merged we can setup future PRs:
    - Stackstorm Integration 
    - Query Hypervisors/Project/Users etc

### Reviewer

Does this PR:

* [x] Have unit tests for the action/sensor and `lib` layers? (no actions ported yet)


lib/openstack_query/query_server.py
```diff
@@ -0,0 +1,144 @@
...
+                projects.append(project)
+
+        for project in projects:
+            servers.extend(
+                QueryServer._run_query_on_project(conn, project, filter_kwargs)
+            )
+        return servers
+
+    @staticmethod
+    def _run_query_on_project(
```
DavidFair, line 122: Should this be called something like `_run_query_on_all_projects`, since the `all_tenants` implies that to an extent?
This would only be runnable by admins because of the all_tenants flag

lib/openstack_query/query_server.py
```diff
@@ -0,0 +1,144 @@
...
+
+        return servers
+
+    @staticmethod
+    def _parse_subset(conn: OpenstackConnection, server_identifiers: Server):
+        query_from = []
+        for server_instance in server_identifiers:
+            server_obj = server_instance
+            if not isinstance(Server, server_instance):
+                server_obj = conn.compute.find_server(server_instance)
```
DavidFair, line 142: As me and Ramzi discovered, this is a really really expensive operation in Openstack
It's (ironically) faster to grab all the servers from a project and filter project side, than to do the auth flow n times for each machine

lib/openstack_query/query_wrapper.py
```diff
@@ -0,0 +1,353 @@
...
+    prop_not_equal_to,
+    prop_greater_than,
+    prop_less_than,
+    prop_younger_than,
+    prop_older_than,
+    prop_older_than_or_equal_to,
+    prop_younger_than_or_equal_to,
+    prop_any_in,
+    prop_not_any_in,
+    prop_matches_regex,
```
DavidFair, line 19: Ditto about another class for these too. We could further split these down into

`QueryPropDate` # For younger or older...etc.
`QueryPropSets`  # For any in or not in 
`QueryPropComparisons`  # For equal / NE / regex

lib/openstack_query/query_wrapper.py
```diff
@@ -0,0 +1,353 @@
...
+        QueryPresets.GREATER_THAN: prop_greater_than,
+        QueryPresets.LESS_THAN: prop_less_than,
+        QueryPresets.YOUNGER_THAN: prop_younger_than,
+        QueryPresets.YOUNGER_THAN_OR_EQUAL_TO: prop_younger_than_or_equal_to,
+        QueryPresets.OLDER_THAN_OR_EQUAL_TO: prop_older_than_or_equal_to,
+        QueryPresets.OLDER_THAN: prop_older_than,
+        QueryPresets.ANY_IN: prop_any_in,
+        QueryPresets.NOT_ANY_IN: prop_not_any_in,
+        QueryPresets.MATCHES_REGEX: prop_matches_regex,
+    }
```
DavidFair, line 47: I was wondering about such a hierarchy based on the above:

```python
class QueryPropBase(abc.ABC):
   @abstractmethod
   def is_supported_query(preset: QueryPresets)

class QueryPropComparison(QueryPropBase):
    __FILTER_FUNCTIONS = {
        QueryPresets.EQUAL_TO: prop_equal_to,
        QueryPresets.NOT_EQUAL_TO: prop_not_equal_to,
        # ...
    }

   def is_supported_query(preset: QueryPresets):
       return preset in _FILTER_FUNCTIONS.keys()
```

Then this class can simply do:
```
def __init__(...):
  self._prop_handlers: List[QueryBase] = [QueryPropDate(), QueryPropComparisons()]

def get_handler(query: QueryPreset) -> QueryBase:
   for i in self._prop_handlers:
       if i.is_supported_query(query):
          return i
```

Then the rest of your code can simply duck type, since it knows that prop class has to support it's query

This will massively curtail / reduce the size of this file and it's responsibilities as it's already verging into a god class

lib/openstack_query/query_wrapper.py
```diff
@@ -0,0 +1,353 @@
...
+        """
+        if prop not in self._PROPERTY_MAPPINGS.keys():
+            raise QueryMappingError(
+                """
+                Error: failed to get property mapping, the property is valid
+                but does not contain an entry in PROPERTY_MAPPINGS.
+                Please raise an issue with repo maintainer
+            """
+            )
+        return prop.name.lower(), self._PROPERTY_MAPPINGS[prop]
```
DavidFair, line 85: Nit: use `.casefold()` instead of lower when doing case insensitive comparisons on Python 3+

lib/openstack_query/query_wrapper.py
```diff
@@ -0,0 +1,353 @@
...
+    ) -> List[Any]:
+        """
+        This method is to be instantiated in subclasses of this class to run openstack query command
+        """
+        raise NotImplementedError(
+            """
+            This static method should be implemented in subclasses of QueryWrapper to
+            run the appropriate openstack command(s)
+        """
+        )
```
DavidFair, line 315: Anything related to not building the query, e.g. running it, should be split out of this class into a dedicated runner instead even if this is an abstract method

Also this should be annotated with `@abstractmethod`

lib/openstack_query/server_manager.py
```diff
@@ -0,0 +1,167 @@
...
+    )
+
+
+def search_servers_older_than(
+    cloud_account: str,
+    query_details: QueryDetails,
+    days: int,
+    hours: int = 0,
+    minutes: int = 0,
+    seconds: int = 0,
```
DavidFair, line 30: Do these defaults imply midnight just to double check?

lib/openstack_query/server_manager.py
```diff
@@ -0,0 +1,167 @@
...
+):
+    return manager_utils.run_query(
+        cloud_account,
+        QueryServer(),
```
DavidFair, line 34: Having seen this used a few times, I think that the `manager_utils` should have a class which takes the `cloud_account` and `QueryX` as part of the construction

As a user I might want to do something akin to:

```
def find_between_interval():

manager = Foo(cloud_account, QueryServer())
preset_older = PresetDetails(
            preset=QueryPresets.OLDER_THAN,
            prop=ServerProperties.SERVER_CREATION_DATE,
           # ...etc.
)

older = manager.run_query(preset_older, ....)

preset_newer = PresetDetails(
            preset=QueryPresets.NEWER_THAN,
            prop=ServerProperties.SERVER_CREATION_DATE,
           # ...etc.
)

newer = manager.run_query(preset_newer)

return set(older) - set(newer)
```

That way you separate query state from a connection state into two calls effectively

DavidFair (changes requested):
Partial review, feel free to go back and forth on this one

lib/openstack_query/manager_utils.py
```diff
@@ -0,0 +1,54 @@
...
+
+def build_and_run_query(
+    cloud_account: str,
+    query_obj: OpenstackQueryWrapper,
+    preset: Optional[PresetDetails] = None,
+    properties_to_select: Optional[Set[ServerProperties]] = None,
+    group_by: Optional[ServerProperties] = None,
+    sort_by: Optional[ServerProperties] = None,
+    output_type: QueryOutputOption = QueryOutputOption.TO_STR,
+) -> Union[str, List[Any]]:
```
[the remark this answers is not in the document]
anish-mudaraddi: it returns either a list of openstack resources or a string - which is a table of the results - hence the List[Any].
I agree with the type alias - will do that

lib/openstack_query/query_server.py
anish-mudaraddi: no we're filtering by project_id - the `all_tenants` forces the search context to search in all projects (just in case that's needed) - this was copied from the previous method that was used

lib/openstack_query/query_server.py
anish-mudaraddi: sure that makes sense :+1:

lib/openstack_query/query_wrapper.py
anish-mudaraddi: this makes sense thanks :+1:

anish-mudaraddi: in your example we could put `prop_equal_to` and `prop_not_equal_to` as static methods in `QueryPropComparison` class

lib/openstack_query/query_wrapper.py
```diff
@@ -0,0 +1,353 @@
...
+        :param properties_dict: dict of query results
+        :param return_html: True if output required in html table format else output plain text table
+        :param kwargs: kwargs to pass to tabulate
+        :return: String (html or plaintext table of results)
+        """
+        headers = list(properties_dict[0].keys())
+        rows = [list(row.values()) for row in properties_dict]
+        return tabulate(
+            rows, headers, tablefmt="html" if return_html else "grid", **kwargs
+        )
```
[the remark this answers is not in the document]
anish-mudaraddi: How would this work? Make a separate class like `QueryOutput` and set `QueryWrapper` to inherit it?

lib/openstack_query/server_manager.py
anish-mudaraddi: no these variables are used to define a time - days, hours, minutes, seconds since the current time

anish-mudaraddi: it's poorly defined - I guess I should allow providing a timestamp instead

lib/openstack_query/manager_utils.py
```diff
@@ -0,0 +1,54 @@
...
+from openstack_query.openstack_query_wrapper import OpenstackQueryWrapper
+
+
+def build_and_run_query(
+    cloud_account: str,
+    query_obj: OpenstackQueryWrapper,
+    preset: Optional[PresetDetails] = None,
+    properties_to_select: Optional[Set[ServerProperties]] = None,
+    group_by: Optional[ServerProperties] = None,
+    sort_by: Optional[ServerProperties] = None,
```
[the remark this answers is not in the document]
anish-mudaraddi: I like having a generic function like this so I can reuse it with every server_manager function - this means that each function only really needs to change the preset details 

I can refactor the code to better utilise dataclasses so theres less args. 
Most of these args will be passed as-is so I think using a dataclass here is appropriate

lib/enums/query/query_presets.py
```diff
@@ -0,0 +1,22 @@
+from enum import Enum, auto
+
+
+class QueryPresets(Enum):
```
meoflynn, line 4: Could we add a docstring to describe what this class does?

lib/enums/query/server_properties.py
```diff
@@ -0,0 +1,15 @@
+from enum import Enum, auto
+
+
+class ServerProperties(Enum):
```
meoflynn, line 4: Could we add a docstring to describe what this class does?

lib/openstack_api/openstack_query.py
```diff
@@ -19,7 +19,6 @@
 
 
 class OpenstackQuery(OpenstackWrapperBase):
```
meoflynn, line 21: Could we add a docstring to describe what this class does?

lib/openstack_query/manager_utils.py
```diff
@@ -0,0 +1,54 @@
...
+    output_type: QueryOutputOption = QueryOutputOption.TO_STR,
+) -> Union[str, List[Any]]:
+    query_obj = populate_query(query_obj, preset, properties_to_select)
```
meoflynn, line 21: Could we add a docstring to describe what this method does?

lib/openstack_query/manager_utils.py
```diff
@@ -0,0 +1,54 @@
...
+
+    out = get_query_output(query_obj, output_type)
+    return out
+
+
+def populate_query(
+    query_obj: OpenstackQueryWrapper,
+    preset_details: Optional[PresetDetails] = None,
+    properties_to_select: Optional[Set[ServerProperties]] = None,
+) -> QueryServer:
```
meoflynn, line 34: Could we add a docstring to describe what this method does?

lib/openstack_query/query_server.py
```diff
@@ -0,0 +1,144 @@
...
+from openstack_api.openstack_connection import OpenstackConnection
+from openstack_query.query_wrapper import QueryWrapper
+from openstack_query.utils import convert_to_timestamp
+
+from enums.query.server_properties import ServerProperties
+from enums.query.query_presets import QueryPresets
+from exceptions.parse_query_error import ParseQueryError
+
+
+class QueryServer(QueryWrapper):
```
meoflynn, line 15: Could we add a docstring to describe what this class does as well?

lib/openstack_query/manager_utils.py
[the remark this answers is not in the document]
anish-mudaraddi: Added type aliases for most things - see `custom_types/openstack_query/aliases.py`

lib/enums/query/server_properties.py
```diff
@@ -0,0 +1,15 @@
...
+    HYPERVISOR_ID = auto()
+    SERVER_ID = auto()
+    SERVER_NAME = auto()
+    SERVER_DESCRIPTION = auto()
+    SERVER_STATUS = auto()
+    SERVER_CREATION_DATE = auto()
+    SERVER_LAST_UPDATED_DATE = auto()
+    FLAVOR_ID = auto()
+    IMAGE_ID = auto()
+    PROJECT_ID = auto()
```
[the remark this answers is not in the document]
anish-mudaraddi: done

lib/openstack_query/query_server.py
anish-mudaraddi: For now the parse_subset function simply accepts whatever you give it - and only accepts openstack resource objects. 

In the docs we should push people to only use this to chain multiple queries

lib/openstack_query/query_wrapper.py
```diff
@@ -0,0 +1,353 @@
+from enum import Enum
+from typing import Tuple, Callable, Any, Dict, Optional, List, Union
+from openstack_api.openstack_wrapper_base import OpenstackWrapperBase
+from openstack_api.openstack_connection import OpenstackConnection
+
+from openstack_query.utils import (
+    check_filter_func,
+    check_kwarg_mapping,
```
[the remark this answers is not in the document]
anish-mudaraddi: I created handlers for presets, kwargs and props to validate and return appropriate values

lib/openstack_query/query_wrapper.py
anish-mudaraddi: Split out query wrapper into builder, runner and output - query_wrapper no instantiates these classes and coordinates on requests to them

QueryWrapper now only maintains a set of public methods which we expose to the users as our api

lib/openstack_query/query_wrapper.py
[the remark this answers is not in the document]
anish-mudaraddi: done - see above

lib/openstack_query/server_manager.py
anish-mudaraddi: I left this as is for now - best if we make an issue for this for a later PR since this one is big enough as is

lib/openstack_query/server_manager.py
anish-mudaraddi: refactored `manager_utils` into class `QueryManager`

lib/openstack_query/query_server.py
```diff
@@ -0,0 +1,144 @@
...
+    _PROPERTY_MAPPINGS = {
```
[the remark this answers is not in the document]
anish-mudaraddi: done

lib/openstack_api/openstack_query.py
anish-mudaraddi: this class is the deprecated class - when the updated openstack_query stuff is approved I will remove this

DavidFair (changes requested):
Partial review

lib/openstack_query/handlers/presets/preset_handler_base.py
```diff
@@ -0,0 +1,154 @@
...
+        preset and that the preset supports the given property
+        :param preset: A QueryPreset Enum for which a filter function mapping may exist for
+        :param prop: A property Enum for which a filter function mapping may exist for
+        """
+        if preset not in self._FILTER_FUNCTIONS.keys():
+            return False
+
+        props_valid_for_preset = self._FILTER_FUNCTION_MAPPINGS.get(preset, None)
+        if props_valid_for_preset:
+            if prop in props_valid_for_preset:
```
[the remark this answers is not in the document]
anish-mudaraddi: there's another if statement under this one so there's some duplication doing it this way. If you don't like nested ifs I can do 

```
if not props_valid_for_preset:
    return False
if prop in props_valid_for_preset:
    return True
if ["*"] == self._FILTER_FUNCTION_MAPPINGS[preset]:
    return True
return False
```

lib/openstack_query/handlers/presets/preset_handler_base.py
```diff
@@ -0,0 +1,154 @@
...
+                f"Preset Argument Error: failed to build filter_function for preset '{preset.name}'"
+            ) from err
+
+        return lambda a: self._filter_func_wrapper(
+            a, filter_func, prop_func, filter_func_kwargs
+        )
+
+    @staticmethod
+    def _filter_func_wrapper(
+        item: Any,
```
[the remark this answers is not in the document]
anish-mudaraddi: I guess - should the alias just be 
`AnyOpenstackObj = Any`

I could create an Optional Alias like
`AnyOpenstackObj = Optional[Server, Hypervisor, ...] `

which one should I do - the latter seems sensible but it's another place to edit when we want to add another query-able resource

lib/openstack_query/handlers/presets/preset_handler_base.py
```diff
@@ -0,0 +1,154 @@
...
+        has_varargs = any(param.kind == param.VAR_POSITIONAL for param in parameters)
+        has_varkwargs = any(param.kind == param.VAR_KEYWORD for param in parameters)
+
+        for param in parameters:
+            if param.kind == param.POSITIONAL_OR_KEYWORD:
+                param_name = param.name
+                if param_name in func_kwargs:
+                    kwargs_value = func_kwargs[param_name]
+                    param_type = param.annotation
+                    if not isinstance(kwargs_value, param_type):
```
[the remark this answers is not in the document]
anish-mudaraddi: I agree we should fail noisily - however in a previous comment you mentioned for `check_kwarg_mapping` method you mentioned that it should return True/False and that False is okay 

I think we should check types and fail fast - otherwise the user will receive an error only after the query is run. In this case after the openstacksdk command is run - which could take 2-3 minutes.

lib/openstack_query/handlers/presets/preset_handler_datetime.py
```diff
@@ -0,0 +1,120 @@
...
+        """
+        Filter function which returns True if property older than a relative amount of time since current time
+        :param prop: prop value to check against
+        :param days: (Optional) relative number of days since current time to compare against
+        :param hours: (Optional) relative number of hours since current time to compare against
+        :param minutes: (Optional) relative number of minutes since current time to compare against
+        :param seconds: (Optional) relative number of seconds since current time to compare against
+        :param prop_timestamp_fmt: (Optional) timestamp format of prop value (default: yyyy-MM-ddTHH:mm:ssZ)
+        """
+        prop_timestamp = datetime.strptime(prop, prop_timestamp_fmt).timestamp()
```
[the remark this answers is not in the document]
anish-mudaraddi: I can add timezone as another parameter to each of these functions - by default I believe openstacksdk assumes UTC time zone - hence I can just convert to UTC in `get_timestamp_in_seconds` function

lib/openstack_query/handlers/presets/preset_handler_generic.py
```diff
@@ -0,0 +1,48 @@
...
+            QueryPresetsGeneric.NOT_EQUAL_TO: self._prop_not_equal_to,
+        }
+
+    def _prop_not_equal_to(self, prop: Any, value: Any) -> bool:
+        """
+        Filter function which returns true if a prop is not equal to a given value
+        :param prop: prop value to check against
+        :param value: given value to check against
+        """
+        return not self._prop_equal_to(prop, value)
```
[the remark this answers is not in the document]
anish-mudaraddi: mistake - I'll make them both non-static

lib/openstack_query/handlers/presets/preset_handler_generic.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @staticmethod
+    def _prop_equal_to(prop: Any, value: Any) -> bool:
+        """
+        Filter function which returns true if a prop is equal to a given value
+        :param prop: prop value to check against
+        :param value: given value to check against
+        """
+        if isinstance(prop, type(value)):
+            if isinstance(prop, object) and hasattr(prop, "__eq__"):
```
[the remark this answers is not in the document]
anish-mudaraddi: yep you're right - this isn't needed

lib/openstack_query/handlers/presets/preset_handler_base.py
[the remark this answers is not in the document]
anish-mudaraddi: I altered the function to return a reason when it fails

lib/openstack_query/handlers/presets/preset_handler_base.py
```diff
@@ -0,0 +1,154 @@
...
+        return False
+
+    def _get_mapping(self, preset: QueryPresets, prop: Enum) -> Optional[FilterFunc]:
+        """
+        Method that returns whether a filter function mapping exists in this handler
+        :param preset: A QueryPreset Enum for which a filter function mapping may exist for
+        :param prop: A property Enum for which a filter function mapping may exist for
+        """
+        if self.check_supported(preset, prop):
+            return self._FILTER_FUNCTIONS.get(preset, None)
```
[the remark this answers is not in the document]
anish-mudaraddi: Do you want me to remove this function entirely?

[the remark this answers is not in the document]
anish-mudaraddi: so we should remove the `_get_mapping` function entirely?

lib/openstack_query/handlers/presets/preset_handler_base.py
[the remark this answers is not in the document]
anish-mudaraddi: I just set the alias `OpenstackResourceObj` to Any for now

lib/openstack_query/handlers/client_side_handler.py
```diff
@@ -0,0 +1,154 @@
...
+
+        filter_func_valid, reason = self._check_filter_func(
+            filter_func, filter_func_kwargs
+        )
+        if not filter_func_valid:
+            raise QueryPresetMappingError(
+                f"Preset Argument Error: failed to build filter_function for preset '{preset.name}', reason: {reason}"
+            )
+
+        return lambda a: self._filter_func_wrapper(
```
DavidFair, line 88: Nit `a` -> `resource` or similar to make it clear what the var in the lambda is

lib/openstack_query/handlers/prop_handler.py
```diff
@@ -0,0 +1,57 @@
+from enum import Enum
+from typing import Any, Optional, Set
+from custom_types.openstack_query.aliases import (
+    PropToPropFuncMap,
+    PropFunc,
+    OpenstackResourceObj,
+)
+
+
+class PropHandler:
```
DavidFair, line 10: This class doesn't derive from `HandlerBase`? If it's not meant to then I'd argue the base class isn't worth it for a single deriving class

On the naming front `PropHandler` -> something like `OpenstackResourceAccessor` to describe what it's doing more

It's not a great name but I'm stumped for something better

lib/openstack_query/handlers/handler_base.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class HandlerBase(ABC):
+    """
+    Abstract Base class for handlers.
+    This class is an abstract class for a generic handler
+    A handler stores a mapping between a preset and a function.
+    """
+
+    def _check_supported(self, preset: QueryPresets, prop: Enum) -> bool:
```
DavidFair, line 14: I think this should be public, rather than private since the single/both deriving classes implement the public version of this

lib/openstack_query/managers/query_manager.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+
+        self._populate_query(
+            preset_details=preset_details,
+            properties_to_select=output_details.properties_to_select,
+        )
+        self._query.run(self._cloud_account)
+        return self._get_query_output(
+            output_details.output_type,
+        )
```
DavidFair, line 42: This is a really tidy way to do it :+1:

lib/openstack_query/queries/query_server.py
```diff
@@ -0,0 +1,124 @@
...
+        ServerProperties.PROJECT_ID: lambda a: a["location"]["project"]["id"],
+    }
+
+    # SERVER_SIDE_FILTER_MAPPINGS - 'filter' keyword arguments that openstack command conn.compute.servers() takes
+    # - documented here https://docs.openstack.org/openstacksdk/latest/user/proxies/compute.html
+    # - all filters are documented here
+    # https://docs.openstack.org/api-ref/compute/?expanded=list-servers-detail#list-server-request
+    _SERVER_SIDE_FILTER_MAPPINGS = {
+        QueryPresetsGeneric.EQUAL_TO: {
+            ServerProperties.USER_ID: lambda **kwargs: {"user_id": kwargs["value"]},
```
DavidFair, line 54: Is this requirement to have
`(value="some_user")` for singular scalar values documented anywhere?

Since I (as a user) might expect it to be positional in this case

lib/openstack_query/queries/query_server.py
```diff
@@ -0,0 +1,124 @@
...
+    ClientSideHandlerDateTime,
+)
+
+from openstack_query.handlers.server_side_handler import ServerSideHandler
+from openstack_query.handlers.prop_handler import PropHandler
+
+from openstack_query.utils import convert_to_timestamp
+
+
+class QueryServer(QueryWrapper):
```
DavidFair, line 25: Nit: I think we should be clearer that this is an adapter (following the adapter pattern) rather than executing the query directly.

I'm on the fence between something like
`ServerQueryAdapter`, `ServerQueryMapping` or something completely different? Thoughts?

lib/openstack_query/queries/query_server.py
```diff
@@ -0,0 +1,124 @@
...
+    def __init__(self):
+        client_side_handlers = [
+            ClientSideHandlerGeneric(self.CLIENT_SIDE_MAPPING_GENERIC),
+            ClientSideHandlerDateTime(self.CLIENT_SIDE_MAPPING_DATETIME),
+            ClientSideHandlerString(self.CLIENT_SIDE_MAPPING_STRING),
+        ]
+        server_side_handler = ServerSideHandler(self._SERVER_SIDE_FILTER_MAPPINGS)
+        prop_handler = PropHandler(self._PROPERTY_MAPPINGS)
+
+        super().__init__(prop_handler, client_side_handlers, server_side_handler)
```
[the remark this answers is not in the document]
DavidFair: The runner below should really be passed through `__init__` to ensure it's set, as I can see this being a bug in-wait if you forget it

lib/openstack_query/queries/query_wrapper.py
```diff
@@ -0,0 +1,130 @@
...
+    ):
+        """
+        Public method that runs the query provided and outputs
+        :param cloud_account: The account from the clouds configuration to use
+        :param from_subset: A subset of openstack resources to run query on instead of querying openstacksdk
+        :param kwargs: keyword args that can be used to configure details of how query is run
+            - valid kwargs specific to resource
+        """
+        filter_func = self.builder.filter_func()
+        filter_kwargs = self.builder.server_side_filters()
```
DavidFair, line 99: Nit, these names need updating to something like `local_filters` and `server_filters` following on from the class renames

DavidFair (changes requested):
Partial review 2/3

lib/openstack_query/runners/query_runner.py
```diff
@@ -0,0 +1,97 @@
...
+            else:
+                results_resource_objects = self._run_query(
+                    conn, filter_kwargs, **kwargs
+                )
+
+        if filter_kwargs is None or force_filter_func_usage:
+            results_resource_objects = self._apply_filter_func(
+                results_resource_objects,
+                filter_func,
+            )
```
DavidFair, line 57: Can we simplify this flow a bit?

In hacky code:

```python
if from_subset and self._server_filters:
   raise ValueError("Cannot parse from a subset and server filter")

if not self._server_filters:
    self._server_filters = get_all_objects()

# Then we are always parsing a subset at this level
with ... as conn:
       subset = if from_subset if from_subset else self._run_query(self._server_filters)
        results_resource_objects = self._parse_subset(conn, self._client_filters, from_subset)
```

This would also drop the need to use `force_filter_func_usage` too

lib/openstack_query/runners/server_runner.py
```diff
@@ -0,0 +1,104 @@
...
+        :param conn: An OpenstackConnection object - used to connect to openstacksdk
+        :param filter_kwargs: An Optional set of filter kwargs to pass to conn.compute.servers()
+            to limit the servers being returned. - see https://docs.openstack.org/api-ref/compute/#list-servers
+
+        :param kwargs: a set of extra meta params that override default openstacksdk command. Current valid kwargs are:
+
+            'from_projects': Union[List[str], List[Project]] - takes a list of projects to run the query on
+
+            'from_subset': Union[List[str], List[Server]] - takes a list of servers to run the query on
+            (forces use of filter functions)
```
DavidFair, line 37: Can these move into the signature? We can just default them to None

lib/openstack_query/runners/server_runner.py
```diff
@@ -0,0 +1,104 @@
...
+            if len(kwargs["from_projects"]) == 0:
+                raise ParseQueryError(
+                    "'from_project' meta kwarg given but no projects given to search in"
+                )
+            for project in kwargs["from_projects"]:
+                if not isinstance(project, Project):
+                    raise ParseQueryError(
+                        "'from_project' only accepts Openstack 'Project' objects"
+                    )
+                projects.append(project)
```
DavidFair, line 51: This should be extracted into a private method and just assign the result
```
projects = self._get_projects(kwargs.get("from_projects", None))

# ...
ProjectIdentifier = Union[str, Project]
def _get_projects(projects: Optional[ProjectIdentifier]) -> return List[ProjectIdentifier]:
    pass

Comment (DavidFair) [reply lib/openstack_query/runners/server_runner.py:51]: I would argue in a future PR we'll want to extract this as it's something that will apply for everything

File: lib/openstack_query/runners/server_runner.py [hunk 39]
```diff
@@ -0,0 +1,104 @@
...
+        This method is a helper function that will run the query on a list of openstack projects given
+        :param conn: An OpenstackConnection object - used to connect to openstacksdk
+        :param projects: A list of openstacksdk projects to run query on
+        :param filter_kwargs: An Optional set of filter kwargs to pass to conn.compute.servers()
+        """
+        return [
+            server
+            for project in projects
+            for server in self._run_query_on_project(conn, project, filter_kwargs)
+        ]
```
DavidFair, line 74: Do we not group by project?
I would argue the that `Dict[ProjectIdentifier, Server]` is more useful and people can always just do `servers = result.values()` at the other end

lib/openstack_query/runners/server_runner.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def _parse_subset(
+        self, _: OpenstackConnection, subset: List[Server]
+    ) -> List[Server]:
+        """
+        This method is a helper function that will check a list of servers to ensure that they are valid Server objects
+        :param subset: A list of openstack Server objects
+        """
+        for server in subset:
+            if not isinstance(server, Server):
```
DavidFair, line 100: Nit: `if any(not isinstance(i, Server) for i in subset):`

lib/openstack_query/utils.py
```diff
@@ -0,0 +1,53 @@
+from datetime import datetime, timedelta
+from exceptions.missing_mandatory_param_error import MissingMandatoryParamError
+
+
```
DavidFair, line 4: Can we rename this to `time_utils` and maybe make a class of static methods, so that we can easily mock / DI it?

lib/openstack_query/utils.py
```diff
@@ -0,0 +1,53 @@
...
+    return current_time - prop_time_in_seconds
+
+
+def convert_to_timestamp(
+    days: int = 0,
+    hours: int = 0,
+    minutes: int = 0,
+    seconds: float = 0,
+    timestamp_fmt: str = "%Y-%m-%dT%H:%M:%SZ",
```
DavidFair, line 37: Nit: I'm sure this format should be built into the datetime lib as a preset?

lib/openstack_query/utils.py
```diff
@@ -0,0 +1,53 @@
...
+):
+    """
+    Helper function to convert a relative time from current time into a timestamp
+    :param days: (Optional) relative number of days since current time
+    :param hours: (Optional) relative number of hours since current time
+    :param minutes: (Optional) relative number of minutes since current time
+    :param seconds: (Optional) relative number of seconds since current time
+    :param timestamp_fmt: (Optional) timestamp format of result (default: yyyy-MM-ddTHH:mm:ssZ)
```
DavidFair, line 45: I think we should drop this as a param and always use ISO 8601. If people want to parse it they can do it client-side

lib/structs/query/query_output_details.py
```diff
@@ -0,0 +1,16 @@
...
+@dataclass
+class QueryOutputDetails:
+    """
+    Structured data to pass to Query<Resource> object. Needed for certain query output actions -
+    setting which properties to select select(), or select_all(),
+    and how to output them to_html(), to_string() etc.
+    """
+
+    properties_to_select: Optional[Set[Enum]]
+    output_type: Optional[QueryOutputTypes]
```
DavidFair, line 16: These are optional, but there's none default assignment making them mandatory?

lib/structs/query/query_preset_details.py
```diff
@@ -0,0 +1,17 @@
...
+
+@dataclass
+class QueryPresetDetails:
+    """
+    Structured data passed to a Query<Resource> object when calling the where() function
+    describes how to filter for the query.
+    """
+
+    preset: QueryPresets
+    prop: Enum
```
DavidFair, line 16: Could we specify which `Enum` or add a stub base class which such as `PropEnum` to subset this down?

tests/lib/openstack_query/handlers/test_client_side_handler.py
```diff
@@ -0,0 +1,207 @@
...
+from parameterized import parameterized
+
+from openstack_query.handlers.client_side_handler import ClientSideHandler
+from exceptions.query_preset_mapping_error import QueryPresetMappingError
+
+from tests.lib.openstack_query.mocks.mocked_query_presets import MockQueryPresets
+from tests.lib.openstack_query.mocks.mocked_props import MockProperties
+
+
+class ClientSideHandlerBaseTests(unittest.TestCase):
```
DavidFair, line 12: Aside: what's your thoughts on switching this project to `pytest in the future`? Since we get fixtures / paramterize etc.

tests/lib/openstack_query/handlers/test_client_side_handler.py
```diff
@@ -0,0 +1,207 @@
...
+            MockQueryPresets.ITEM_1: ["*"],
+            MockQueryPresets.ITEM_2: [MockProperties.PROP_1, MockProperties.PROP_2],
+            MockQueryPresets.ITEM_3: [MockProperties.PROP_3, MockProperties.PROP_4],
+        }
+        self.instance = ClientSideHandler(_FILTER_FUNCTION_MAPPINGS)
+        self.instance._FILTER_FUNCTIONS = {
+            MockQueryPresets.ITEM_1: "item1_func",
+            MockQueryPresets.ITEM_2: "item2_func",
+            MockQueryPresets.ITEM_3: "item3_func",
+        }
```
DavidFair, line 33: If we're having to mess with the object's privates, maybe we should have a `FakeClientSideHandler` or similar?

tests/lib/openstack_query/handlers/test_client_side_handler.py
```diff
@@ -0,0 +1,207 @@
...
+            )
+        )
+
+    @patch(
+        "openstack_query.handlers.client_side_handler.ClientSideHandler._check_filter_func"
+    )
+    @patch(
+        "openstack_query.handlers.client_side_handler.ClientSideHandler._filter_func_wrapper"
+    )
+    def test_get_filter_func_valid(
```
DavidFair, line 71: Do the names need changing in the tests too?

tests/lib/openstack_query/handlers/test_client_side_handler.py
```diff
@@ -0,0 +1,207 @@
...
+            )
+
+        # when the preset is valid, but property is invalid
+        with self.assertRaises(QueryPresetMappingError):
+            self.instance.get_filter_func(
+                MockQueryPresets.ITEM_2,
+                MockProperties.PROP_3,
+                mock_prop_func,
+                mock_kwargs,
+            )
```
DavidFair, line 119: Can we split this into separate test cases, so it's obvious if something breaks it will either be
`test_get_filter_func_invalid_prop`
`test_get_filter_func_invalid_preset`

tests/lib/openstack_query/handlers/test_client_side_handler.py
```diff
@@ -0,0 +1,207 @@
...
+        mock_check_filter_func.return_value = False, "some-error"
+        with self.assertRaises(QueryPresetMappingError):
+            self.instance.get_filter_func(
+                MockQueryPresets.ITEM_1,
+                MockProperties.PROP_1,
+                mock_prop_func,
+                mock_kwargs,
+            )
+
+    def test_filter_func_wrapper(self):
```
DavidFair, line 131: Ditto on splitting apart the good and invalid paths into own cases

tests/lib/openstack_query/handlers/test_client_side_handler.py
```diff
@@ -0,0 +1,207 @@
...
+                "all possible",
+                {"arg1": 12, "arg2": "non-default", "some_kwarg": "some-val"},
+            ),
+            (
+                "different order",
+                {"arg2": "non-default", "some_kwarg": "some-val", "arg1": 12},
+            ),
+        ]
+    )
+    def test_check_filter_func(self, name, valid_kwargs_to_test):
```
DavidFair, line 182: I would argue this parametrised test is the pattern we should use to test invalid too, as we can add edge cases easily in the future too

tests/lib/openstack_query/handlers/test_client_side_handler_datetime.py
```diff
@@ -0,0 +1,358 @@
...
+        [(f"test {preset.name}", preset) for preset in QueryPresetsDateTime]
+    )
+    def test_check_supported_all_presets(self, name, preset):
+        """
+        Tests that handler supports all generic query client_side
+        """
+        self.assertTrue(self.instance.check_supported(preset, MockProperties.PROP_1))
+
+    @parameterized.expand(
+        [
```
DavidFair, line 41: Can we extract this list out to be a class variable, as it should help the format mangling too

tests/lib/openstack_query/handlers/test_client_side_handler_datetime.py
```diff
@@ -0,0 +1,358 @@
...
+            (
+                "prop timestamp is older by 1 day",
+                1,
+                0,
+                0,
+                0,
```
DavidFair, line 47: Having seen this usage, I think we should use https://peps.python.org/pep-3102/ to force kwarg only for the units. 
Whilst this is the test case params, there's nothing currently to stop someone using the comparators with a random list of values (from a future readers POV)

tests/lib/openstack_query/handlers/test_client_side_handler_string.py
```diff
@@ -0,0 +1,74 @@
...
+            ("No alphabetic characters", "[A-Za-z]+", "123", False),
+            ("Alphabetic and numeric characters", "[A-Za-z0-9]+", "abc123", True),
+            ("Empty string, no match", "[A-Za-z]+", "", False),
+        ]
+    )
+    def test_prop_matches_regex_valid(
+        self, name, regex_string, test_prop, expected_out
+    ):
+        out = self.instance._prop_matches_regex(test_prop, regex_string)
+        assert out == expected_out
```
DavidFair, line 50: We should patch the regex call, and just assert it's called as we expect rather than testing the underlying implementation

tests/lib/openstack_query/handlers/test_server_side_handler.py
```diff
@@ -0,0 +1,139 @@
...
+
+    @patch(
+        "openstack_query.handlers.server_side_handler.ServerSideHandler._check_filter_mapping"
+    )
+    @patch(
+        "openstack_query.handlers.server_side_handler.ServerSideHandler._get_mapping"
+    )
+    def test_get_filters_valid(self, mock_get_mapping, mock_check_filter_mapping):
+        mock_params = {"arg1": "val1", "arg2": "val2"}
+        mock_filter_func = MagicMock()
```
DavidFair, line 82: Do we have any tests for our filter funcs separately?

tests/lib/openstack_query/managers/test_query_manager.py
```diff
@@ -0,0 +1,82 @@
...
+
+        # with no properties to select
+        self.query.reset_mock()
+        self.instance._populate_query(MOCKED_PRESET_DETAILS, None)
+        self.query.select_all.assert_called_once()
+        self.query.where.assert_called_once_with(
+            MOCKED_PRESET_DETAILS.preset,
+            MOCKED_PRESET_DETAILS.prop,
+            MOCKED_PRESET_DETAILS.args,
+        )
```
DavidFair, line 82: Split this into a separate case

tests/lib/openstack_query/managers/test_server_manager.py
```diff
@@ -0,0 +1,195 @@
...
+                    "minutes": 1,
+                    "seconds": 0,
+                    "prop_timestamp_fmt": "%Y-%m-%dT%H:%M:%SZ",
+                },
+            ),
+            output_details=MOCKED_OUTPUT_DETAILS,
+        )
+        self.assertEqual(res, "some-output")
+
+    @patch("openstack_query.managers.query_manager.QueryManager._build_and_run_query")
```
DavidFair, line 73: We can move this patch to class level, since it applies to everything

tests/lib/openstack_query/queries/test_query_wrapper.py
```diff
@@ -0,0 +1,134 @@
...
+        )
+        self.assertEqual(res, self.instance)
+
+        # with 2 props
+        mock_query_output.reset_mock()
+        res = self.instance.select(MockProperties.PROP_1, MockProperties.PROP_2)
+        mock_query_output.parse_select.assert_called_once_with(
+            MockProperties.PROP_1, MockProperties.PROP_2, select_all=False
+        )
+        self.assertEqual(res, self.instance)
```
DavidFair, line 50: ^

tests/lib/openstack_query/runners/test_query_runner.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @patch("openstack_query.runners.query_runner.QueryRunner._apply_filter_func")
+    @patch("openstack_query.runners.query_runner.QueryRunner._run_query")
+    def test_run_only_filter_func(self, mock_run_query, mock_apply_filter_func):
+        mock_run_query.return_value = ["openstack-resource-1", "openstack-resource-2"]
+        mock_apply_filter_func.return_value = ["filtered-openstack-resource"]
+
+        res = self.instance.run(
+            cloud_account="test",
+            filter_func="some-filter-func",
```
DavidFair, line 26: Pass a mock here, rather than a string
Then we can also assert that it's not called at this level (until we want it to be called)

tests/lib/openstack_query/test_utils.py
```diff
@@ -0,0 +1,77 @@
...
+from nose.tools import raises
+
+import openstack_query.utils as utils
+from exceptions.missing_mandatory_param_error import MissingMandatoryParamError
+
+
+class UtilsTests(unittest.TestCase):
+    """
+    Runs various tests to ensure that utils module functions expectedly
+    """
```
DavidFair, line 15: Were missing a test for `currentTime`, this can simply be a patch to check we forward on correctly

I want to keep currentTime as we can simply do this in other tests
```
mocked = autospec(TimeUtils)
# ...
mocked.current_time.assert_called_once()
```

DavidFair (changes requested):
Review 3/3

lib/openstack_query/handlers/prop_handler.py
anish-mudaraddi: `HandlerBase` has two deriving classes - `ClientSideHandler` and `ServerSideHandler` 
`PropHandler` should derive it - but it only takes `prop` as a parameter (not preset as the other two)

I can remove it if you still think it should

lib/openstack_query/queries/query_server.py
anish-mudaraddi: the `EQUAL_TO` preset - which is the filter we are setting here - takes one parameter - `value` - which is the value to compare against.
Yes I think it should be a positional argument

lib/openstack_query/queries/query_server.py
anish-mudaraddi: makes sense - I'll go with ServerQueryAdapter and rename the directory

lib/openstack_query/queries/query_server.py
```diff
@@ -0,0 +1,124 @@
...
+    CLIENT_SIDE_MAPPING_STRING = {
+        QueryPresetsString.MATCHES_REGEX: [ServerProperties.SERVER_NAME],
+    }
+
+    def __init__(self):
+        client_side_handlers = [
+            ClientSideHandlerGeneric(self.CLIENT_SIDE_MAPPING_GENERIC),
+            ClientSideHandlerDateTime(self.CLIENT_SIDE_MAPPING_DATETIME),
+            ClientSideHandlerString(self.CLIENT_SIDE_MAPPING_STRING),
+        ]
```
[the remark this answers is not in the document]
anish-mudaraddi: how would this work? I create a dataclass like so:
```
@dataclass
class QueryClientSideHandlers:
    """
    Stores client side preset handlers
    """

    generic_handler: ClientSideHandlerGeneric
    string_handler: ClientSideHandlerString
    datetime_handler: Optional[ClientSideHandlerDateTime]
    integer_handler: Optional[ClientSideHandlerInteger]
```

and I instantiate it in `QueryServer` like so 
```
def _get_client_side_handlers(self) -> QueryClientSideHandlers:

    return QueryClientSideHandlers(

        # set generic query preset mappings
        generic_handler=ClientSideHandlerGeneric({...})

        # set string query preset mappings
        string_handler=ClientSideHandlerString({...})

        # set datetime query preset mappings
        datetime_handler=ClientSideHandlerDateTime({...})

        # set integer query preset mappings
        integer_handler=None
    )
```

Do I then iterate through the Dataclass attributes to make a list of handlers? This is what `QueryBuilder` expects

lib/openstack_query/queries/query_server.py
anish-mudaraddi: I don't want to rename this - simply because this will - most likely - be instantiated by users to run their queries if they wish to do so programmatically - naming it something like `ServerQueryAdapter` may confuse them

lib/openstack_query/handlers/prop_handler.py
DavidFair: Ah I didn't see the other two deriving
Does it make sense to derive `PropHandler` from this or are we forcing an arbitrary relation between the two? I don't have a strong feeling either way on this :thinking:

lib/openstack_query/queries/query_server.py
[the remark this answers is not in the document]
DavidFair: Actually I was thinking just pass it through `__init__` as before in lieu of the list?

lib/custom_types/openstack_query/aliases.py
```diff
@@ -1,23 +1,24 @@
-from typing import List, Callable, Any, Dict
+from typing import List, Callable, Any, Dict, Union, Literal
 from enum import Enum
 from enums.query.query_presets import QueryPresets
 
+OpenstackResourceObj = Any
 
-PropFunc = Callable[[Any], Any]
-PropToPropFuncMap = Dict[Enum, PropFunc]
+PropFunc = Callable[[OpenstackResourceObj], Dict[str, Any]]
```
DavidFair, line 7: Nit: Is the `Dict[str, Any]` here related to any of the others?

I'm looking at `PresetKwargs` for example or `ServerSideFilters` to see if there's any overlap

lib/openstack_query/query_builder.py
```diff
@@ -0,0 +1,94 @@
...
+        self._client_side_handlers = client_side_handlers
+        self._prop_handler = prop_handler
+        self._server_side_handler = server_side_handler
+
+        self._filter_func = None
+        self._server_side_filters = None
+
+    @property
+    def filter_func(self):
+        return self._filter_func
```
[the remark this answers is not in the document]
anish-mudaraddi: Don't need to raise error - you can have a query which returns all values - hence no server-side or client-side filters - in that case we want this function to return None

lib/openstack_query/query_output.py
```diff
@@ -0,0 +1,120 @@
...
+    TODO: Class should also handle grouping and sorting results
+    """
+
+    def __init__(self, prop_handler: PropHandler):
+        self._prop_handler = prop_handler
+        self._props = set()
+        self._results = []
+
+    @property
+    def results(self):
```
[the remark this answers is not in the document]
anish-mudaraddi: yes

lib/openstack_query/runners/server_runner.py
anish-mudaraddi: You mentioned previously that openstack `find_server` is slow - I need to use it here to get each project from its identifier - I think in this case its worthwhile to do since you mentioned that `from_project` would apply for everything

lib/openstack_query/runners/server_runner.py
anish-mudaraddi: I can do this for this method `_run_query_on_projects` - but overall I want `run_query` to return all servers not grouped by anything - I want the user to explicitly tell us to group by projects - via calling `group_by` 
e.g.: `QueryServer().group_by(Project).to_list()`

lib/openstack_query/handlers/prop_handler.py
anish-mudaraddi: yea I think we should - since all three should implement `check_supported` - its just that this function takes different arguments - `check_supported` in all three does the same thing

lib/openstack_query/query_output.py
```diff
@@ -0,0 +1,120 @@
...
+        return self._results
+
+    def sort_by(self, sort_by: Enum, reverse=False):
+        """
+        Public method used to configure sorting results
+        :param sort_by: name of property to sort by
+        :param reverse: False is sort by ascending, True is sort by descending, default False
+        """
+        raise NotImplementedError
```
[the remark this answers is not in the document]
anish-mudaraddi: yes - agreed

lib/openstack_query/runners/server_runner.py
anish-mudaraddi: I've changed this function to return a dictionary `{project_id: [servers]}` - then I flatten this into a list of servers in `run_query` - if we decide that we want to group by projects as default then I can change `run_query` but for now I left it as is - changing `run_query` will mean reworking `QueryOutput` a bit

tests/lib/openstack_query/handlers/test_client_side_handler.py
anish-mudaraddi: I wanted to use `pytest` - but all the other tests for stackstorm defaulted to using `nose.tools` and `unittest` - so I stuck with that. 

When we move this out - we may want to refactor to use `pytest`

lib/openstack_query/runners/query_runner.py
```diff
@@ -23,50 +24,49 @@ def __init__(self, connection_cls=OpenstackConnection):
...
+                else self._run_query(conn, server_side_filters, **kwargs)
+            )
 
-        if filter_kwargs is None or force_filter_func_usage:
-            results_resource_objects = self._apply_filter_func(
-                results_resource_objects,
-                filter_func,
+        if client_side_filter_func and not server_side_filters:
+            resource_objects = self._apply_client_side_filter(
+                resource_objects, client_side_filter_func
```
DavidFair, line 56: Any reason we can't have it do a client side after a server side?

tests/lib/openstack_query/runners/test_server_runner.py
```diff
@@ -1,10 +1,13 @@
 import unittest
+from openstack.exceptions import ResourceNotFound
+
 from unittest.mock import Mock, MagicMock, call, patch
 from openstack_query.runners.server_runner import ServerRunner
 
 from openstack.compute.v2.server import Server
 from openstack.identity.v3.project import Project
 from exceptions.parse_query_error import ParseQueryError
+from nose.tools import raises
```
DavidFair, line 10: This circles back to shall we double down on `unitTest` for just this lib, or consider pytest?

lib/openstack_query/runners/server_runner.py
DavidFair: I'm more than happy with this, since it should also make it easier when we do a `groupBy` in the query layer later on too :+1:

tests/lib/openstack_query/runners/test_server_runner.py
DavidFair: Ah I saw your comment above on a refresh, ACK: on Pytest when we refactor

lib/openstack_query/runners/query_runner.py
anish-mudaraddi: Well - right now we allow only one Preset per query - each preset has:
- an associated `client_side_filter_func`
- an optional set of `server_side_filters` 
They both do the same filtering - hence we only need one or the other

lib/custom_types/openstack_query/aliases.py
anish-mudaraddi: I renamed type aliases and added comments to them so they are less confusing

tests/lib/openstack_query/handlers/test_client_side_handler.py
anish-mudaraddi: what do you mean by that? 
Here we are testing `ClientSideHandler` - we mock a set of filter function mappings that would be passed as input to the constructor
I don't see how a `FakeClientSideHandler` would help here?

tests/lib/openstack_query/handlers/test_client_side_handler.py
anish-mudaraddi: No they don't - we're testing `ClientSideHandler` object - its given that the filter function works only on the client side

tests/lib/openstack_query/handlers/test_server_side_handler.py
anish-mudaraddi: yes - but filter funcs are handled in `ClientSideHandlers` - what are you referring to here? 
This function checks if a set of kwargs are returned that we can pass to openstacksdk command

tests/lib/openstack_query/test_utils.py
anish-mudaraddi: do we really need a test for this - I can't simply patch it - I have to use `freezegun` package
https://stackoverflow.com/questions/13073281/how-to-mock-pythons-datetime-now-in-a-class-method-for-unit-testing

(that's why I made that method to begin with)

tests/lib/openstack_query/handlers/test_client_side_handler_datetime.py
anish-mudaraddi: I refactored this class - so I don't think this is an issue anymore?

lib/openstack_query/queries/query_server.py
anish-mudaraddi: Sure :+1:

meoflynn (changes requested):
Initial full review

DavidFair (approved):
Approved, pending Megan's review

lib/openstack_query/handlers/kwarg_handler.py
```diff
@@ -0,0 +1,88 @@
...
+
+from openstack_query.handlers.handler_base import HandlerBase
+from exceptions.query_preset_mapping_error import QueryPresetMappingError
+
+
+class KwargHandler(HandlerBase):
+    """
+    This class stores a dictionary which maps a preset and property pair to a function which returns a dictionary which
+    can be passed as kwargs to an openstack list resource function - and will act as a filter which
+    corresponds to the preset. This dictionary is called KWARG_MAPPINGS
```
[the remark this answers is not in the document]
DavidFair: Yeah that's a lot better, thanks

lib/openstack_query/handlers/presets/preset_handler_datetime.py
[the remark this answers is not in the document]
DavidFair: Let's just stick with UTC and drop the parameter then

lib/openstack_query/query_builder.py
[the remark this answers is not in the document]
DavidFair: Okay thanks for explaining that

lib/custom_types/openstack_query/aliases.py
DavidFair: It's much better now :+1:

lib/openstack_query/runners/query_runner.py
DavidFair: Based on our discussion in person: definitely for a future PR but I can see us only allowing a single server side filter with n client side

lib/openstack_query/handlers/handler_base.py
```diff
@@ -0,0 +1,26 @@
...
+        :param preset: An Enum which represents a preset
+        :param prop: An Enum which represents a property for preset to act on
+        """
+
+    @abstractmethod
+    def preset_known(self, preset: QueryPresets) -> bool:
+        """
+        Method used to return if a preset is supported by the handler
+        :param preset: An Enum which represents a preset
+        """
```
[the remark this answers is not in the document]
anish-mudaraddi: no they're abstract methods - the implementation of these methods is done in the deriving classes.

tests/lib/openstack_query/handlers/test_client_side_handler_datetime.py
```diff
@@ -0,0 +1,158 @@
...
+        Setup for tests
+        """
+        super().setUp()
+        # sets filter function mappings so that PROP_1 is valid for all client_side
+        _FILTER_FUNCTION_MAPPINGS = {
+            preset: [MockProperties.PROP_1] for preset in QueryPresetsDateTime
+        }
+        self.instance = ClientSideHandlerDateTime(_FILTER_FUNCTION_MAPPINGS)
+
+        self.test_cases = {
```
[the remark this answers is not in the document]
anish-mudaraddi: Yea it's a little confusing 

We set the current time as 2023-06-04 10:30 AM
We then calculate a relative timestamp from the current time - based on `days`, `hours`, `minutes`, `seconds`
So if we give `days=1` - we expect the relative timestamp to be 2023-06-03 10:30AM

Then we just compare that against `prop` and see if the `prop` timestamp is older or younger than the `relative` timestamp

This just allows you to put in `days=60` for example and find all servers older than 60 days (since prop would be the timestamp each server was created)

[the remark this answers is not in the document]
anish-mudaraddi: The names for the test e.g. `older_than_1_day` is what we expect the difference to be between the `prop` timestamp and the calculated `relative` timestamp

tests/lib/openstack_query/test_query_output.py
```diff
@@ -0,0 +1,145 @@
...
+        self.assertEqual(expected_out, res)
+
+    @patch("openstack_query.query_output.QueryOutput._generate_table")
+    def test_to_string(self, mock_generate_table):
+        """
+        Tests that to_string function returns correct value
+        """
+        """
+        Tests that to_html function returns correct value
+        """
```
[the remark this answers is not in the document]
anish-mudaraddi: yep - fixed :+1:

DavidFair (approved):
LGTM, I'm happy approving this on Megan's behalf

lib/openstack_query/handlers/handler_base.py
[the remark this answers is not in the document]
DavidFair: I'm going to mark this as resolved :+1:

lib/openstack_query/queries/query_server.py
```diff
@@ -0,0 +1,124 @@
...
+class QueryServer(QueryWrapper):
+    """
+    Query class for querying Openstack Server objects.
+    Define property mappings, kwarg mappings and filter function mappings related to servers here
+    """
+
+    # PROPERTY_MAPPINGS set how to get properties of a openstack server object
+    # possible properties are documented here:
+    # https://docs.openstack.org/openstacksdk/latest/user/resources/compute/v2/server.html#openstack.compute.v2.server.Server
+    _PROPERTY_MAPPINGS = {
```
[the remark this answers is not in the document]
DavidFair: This is much clearer now, thanks :+1:

tests/lib/openstack_query/handlers/test_client_side_handler_datetime.py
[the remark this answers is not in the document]
DavidFair: So this effectively gives us fixed points to backwards from then is my understanding

DavidFair:
Really good work, thanks everyone for reviewing and working on this
I'm really looking forward to seeing this running in anger :beers:

PR #74, opened 2023-07-11
Openstack Query Library - Various Bugfixes

anish-mudaraddi:
### Description:
Some bugfixes for Openstack query library 
- some import issues
- fixes so code works for python 3.6
    - removal of `Literal` typing - not supported in python 3.6 - which our StackStorm VM uses
### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:


lib/openstack_query/handlers/client_side_handler_datetime.py
```diff
@@ -15,16 +17,17 @@ class ClientSideHandlerDateTime(ClientSideHandler):
     Filter functions which map to QueryPresetsDateTime are defined here
     """
 
-    def __init__(self, filter_function_mappings: PresetPropMappings):
-        super().__init__(filter_function_mappings)
+    def __init__(self, _filter_function_mappings: PresetPropMappings):
```
DavidFair, line 20: Nit: this shouldn't be a private var in a constructor, even if it gets mapped there ultimately

lib/openstack_query/runners/server_runner.py
```diff
@@ -98,13 +101,14 @@ def _run_query_on_project(
         """
         server_filters = {"project_id": project["id"], "all_tenants": True}
         server_filters.update(filter_kwargs if filter_kwargs else {})
-        return list(conn.compute.servers(filters=server_filters))
+        return list(conn.compute.servers(all_projects=False, **server_filters))
```
DavidFair, line 104: Future issue, we should allow someone to pass `all_project=True` for some of the st2 queries

DavidFair (changes requested):

lib/openstack_query/runners/server_runner.py
anish-mudaraddi: no it will always be false here - since were going through each project to find the servers
we can do `all_project=True` instead of iterating through projects but we run into a bug where openstack can't list more than 1000 VMs

lib/openstack_query/managers/query_manager.py
```diff
@@ -68,11 +70,13 @@ def _populate_query(
...
-            self._query.select(properties_to_select)
+            self._query.select(*properties_to_select)
         else:
             self._query.select_all()
 
         if preset_details:
             self._query.where(
-                preset_details.preset, preset_details.prop, preset_details.args
+                preset=preset_details.preset,
+                prop=preset_details.prop,
```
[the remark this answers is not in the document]
anish-mudaraddi: i don't want to - since this is calling a public facing method `where` - I don't want users to create a dataclass to use it

tests/lib/openstack_query/handlers/test_client_side_handler.py
```diff
@@ -8,6 +8,8 @@
 from tests.lib.openstack_query.mocks.mocked_query_presets import MockQueryPresets
 from tests.lib.openstack_query.mocks.mocked_props import MockProperties
 
+# pylint:disable=protected-access, unused-argument
```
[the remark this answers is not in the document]
anish-mudaraddi: this is required for tests since we have unittests for a number of private/protected methods - hence the file scope

DavidFair (approved):

PR #76, opened 2023-07-13
Adding openstack query actions

anish-mudaraddi:
### Description:
Added actions for server query which match the ones in `todo`

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:


actions/src/server_query_actions.py
```diff
@@ -0,0 +1,40 @@
...
+
+from enums.query.props.server_properties import ServerProperties
+from enums.query.query_output_types import QueryOutputTypes
+from enums.cloud_domains import CloudDomains
+
+from openstack_query.managers.server_manager import ServerManager
+from structs.query.query_output_details import QueryOutputDetails
+
+
+class ServerQueryActions(Action):
```
meoflynn, line 12: Could we add a docstring to describe this class?

meoflynn (changes requested):
Looks good, just one minor change required to add a docstring to one class

actions/server.search.all.servers.yaml
DavidFair, line 1: Re: file naming

I'm wondering if we group all the search objects together, so in the web UI they don't spread around?
Something like `search.<object>.<query>`

E.g. for this file: `search.server.strings` - but I'm not happy with `strings`

actions/server.search.created.at.relative.yaml
DavidFair, line 1: Then this one becomes
`search.server.relative.created`

Since you can simply look at the last work in the `relative` sub-group to see, oh I want `relative.updated` or `relative.created`?

If we do go for this approach I think it's worth throwing into the README as a naming standard so we stick to whatever we agree with

actions/server.search.created.at.relative.yaml
```diff
@@ -0,0 +1,87 @@
...
+entry_point: src/server_query_actions.py
+name: server.search.created.at.relative
+parameters:
+  timeout:
+    default: 5400
+  submodule:
+    default: search_servers_older_than_relative_to_now
+    enum:
+      - "search_servers_older_than_relative_to_now"
+      - "search_servers_younger_than_relative_to_now"
```
DavidFair, line 13: Does this make sense to have? I can't imagine how this works without a time machine :sweat_smile:

DavidFair: Instead I think this should take a ISO8601 style time (without the TZ stuff), or default to `now` as a paremter then it makes sense?

This might be feature creep so feel free to put it into the TODO pile

actions/server.search.all.servers.yaml
```diff
@@ -0,0 +1,65 @@
...
+    immutable: true
+  cloud_account:
+    description: The clouds.yaml account to use whilst performing this action
+    required: true
+    type: string
+    default: "dev"
+    enum:
+      - "dev"
+      - "prod"
+  properties_to_select:
```
DavidFair, line 21: I can't see a property to manipulate e.g. if I wanted to find servers belonging to a `image_id` for example?

actions/server.search.created.at.relative.yaml
```diff
@@ -0,0 +1,87 @@
...
+      - "to_str"
+    description: "
+    A string representing how to return the results of the query
+      - 'to_html' - a tabulate table (in html)
+      - 'to_list' - properties dicts as a python list
+      - 'to_object_list - as a list of openstack resources
+      - 'to_str' - a tabulate table"
+    required: true
+  days:
+    description: "(Optional) Number of days threshold for selecting servers"
```
DavidFair, line 68: Nit: call these `relative_days` `relative_hours`...etc. so it makes it clear that these do not control the fixed point of time we are comparing to?

actions/server.search.name.in.yaml
```diff
@@ -0,0 +1,71 @@
...
+enabled: true
+entry_point: src/server_query_actions.py
+name: server.search.name
+parameters:
+  timeout:
+    default: 5400
+  submodule:
+    default: search_servers_name_in
+    enum:
+      - "search_servers_name_in"
```
DavidFair, line 12: Cant this be done with the `server.search.all` action? Since effectively we start mapping 1:1 actions:properties for each object with this approach

DavidFair (changes requested):
YAML section review

actions/src/server_query_actions.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    Stackstorm Action class that dynamically dispatches actions related to Server Queries to the corresponding
+    method in the ServerManager class
+    Actions that will be handled by this class follow the format server.search.*
+    """
+
+    def __init__(self, *args, **kwargs):
+        self._server_manager = ServerManager(cloud_account=CloudDomains)
```
DavidFair, line 20: Why do we pass in the Enum here? If this is a hangover from DI + Testing we should just make `CloudDomains` to default type in `init`

actions/src/server_query_actions.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+        Dynamically dispatches to the ServerManager method wanted
+        :param submodule: submodule name which corresponds to function in ServerManager
+        :param cloud_account: A string representing the account from the clouds configuration to use
+        :param properties_to_select: A list of strings representing the Server Properties to return for each Server query result
+        :param return_type: A string representing how to return the results of the query
+        :param kwargs: A set of extra parameters to pass to manager method, if any
+        """
+        server_mgr = self._server_manager(CloudDomains[cloud_account.upper()])
+        prop_enums = [ServerProperties[prop.upper()] for prop in properties_to_select]
```
DavidFair, line 40: Do we get a sane error message here if a user mistypes a prop name? Just because Stackstorm can sometimes give some really misleading errors, so we might have to manually check + raise a custom one if it's confusing

actions/src/server_query_actions.py
```diff
@@ -0,0 +1,47 @@
...
+        return getattr(server_mgr, submodule)(
+            output_details=QueryOutputDetails(
+                properties_to_select=prop_enums,
+                output_type=QueryOutputTypes[return_type.upper()],
```
DavidFair, line 44: Nit: can we split the construction of the data objects and call into separate lines, as we effectively have 5 operations in a single statement for future developers to parse

tests/actions/test_server_query_actions.py
```diff
@@ -0,0 +1,64 @@
...
+
+        self.action._server_manager.assert_called_once_with(CloudDomains.DEV)
+        mock_method.assert_called_once_with(
+            output_details=QueryOutputDetails(
+                properties_to_select=[
+                    ServerProperties.SERVER_ID,
+                    ServerProperties.SERVER_NAME,
+                ],
+                output_type=QueryOutputTypes.TO_STR,
+            ),
```
DavidFair, line 61: Ditto on splitting the expected data class out of the assertion, so it's two steps to a future reader

tests/lib/openstack_query/runners/test_server_runner.py
```diff
@@ -60,7 +60,7 @@ def test_get_projects_with_valid_project_identifiers(self):
 
     @raises(ParseQueryError)
     def test_get_projects_with_invalid_project_identifier(self):
-        """
+        """https://github.com/stfc/st2-cloud-pack/pull/74
```
DavidFair, line 63: Should this PR be here?

DavidFair (changes requested):

DavidFair:
Also: re: the querying stuff, don't go too mad on adding actions
There's some discussion to be had, but I would favour bespoke actions such as `search.servers.build_state`, then get the generic ones as a CLI that people can use locally

Since the ops team will reach for awk / grep, rather than go to Stackstorm for generic ones (hence wanting the CLI)

But I'm happy to figure out where that line is between Stackstorm / Query Lib on your local machine

actions/server.search.created.at.relative.yaml
anish-mudaraddi: > Does this make sense to have? I can't imagine how this works without a time machine sweat_smile

The `search_servers_younger_than_relative_to_now` means we search for servers younger than a timestamp that is calculated from `days` `hours` `minutes` `seconds` in the past from the current timestamp
so it is necessary - it doesn't find servers younger than current timestamp - which I guess you interpreted it ass

actions/server.search.all.servers.yaml
anish-mudaraddi: this action currently lists all servers - this action just limits which property the output table shows for each server
I can alter this action to allow you to search by property - I was going to create separate actions for this though i.e. search.servers.with.property where property is something people would search for

DavidFair: Hmmm, then maybe this should be called:
`list.server.all_projects` to bikeshed further?

I think we should document all the verbs, such as `list`, `search`, `notify` or `email` ....etc. in the README

actions/server.search.created.at.relative.yaml
anish-mudaraddi: I'm going to change this function in a future PR - leaving it for now

actions/server.search.all.servers.yaml
anish-mudaraddi: I'm going change this action to emulate the one that existed before `server.list`.
so this function will: 
 - allow running generic queries on all servers - property `equal_to`, not `equal_to` etc. 
 - allow running string queries on all servers - `any_in`, `not_any_in`

actions/server.search.created.at.relative.yaml
anish-mudaraddi: going to change this action in a future PR - leaving for now

actions/server.search.name.in.yaml
anish-mudaraddi: yes - will be changing the `server.search.all` action to incorporate generic and string queries. I will remove this action

actions/src/server_query_actions.py
anish-mudaraddi: good point :+1:  - although error handling will need to be done in the action - is that okay?

tests/lib/openstack_query/runners/test_server_runner.py
anish-mudaraddi: oops no it shouldn't

anish-mudaraddi:
closing this PR in favor of #77

PR #77, opened 2023-07-17
Openstack Add Server Actions

anish-mudaraddi:
### Description:
A more updated version of the PR than #76 

Server Actions include the following actions:
`server.search.by.property` - allow user to give a property, and one or more values to search for/against
`server.search.by.regex` - same as before but will allow entering a regex pattern instead of a comma-spaced set of values
`server.search.by.datetime` - allow user to give a datetime property 
    - either `server_creation_date` or `server_last_updated_date` and search for older_than / younger_than etc

Future PR to add these actions for each new Manager object 

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/server.search.by.property.yaml
+++ b/actions/server.search.by.property.yaml
@@ -1,0 +1,88 @@
+---
+description: Search for servers with a selected property matching, or not matching given value(s)
+enabled: true
+entry_point: src/server_query_actions.py
+name: server.search.by.property
+parameters:
+  timeout:
+    default: 5400
+  submodule:
+    default: "search_by_property"
+    immutable: true
+    type: string
+  cloud_account:
+    description: The clouds.yaml account to use whilst performing this action
+    required: true
+    type: string
+    default: "dev"
+    enum:
+      - "dev"
+      - "prod"
+  properties_to_select:
+    default:
+      - "flavor_id"
+      - "hypervisor_id"
+      - "image_id"
+      - "project_id"
+      - "server_creation_date"
+      - "server_description"
+      - "server_id"
+      - "server_last_updated_date"
+      - "server_name"
+      - "server_status"
+      - "user_id"
+    type: array
+    description: "
+    A comma-spaced list of server properties to display for the resulting servers - leave empty for all properties. One of:
+      - 'flavor_id'
+      - 'hypervisor_id'
+      - 'image_id'
+      - 'project_id'
+      - 'server_creation_date'
+      - 'server_description'
+      - 'server_id'
+      - 'server_last_updated_date'
+      - 'server_name'
+      - 'server_status'
+      - 'user_id'
+    Anything else will raise an error"
+    required: false
+  output_type:
+    default: "to_str"
+    type: string
+    enum:
+      - "to_html"
+      - "to_object_list"
+      - "to_list"
+      - "to_str"
+    description: "
+    A string representing how to return the results of the query
+      - 'to_html' - a tabulate table (in html)
+      - 'to_list' - properties dicts as a python list
+      - 'to_object_list - as a list of openstack resources
+      - 'to_str' - a tabulate table"
+    required: true
+  property_to_search_by:
+    default: "server_name"
+    description: "choose property to search by"
+    enum:
+      - "flavor_id"
+      - "hypervisor_id"
+      - "image_id"
+      - "server_description"
+      - "server_id"
+      - "server_name"
+      - "server_status"
+      - "user_id"
+    type: string
+    required: true
+  search_mode:
+    default: true
+    description: "True means return matching, False means return not matching"
+    required: true
+    type: boolean
+  values:
+    description: "a comma-spaced list of values to search for"
+    required: true
+    type: array
+runner_type: python-script
--- a/lib/openstack_query/managers/server_manager.py
+++ b/lib/openstack_query/managers/server_manager.py
@@ -1,9 +1,10 @@
-from typing import List
-
+from typing import List, Dict, Any
+import re
 
 from enums.query.query_presets import (
     QueryPresetsDateTime,
     QueryPresetsString,
+    QueryPresetsGeneric,
 )
 from enums.query.props.server_properties import ServerProperties
 from enums.cloud_domains import CloudDomains
@@ -11,7 +12,6 @@ from enums.cloud_domains import CloudDomains
 from openstack_query.queries.server_query import ServerQuery
 from openstack_query.managers.query_manager import QueryManager
 
-from structs.query.query_output_details import QueryOutputDetails
 from structs.query.query_preset_details import QueryPresetDetails
 from custom_types.openstack_query.aliases import QueryReturn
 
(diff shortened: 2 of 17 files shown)


lib/openstack_query/managers/server_manager.py
```diff
@@ -26,168 +27,114 @@ class ServerManager(QueryManager):
     def __init__(self, cloud_account: CloudDomains):
         QueryManager.__init__(self, query=ServerQuery(), cloud_account=cloud_account)
 
-    def search_all_servers(self, output_details: QueryOutputDetails) -> QueryReturn:
+    def search_all(self, **kwargs) -> QueryReturn:
         """
         method that returns a list of all servers
-        :param output_details: A dataclass containing config info on how results should be returned
+        :param kwargs: A set of optional kwargs to pass to the query
         """
```
DavidFair, line 34: We should document what the accepted kwargs are, or at least link to the signature rather than leave a future dev cursing your name

lib/openstack_query/managers/server_manager.py
```diff
@@ -26,168 +27,114 @@ class ServerManager(QueryManager):
...
+        Uses UTC timezone
+        :param search_mode: A string representing what type of datetime query will be run - set as a Datetime Preset
+        that the query will use
+        :param property_to_search_by: A string representing a datetime property Enum that the preset will be used on
+        - must be datetime compatible
+        :param days: (Optional) Number of relative days in the past from now to use as threshold
+        :param hours: (Optional) Number of relative hours in the past from now to use as threshold
+        :param minutes: (Optional) Number of relative minutes in the past from now to use as threshold
+        :param seconds: (Optional) Number of relative seconds in the past from now to use as threshold
+        :param kwargs: A set of optional kwargs to pass to the query
```
DavidFair, line 63: Ditto, what are the optional kwargs available to me?

lib/openstack_query/managers/server_manager.py
```diff
@@ -26,168 +27,114 @@ class ServerManager(QueryManager):
...
-        method that returns a list of all servers which were last updated after a time relative to now.
-        Uses UTC timezone
-        :param output_details: A dataclass containing config info on how results should be returned
-        :param days: number of days since current time
-        :param hours: number of hours since current time
-        :param minutes: number of minutes since current time
-        :param seconds: number of seconds since current time
+        method that builds and runs a query to find Openstack servers with a selected property
+        matching, or not matching given value(s)
+        :param search_mode: A boolean representing what to return from query, if True - use the preset ANY_IN/EQUAL_TO,
```
DavidFair, line 89: I'm not keen on using a `bool` to represent a non-true/false condition

Can we just use an enum with the values:
`NOT_ANY_IN`
`NOT_EQUAL_TO`
`ANY_IN`
`EQUAL_TO`

The fact the last two are equal in behaviour at this level is an implementation detail

lib/openstack_query/managers/server_manager.py
```diff
@@ -26,168 +27,114 @@ class ServerManager(QueryManager):
...
-        """
-        method that returns a list of all servers which have a name in a given list
-        :param output_details: A dataclass containing config info on how results should be returned
-        :param names: a list of server names to check against
-        """
+        # If values contains only one value - use EQUAL_TO/NOT_EQUAL_TO as the preset instead to speed up query
+        if len(values) == 1:
+            res = {
+                QueryPresetsString.ANY_IN: QueryPresetsGeneric.EQUAL_TO,
+                QueryPresetsString.NOT_ANY_IN: QueryPresetsGeneric.NOT_EQUAL_TO,
```
DavidFair, line 104: I can't think of a way, but it would be nice to get whatever base query runner (we abstract this to in the future) to do the fixup for all it's delegated classes, since it's an easy win

lib/openstack_query/managers/server_manager.py
```diff
@@ -26,168 +27,114 @@ class ServerManager(QueryManager):
...
-        self, output_details: QueryOutputDetails, names: List[str]
-    ) -> QueryReturn:
-        """
-        method that returns a list of all servers which have a name in a given list
-        :param output_details: A dataclass containing config info on how results should be returned
-        :param names: a list of server names to check against
-        """
+        # If values contains only one value - use EQUAL_TO/NOT_EQUAL_TO as the preset instead to speed up query
+        if len(values) == 1:
+            res = {
```
DavidFair, line 102: Nit: `res` is pretty generic, could we give it a better name?

lib/openstack_query/managers/server_manager.py
```diff
@@ -26,168 +27,114 @@ class ServerManager(QueryManager):
...
-        self, output_details: QueryOutputDetails, names: List[str]
-    ) -> QueryReturn:
+    def search_by_regex(self, property_to_search_by: str, pattern: str, **kwargs):
         """
-        method that returns a list of all servers which do not have a name matching any from a given list
-        :param output_details: A dataclass containing config info on how results should be returned
-        :param names: a list of server names to check against
+        method that builds and runs a query to find Openstack servers with a selected property matching regex.
+        :param property_to_search_by: A string representing a string property Enum that the preset will be used on
+        :param pattern: A string representing a regex pattern
```
DavidFair, line 125: This is missing `kwargs` docstring

lib/structs/query/query_output_details.py
```diff
@@ -12,5 +12,18 @@ class QueryOutputDetails:
...
     output_type: Optional[QueryOutputTypes] = None
+
+    @staticmethod
+    def from_kwargs(prop_cls: PropEnum, **kwargs):
+        """
+        method to parse and create this dataclass from kwargs
+        :param prop_cls a PropEnum class to get enum from for properties to select
+        :param kwargs: A set of kwargs to parse and use to set attributes to dataclass
+        """
+        props = {prop_cls.from_string(prop) for prop in kwargs["properties_to_select"]}
```
DavidFair, line 25: Are the kwargs a nested dict? I can't think of a case where they can pass the same name twice in properties_to_select (or maybe the set should be moved upwards)

The list comprehension could then just be done here, rather than going to a set, then a list

tests/lib/openstack_query/test_query_methods.py
```diff
@@ -151,7 +151,7 @@ def test_to_list_as_objects_true(self):
         """
         mock_query_results_list = ["object-1", "object-2"]
 
-        # pylint: disable=too-few-public-methods
+        # pylint: disable=protected-access
```
DavidFair, line 154: Why has this changed to protected-access, even though nothing else has changed in the file?

DavidFair (changes requested):

actions/server.search.by.property.yaml
```diff
@@ -0,0 +1,88 @@
...
+      - "user_id"
+    type: string
+    required: true
+  search_mode:
+    default: true
+    description: "True means return matching, False means return not matching"
+    required: true
+    type: boolean
+  values:
+    description: "a comma-spaced list of values to search for"
```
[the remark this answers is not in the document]
anish-mudaraddi: well it will find any values that match so effectively an OR

lib/openstack_query/managers/server_manager.py
anish-mudaraddi: that was my issue with changing this to `kwargs` - link would make sense

lib/openstack_query/managers/server_manager.py
anish-mudaraddi: I'm not a fan of using `EQUAL_TO` here - adds unnecessary checks to ensure that user only entered one value - which is handled internally if it were the case
I can change it to `ANY_IN`, `NOT_ANY_IN` - but it seems less intuitive to me than false/true

lib/structs/query/query_output_details.py
anish-mudaraddi: sure

lib/openstack_query/managers/server_manager.py
anish-mudaraddi: I don't understand what you mean?

lib/openstack_query/managers/server_manager.py
anish-mudaraddi: I changed this after talking to Jacob

DavidFair (approved):
LGTM, looking forward to being able to demo these

PR #83, opened 2023-07-25
Email Enhancement

anish-mudaraddi:
### Description:

This is a rewrite of the email_api. Written in a way to decouple it from all other modules. 

An email message now consists of  a list of templates - loaded sequentially
Each template accepts a set of parameters which are used to substitute corresponding values in the template - like mail-merge. 

We use the jinja templating engine to render templates - jinja allows some complex templating as well as simple substitution which we use for now

This PR adds a simple test action in StackStorm to send a single test email  

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:


email_templates/plaintext/shutoff_error.txt.j2
```diff
@@ -0,0 +1,15 @@
+Dear Cloud User {{username}},
+
+We are writing to inform you one or more virtual machines (VMs) associated with your account are currently in a shutoff or errored state and have not been restarted or deleted for the past X days.
```
meoflynn, line 3: I think we need to add a template variable here for the number of days a VM has been in a given state

meoflynn (changes requested):

lib/email_api/email_templates.yaml
DavidFair, line 1: Nit: `email_templates_schema.yaml`

lib/email_api/email_templates.yaml
```diff
@@ -0,0 +1,34 @@
...
+  schema:
+    username: null
+    test_message: "This is a test email"
+  html_filepath: "html/test.html.j2"
+  plaintext_filepath: "plaintext/test.txt.j2"
+
+footer:
+  schema: null
+  html_filepath: "html/footer.html"
+  plaintext_filepath: "plaintext/footer.txt"
```
DavidFair, line 26: Why do we have a dedicated footer rather than it being in wrapper? As conceptually I can now send a `footer` email or `shutoff_error` email

lib/email_api/emailer.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+    Emailer class is responsible for building and sending emails.
+    """
+
+    # Holds absolute dirpath to directory where email attachements files are stored
+    EMAIL_ATTACHMENTS_ROOT_DIR = os.path.normpath(
+        os.path.join(
+            os.path.dirname(os.path.realpath(__file__)), "../email_attachments"
+        )
+    )
```
DavidFair, line 24: Any reason we're not using [Path](https://docs.python.org/3/library/pathlib.html)?

lib/email_api/emailer.py
```diff
@@ -0,0 +1,147 @@
...
+        """
+        send multiple emails
+        :param emails: (Dict) keys are list of email addresses, values are configured EmailParams dataclass
+        :param as_html: whether to send emails using html message body (if True), or plaintext (if False) (Default True)
+        """
+        for email_to, email_params in emails.items():
+            self.send_email(email_to, email_params, as_html)
+
+    def send_email(
+        self, email_to: Tuple[str], email_params: EmailParams, as_html: bool = True
```
DavidFair, line 41: Building on the above, the data class could simply be

```
email_addresses: List[str]
```

And we just loop regardless if it's 1 or n

lib/email_api/emailer.py
```diff
@@ -0,0 +1,147 @@
...
+        :param as_html: whether to send email using html message body (if True), or plaintext (if False) (Default True)
+
+        """
+        send_to = email_params.email_from
+        if email_params.email_cc:
+            send_to += ", "
+            send_to += ", ".join(email_params.email_cc)
+
+        msg = Emailer._build_message(email_params, email_to, as_html)
+        with SMTP_SSL(
```
DavidFair, line 56: We should add logging here, to track what exact calls are being made in-case we get blamed for something over SMTP

lib/email_api/emailer.py
```diff
@@ -0,0 +1,147 @@
...
+    def _setup_smtp(server, smtp_account):
+        """
+        setup SMTP
+        :param smtp_account: SMTP config
+        """
+        server.ehlo()
+        if smtp_account.secure:
+            server.starttls()
+        if smtp_account.smtp_auth:
+            server.login(smtp_account.username, smtp_account.password)
```
DavidFair, line 72: Can you think of a case where we wouldn't use TLS/Login
AFAIK we don't have any anonymous relays, so these if conditions are extra bloat IMO

lib/email_api/template_handler.py
```diff
@@ -0,0 +1,170 @@
...
+                return safe_load(stream)
+            except YAMLError as exc:
+                raise EmailTemplateError(
+                    "could not load template metadata file"
+                ) from exc
+
+    @staticmethod
+    def _parse_template_attrs(
+        template_name: str, template_schema: Dict[str, str], given_vals: Dict[str, str]
+    ) -> Dict[str, str]:
```
DavidFair, line 54: Ditto about the types here, there's a lot of pairs of dicts and strings floating that imply to me we have correlated types that can be tied into a struct

lib/structs/email_params.py
```diff
@@ -1,28 +1,74 @@
...
+        :param dictionary: dictionary to use to instantiate dataclass
+        """
+        field_set = {field.name for field in fields(EmailParams) if field.init}
+        filtered_arg_dict = {
+            key: value for key, value in dictionary.items() if key in field_set
+        }
+        return EmailParams(**filtered_arg_dict)
+
+    @staticmethod
+    def from_template_mappings(template_mappings: TemplateMappings, **kwargs):
```
DavidFair, line 45: This should be a class at this point, and it should manage and validate it's own templates instead of being half-struct / half class

DavidFair, line 45: Since we have multiple from_dict helpers now, I think it's worth bringing in a lib such as https://github.com/Fatal1ty/mashumaro to handle this for us

DavidFair (changes requested):

lib/email_api/emailer.py
```diff
@@ -0,0 +1,147 @@
...
+        os.path.join(
+            os.path.dirname(os.path.realpath(__file__)), "../email_attachments"
+        )
+    )
+
+    def __init__(self, smtp_account: SMTPAccount):
+        self._smtp_account = smtp_account
+
+    def send_multiple_emails(
+        self, emails: Dict[Tuple[str], EmailParams], as_html: bool = True
```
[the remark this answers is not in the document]
anish-mudaraddi: sure that's fine

lib/email_api/emailer.py
```diff
@@ -0,0 +1,147 @@
...
+                        os.path.join(Emailer.EMAIL_ATTACHMENTS_ROOT_DIR, rel_filepath)
+                    ),
+                    "rb",
+                ) as file:
+                    part = MIMEApplication(file.read(), Name=filename)
+                part["Content-Disposition"] = f"attachment; filename={filename}"
+                msg.attach(part)
+            except FileNotFoundError as exp:
+                raise RuntimeError(f"Failed to attach file to email: {exp}") from exp
+        return msg
```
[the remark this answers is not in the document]
anish-mudaraddi: I recall jacob mentioned it would be good - I'll bring it up with him if he still wants it

lib/structs/email_params.py
anish-mudaraddi: just to simplify this a bit - I'm going to remove this method from this dataclass - hence it acts as just a dataclass for storing user inputted values.

anish-mudaraddi: I'll remove this function. 
`EmailParams` will just hold user inputted values 
template_mappings will be evaluated in `Emailer` when building the email body instead of here

lib/email_api/emailer.py
```diff
@@ -0,0 +1,101 @@
...
+                if email_params.email_cc:
+                    send_to += email_params.email_cc
+
+                server.sendmail(
+                    email_params.email_from,
+                    send_to,
+                    self._build_email(email_params).as_string(),
+                )
+
+    def _build_email(self, email_params: EmailParams) -> MIMEMultipart:
```
meoflynn, line 54: Could we add a docstring to this method?

tests/lib/email_api/test_emailer.py
```diff
@@ -0,0 +1,207 @@
...
+            )
+            self.assertEqual(res, mock_attach_files.return_value)
+        else:
+            self.assertEqual(res, mock_mime_multipart.return_value)
+
+    @parameterized.expand(
+        [("test_with_as_html_true", True), ("test_with_as_html_false", False)]
+    )
+    @patch("email_api.emailer.MIMEText")
+    def test_build_email_body(self, _, as_html, mock_mime_text):
```
meoflynn, line 128: Could we add a docstring to this unit test?

tests/structs/test_email_params.py
```diff
@@ -0,0 +1,27 @@
...
+from structs.email.email_params import EmailParams
+from structs.email.email_template_details import EmailTemplateDetails
+
+
+class TestEmailParams(unittest.TestCase):
+    """
+    Tests that methods in EmailParams dataclass work expectedly
+    """
+
+    def test_from_dict(self):
```
meoflynn, line 11: Could we add a docstring to this unit test?

tests/structs/test_smtp_account.py
meoflynn, line 1: Could we add docstrings to these unit tests?

meoflynn (changes requested):

DavidFair (approved):

meoflynn (approved):

PR #89, opened 2023-07-31
Add user query

meoflynn:
### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:


anish-mudaraddi:
The StackStorm Actions we added `user.search.by.property` and `user.search.by.regex` both only allow searching for stfc users - I think another PR needs to be made to add selecting of domain - since it involves making a generic change to the Manager API which will affect all Managers.

anish-mudaraddi (approved):
me and @meoflynn  pair programmed this together and we're both happy with this

PR #92, opened 2023-08-07
Add Pagination Functionality

### LINKED ISSUE #79: Query Library: refactor ServerRunner to make use of marker server-side filters - to reduce time to list servers
Currently program produces error when trying to list more than 1000 servers 
Workaround for this - list all project and aggregate servers for each project - this is slow

marker filter can speed this up- represented here:
https://docs.openstack.org/api-ref/compute/?expanded=list-servers-detail#list-servers

marker filter needs to be given pointing to the last item retrieved when running `openstack.compute.servers` to keep listing
providing this will significantly speed up listing servers

anish-mudaraddi:
### Description:

Adds pagination function for queries we expect to output a lot of values (> 1000). This significantly speeds up the process

Add ability for users to specify meta params for each Runner class - these meta params will dictate how the query will be run. 
e.g:
      - Every user query requires a domain_id for it to work on - this can be specified by meta_param - "from_domain"
      
Fixes #79 

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/user.search.by.property.yaml
+++ b/actions/user.search.by.property.yaml
@@ -68,4 +68,13 @@ parameters:
     description: "a comma-spaced list of values to search for"
     required: true
     type: array
+  from_domain:
+    description: "The user domain to search in. STFC - for STFC users. DEFAULT - for admins. OPENID - for IRIS-IAM users"
+    default: "STFC"
+    type: string
+    required: true
+    enum:
+      - "STFC"
+      - "DEFAULT"
+      - "OPENID"
 runner_type: python-script
--- a/actions/user.search.by.property.yaml
+++ b/actions/user.search.by.property.yaml
@@ -69,7 +69,7 @@ parameters:
     required: true
     type: array
   from_domain:
-    description: "The user domain to search in. STFC - for STFC users. DEFAULT - for admins. OPENID - for IRIS-IAM users"
+    description: "The user domain to search in. STFC - for STFC users. DEFAULT - for admins and service accounts. OPENID - for IRIS-IAM users"
     default: "STFC"
     type: string
     required: true
--- a/lib/openstack_query/runners/query_runner.py
+++ b/lib/openstack_query/runners/query_runner.py
@@ -27,6 +27,7 @@ class QueryRunner(OpenstackWrapperBase):
 
     # Sets the limit for getting values from openstack
     _LIMIT_FOR_PAGINATION = 1000
+    _PAGINATION_CALL_LIMIT = 1000
 
     def __init__(self, marker_prop_func: PropFunc, connection_cls=OpenstackConnection):
         OpenstackWrapperBase.__init__(self, connection_cls)
@@ -129,10 +130,14 @@ class QueryRunner(OpenstackWrapperBase):
         query_res = []
 
         curr_marker = None
+        num_calls = 1
         while True:
+            num_calls += 1
+            if num_calls > self._PAGINATION_CALL_LIMIT:
+                break
+
             prev = None
             for i, resource in enumerate(paginated_call(**paginated_filters)):
-
                 # Workaround for Endless loop error detected if querying for stfc users (via ldap)
                 # for loop doesn't seem to terminate - and outputs the same value when given "limit" and "marker"
                 if prev == resource:
--- a/lib/openstack_query/runners/query_runner.py
+++ b/lib/openstack_query/runners/query_runner.py
@@ -31,7 +31,7 @@ class QueryRunner(OpenstackWrapperBase):
 
     def __init__(self, marker_prop_func: PropFunc, connection_cls=OpenstackConnection):
         OpenstackWrapperBase.__init__(self, connection_cls)
-        self.page_marker_prop_func = marker_prop_func
+        self._page_marker_prop_func = marker_prop_func
 
     def run(
         self,
@@ -132,8 +132,12 @@ class QueryRunner(OpenstackWrapperBase):
         curr_marker = None
         num_calls = 1
         while True:
+            logger.debug("starting page loop, completed %s loops", num_calls)
             num_calls += 1
             if num_calls > self._PAGINATION_CALL_LIMIT:
+                logger.warning(
+                    "max page loops reached %s - terminating early", num_calls
+                )
                 break
 
             prev = None
@@ -141,6 +145,9 @@ class QueryRunner(OpenstackWrapperBase):
                 # Workaround for Endless loop error detected if querying for stfc users (via ldap)
                 # for loop doesn't seem to terminate - and outputs the same value when given "limit" and "marker"
                 if prev == resource:
+                    logger.warning(
+                        "duplicate entries found, likely an endless page loop - terminating early"
+                    )
                     break
 
                 query_res.append(resource)
@@ -148,7 +155,12 @@ class QueryRunner(OpenstackWrapperBase):
                 if i == self._LIMIT_FOR_PAGINATION - 1:
                     # restart the for loop with marker set
                     paginated_filters.update(
-                        {"marker": self.page_marker_prop_func(resource)}
+                        {"marker": self._page_marker_prop_func(resource)}
+                    )
+                    logger.debug(
+                        "page limit reached: %s - setting new marker: %s",
+                        self._LIMIT_FOR_PAGINATION,
+                        paginated_filters["marker"],
                     )
                     break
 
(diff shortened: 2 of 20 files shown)


lib/openstack_query/runners/user_runner.py
```diff
@@ -18,46 +18,24 @@ class UserRunner(QueryRunner):
     ServerRunner encapsulates running any openstacksdk Server commands
```
meoflynn, line 18: Could we change the reference in this docstring from `Server` to `User`?

lib/openstack_query/runners/user_runner.py
```diff
@@ -72,6 +50,39 @@ def _get_user_domain(
...
+
+        # set to default domain since neither given
+        elif "domain_id" not in filter_kwargs.keys():
+            filter_kwargs.update(
+                {"domain_id": self._get_user_domain(conn, self.DEFAULT_DOMAIN)}
+            )
+
+        return self._run_paginated_query(conn.identity.users, filter_kwargs)
+
     def _parse_subset(self, _: OpenstackConnection, subset: List[User]) -> List[User]:
```
meoflynn, line 86: Could we change the docstring so that it refers to `User` instead of `Server`?

tests/lib/openstack_query/runners/test_server_runner.py
```diff
@@ -27,134 +27,124 @@ def setUp(self):
...
-        """
-        self.conn.identity.projects.return_value = ["project1", "project2"]
-        res = self.instance._get_projects(self.conn)
-        self.conn.identity.projects.assert_called_once()
-        self.assertEqual(res, ["project1", "project2"])
-
-    def test_get_projects_with_valid_project_identifiers(self):
-        """
-        Tests _get_project method works expectedly - with valid project identifiers
+        Tests _parse_meta_params method works expectedly - with valid from_project argument
```
meoflynn, line 32: Is this unit test for testing we get lists of servers based on a project given?

meoflynn (changes requested):
I've just done an initial review, looks good to me so far

tests/lib/openstack_query/runners/test_server_runner.py
anish-mudaraddi: yes - we're testing that `parse_meta_params` works when giving valid project identifiers

lib/openstack_query/runners/query_runner.py
```diff
@@ -45,18 +48,54 @@ def run(
...
+        query_res = []
+
+        curr_marker = None
+        while True:
+            for i, server in enumerate(paginated_call(**paginated_filters)):
+                query_res.append(server)
+                # openstacksdk calls break after going over pagination limit
+                if i == self._LIMIT_FOR_PAGINATION - 1:
+                    # restart the for loop with marker set
+                    paginated_filters.update({"marker": server["id"]})
```
meoflynn, line 89: There's a reference to a server id here, is this meant to be just resource id instead?

meoflynn (changes requested):

actions/user.search.by.property.yaml
```diff
@@ -68,4 +68,13 @@ parameters:
     description: "a comma-spaced list of values to search for"
     required: true
     type: array
+  from_domain:
+    description: "The user domain to search in. STFC - for STFC users. DEFAULT - for admins. OPENID - for IRIS-IAM users"
```
DavidFair, line 72: Nit: Default Admins and Service Accounts

lib/openstack_query/runners/query_runner.py
```diff
@@ -24,8 +25,12 @@ class QueryRunner(OpenstackWrapperBase):
     Runner classes encapsulate running any openstacksdk commands
     """
 
-    def __init__(self, connection_cls=OpenstackConnection):
+    # Sets the limit for getting values from openstack
+    _LIMIT_FOR_PAGINATION = 1000
+
+    def __init__(self, marker_prop_func: PropFunc, connection_cls=OpenstackConnection):
         OpenstackWrapperBase.__init__(self, connection_cls)
+        self.page_marker_prop_func = marker_prop_func
```
DavidFair, line 33: This should just use the derived class's implementation rather than passing a function through `__init__`

lib/openstack_query/runners/query_runner.py
```diff
@@ -101,6 +110,50 @@ def run(
...
+        paginated_filters = {"limit": self._LIMIT_FOR_PAGINATION, "marker": None}
+        paginated_filters.update(server_side_filters)
+        query_res = []
+
+        curr_marker = None
+        while True:
+            prev = None
+            for i, resource in enumerate(paginated_call(**paginated_filters)):
+                # Workaround for Endless loop error detected if querying for stfc users (via ldap)
+                # for loop doesn't seem to terminate - and outputs the same value when given "limit" and "marker"
```
DavidFair, line 135: We should also add an additional safety net such as

```
num_calls = 1
while True:
    num_calls += 1
    if num_calls > _PAGINATION_CALL_LIMIT  # 1000:
        break
```

lib/openstack_query/runners/server_runner.py
```diff
@@ -23,11 +23,34 @@ class ServerRunner(QueryRunner):
...
         conn: OpenstackConnection,
-        filter_kwargs: Optional[Dict[str, str]] = None,
         from_projects: Optional[List[ProjectIdentifier]] = None,
+    ):
+        """
+        This method is a helper function that will parse a set of query meta params related to openstack server queries.
+        :param conn: An OpenstackConnection object - used to connect to openstack and parse meta params
+        """
+        try:
+            return {
```
DavidFair, line 36: IMO this try is too large, can we just do

```
try:
   project = conn.identity.find_project(...)
except:
   ...

return { ... }
```

To really restrict the scope to the only step that should fail

lib/openstack_query/runners/server_runner.py
```diff
@@ -37,89 +60,37 @@ def _run_query(
...
-        for i, project in enumerate(projects):
-            logger.debug("running query on project %s / %s", i + 1, total)
-            query_res.update(
-                {
-                    project["id"]: self._run_query_on_project(
-                        conn, project, filter_kwargs
-                    )
-                }
+        # so we can search in all projects instead of default set in clouds.yaml
+        filter_kwargs.update({"all_tenants": True})
```
DavidFair, line 70: This will only work with admin creds, can we only do it if we know we're in all projects mode effectively?

lib/openstack_query/runners/server_runner.py
```diff
@@ -37,89 +60,37 @@ def _run_query(
...
+
+        query_res = []
+        if meta_params["projects"]:
+            logger.debug(
+                "running query on each project. Found %s projects",
+                len(meta_params["projects"]),
```
DavidFair, line 76: Nit: The else condition is tiny, handle that and return, then this block can move down an indent

DavidFair (changes requested):

lib/openstack_query/runners/server_runner.py
anish-mudaraddi: pretty much every server-side filter is admin-only. Openstacksdk ignores any filter like "project_id" if you don't have admin rights - setting all_tenants here will actually produce an error rather than erroneous results

anish-mudaraddi: plus `all_tenants` is required otherwise we only find servers for a single project that's mentioned in `clouds.yaml`

lib/openstack_query/runners/query_runner.py
anish-mudaraddi: Could you explain what you mean?
we get the prop func from `prop_handler` - I can set that separately in the derived classes I guess - but we'll have some code duplication

tests/lib/openstack_query/runners/test_query_runner.py
```diff
@@ -33,14 +43,15 @@ def test_run_with_only_client_side_filter(
         mock_cloud_domain = MagicMock()
         mock_cloud_domain.name = "test"
 
-        # test with no kwargs as we test with kwargs in server-side filter
```
meoflynn, line 36: Why have we removed this comment from this unit test?

anish-mudaraddi: the test is `test_run_with_only_client_side_filter` so its sort of a redundant comment because its just saying that we're testing with no server-side filters

DavidFair (dismissed):

meoflynn (approved):

DavidFair (approved):

lib/openstack_query/runners/query_runner.py
anish-mudaraddi: picking this up in another PR - issue related #100

Issue #93, opened 2023-08-07
Query Library: Add Sorting and Grouping Functionality


anish-mudaraddi:
`sort_by` and `group_by` functions in public API need to be implemented.

Ideally it should allow setting multiple keys to sort by  - which would then apply to the output whenever an output function like to_str, to_obj_list, etc are called. e.g.

```
q1.select(prop1, prop2).where(EQUAL_TO, prop3, "some-val")
q1.sort_by(prop1, ASC).sort_by(prop2, DESC)
q1.to_list()
```
should output a list of dicts like:
```
[
    {"prop1": "a", "prop2": 3},
    {"prop1": "a", "prop2": 1},
    {"prop1": "b", "prop2": 5}
    ...
]
```

the same query should also allow grouping like so 
```
q1.group_by(prop1)
q1.to_list()
```
should output a list of list of dicts like:
```
[
    [
        {"prop1": "a", "prop2": 3},
        {"prop1": "a", "prop2": 1},
    ],
    [
        {"prop1": "b", "prop2": 5}
        ...
     ],
     ....
]

```
We should also allow grouping based on ranges like so:
```
q1.group_by("prop2", ranges=[range(1, 5), [6, 10]], include_unsorted_group=True)
```
which should output
```
[
    [
        {"prop1": "a", "prop2": 3},
        {"prop1": "a", "prop2": 1},
        {"prop1": "b", "prop2": 5}
    ]
    [
        {"prop1": "a",  "prop2": 6}
        {"prop1": "b",  "prop2": 10}
    ] 
    # unsorted group - those that don't fall into any given range
    [
       {"prop1": "a", "prop2": 0}
       {"prop1": "b", "prop2": 7}
    ]
]
```
note: we should allow users to `group_by` and `sort_by` any valid property for that resource

note: whilst its possible that users may want to `group_by` multiple times - i.e. groups within groups - we will not allow it for now

note: sorting should happen after grouping when outputting
note: when printing to table with grouping - multiple tables must be outputted

PR #97, opened 2023-08-14
ENH: Openstack query add group and sort functionality

anish-mudaraddi:
### Description:

Adds grouping and sorting functionality - PR for #93 

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/openstack_query/query_parser.py
+++ b/lib/openstack_query/query_parser.py
@@ -1,11 +1,11 @@
-from typing import List, Dict, Tuple, Union
+from typing import List, Dict, Tuple, Union, Optional, Set, Callable
 
 from enums.query.props.prop_enum import PropEnum
 from exceptions.query_property_mapping_error import QueryPropertyMappingError
 from openstack_query.handlers.prop_handler import PropHandler
 from custom_types.openstack_query.aliases import OpenstackResourceObj, PropValue
 
-from lib.exceptions.parse_query_error import ParseQueryError
+from exceptions.parse_query_error import ParseQueryError
 
 
 class QueryParser:
@@ -36,47 +36,60 @@ class QueryParser:
         Public method used to configure sorting results
         :param sort_by: tuples of property name to sort by and order "asc or desc"
         """
+        self._sort_by = set()
         for prop_name, order in sort_by:
-            self._check_prop_valid(sort_by)
-            if order.upper() not in ["ASC", "DESC"]:
+            self._check_prop_valid(prop_name)
+            order = order.upper()
+            if order not in ["ASC", "DESC"]:
                 raise ParseQueryError(
                     f"order specification given for sort-by '{order}' "
                     "is invalid, choose ASC (ascending) or DESC (descending)"
                 )
-            self._sort_by.add((sort_by, True if order == "ASC" else False))
+            self._sort_by.add((prop_name, True if order == "DESC" else False))
 
     def parse_group_by(
         self,
         group_by: PropEnum,
-        group_ranges: Dict[str, List[PropValue]],
-        include_ungrouped_results: bool,
+        group_ranges: Optional[Dict[str, List[PropValue]]] = None,
+        include_missing: Optional[bool] = False,
     ) -> None:
         """
         Public method used to configure grouping results.
         :param group_by: name of the property to group by
         :param group_ranges: a dictionary containing names of the group and list of prop values
         to select for for that group
-        :param include_ungrouped_results: a flag which if set includes an extra grouping for values
+        :param include_missing: a flag which if set includes an extra grouping for values
         that don't fall into any group specified in group_ranges
         """
         self._check_prop_valid(group_by)
+        self._group_by = group_by
         all_prop_list = set()
-        for name, prop_list in group_ranges.items():
-            self._group_mappings.update(
-                {name: lambda obj: self._prop_handler.get_prop(group_by) in prop_list}
-            )
-            all_prop_list.update(prop_list)
 
-        # if ungrouped group wanted - filter for all not in any range specified
-        if include_ungrouped_results:
-            self._group_mappings.update(
-                {
-                    "ungrouped results": lambda obj: self._prop_handler.get_prop(
-                        group_by
-                    )
-                    not in all_prop_list
-                }
-            )
+        if group_ranges:
+            for name, prop_list in group_ranges.items():
+                group_vals = tuple(prop_list)
+                self._group_mappings.update(
+                    {
+                        name: (
+                            lambda obj, lst=group_vals: self._prop_handler.get_prop(
+                                obj, group_by
+                            )
+                            in lst
+                        )
+                    }
+                )
+                all_prop_list.update(prop_list)
+
+            # if ungrouped group wanted - filter for all not in any range specified
+            if include_missing:
+                self._group_mappings.update(
+                    {
+                        "ungrouped results": lambda obj: self._prop_handler.get_prop(
+                            obj, group_by
+                        )
+                        not in all_prop_list
+                    }
+                )
 
     def run_parser(
         self, obj_list: List[OpenstackResourceObj]
@@ -85,19 +98,64 @@ class QueryParser:
         Public method used to parse query runner output - performs specified sorting and grouping
         :param obj_list: a list of openstack resource objects to parse
         """
-        parsed_obj_list = obj_list
+        if not obj_list:
+            return obj_list
+
+        # we sort first - assuming sorting is commutative to grouping
         if self._sort_by:
-
-            def sort_function(item):
-                return [
-                    item[key] if order else -item[key] for key, order in self._sort_by
-                ]
-
-            parsed_obj_list = sorted(parsed_obj_list, key=sort_function)
+            obj_list = self._run_sort(obj_list, self._sort_by)
 
         if self._group_by:
-            parsed_obj_list = {
-                name: [item for item in obj_list if map_func(item)]
-                for name, map_func in self._group_mappings.items()
-            }
-        return parsed_obj_list
+            # if group mappings not specified - make a group for each unique value found for prop
+            if not self._group_mappings:
+                self._group_mappings = self._build_unique_val_groups(
+                    obj_list, group_by_prop=self._group_by
+                )
+            obj_list = self._run_group_by(obj_list, self._group_mappings)
+
+        return obj_list
+
+    def _run_sort(
+        self,
+        obj_list: List[OpenstackResourceObj],
+        sort_by_specs: Set[Tuple[PropEnum, bool]],
+    ) -> List[OpenstackResourceObj]:
+        for key, reverse in reversed(tuple(sort_by_specs)):
+            obj_list.sort(
+                key=lambda obj: self._prop_handler.get_prop(obj, key), reverse=reverse
+            )
+        return obj_list
+
+    def _build_unique_val_groups(
+        self,
+        obj_list: List[OpenstackResourceObj],
+        group_by_prop: PropEnum,
+    ) -> Dict[str, Callable[[OpenstackResourceObj], bool]]:
+
+        unique_vals = set(
+            self._prop_handler.get_prop(obj, group_by_prop) for obj in obj_list
+        )
+        group_mappings = {}
+
+        # build groups
+        for val in unique_vals:
+            group_mappings.update(
+                {
+                    f"{group_by_prop.name} with value {val}": lambda obj, test_val=val: self._prop_handler.get_prop(
+                        obj, group_by_prop
+                    )
+                    == test_val
+                }
+            )
+
+        return group_mappings
+
+    @staticmethod
+    def _run_group_by(
+        obj_list: List[OpenstackResourceObj],
+        group_mappings: Dict[str, Callable[[OpenstackResourceObj], bool]],
+    ) -> Dict[str, List[OpenstackResourceObj]]:
+        return {
+            name: [item for item in obj_list if map_func(item)]
+            for name, map_func in group_mappings.items()
+        }
--- a/lib/openstack_query/query_parser.py
+++ b/lib/openstack_query/query_parser.py
@@ -1,9 +1,12 @@
 from typing import List, Dict, Tuple, Union, Optional, Callable
+import logging
 from collections import OrderedDict
 from enums.query.props.prop_enum import PropEnum
 from openstack_query.handlers.prop_handler import PropHandler
 from custom_types.openstack_query.aliases import OpenstackResourceObj, PropValue
 from exceptions.query_property_mapping_error import QueryPropertyMappingError
+
+logger = logging.getLogger(__name__)
 
 
 class QueryParser:
@@ -17,6 +20,10 @@ class QueryParser:
         self._sort_by = {}
         self._group_by = None
         self._group_mappings = {}
+
+    @property
+    def group_by_prop(self):
+        return self._group_by
 
     def _check_prop_valid(self, prop: PropEnum) -> None:
         """
@@ -35,9 +42,21 @@ class QueryParser:
         :param sort_by: tuples of property name to sort by and order "asc or desc"
         """
         self._sort_by = {}
-        for prop_name, order in sort_by:
-            self._check_prop_valid(prop_name)
-            self._sort_by.update({prop_name: order})
+        for prop, order in sort_by:
+            self._check_prop_valid(prop)
+            order_log_str = "DESC" if order else "ASC"
+            logging.debug(
+                "adding sorting params: %s, order: %s", prop.name, order_log_str
+            )
+            self._sort_by.update({prop: order})
+
+        logging.debug(
+            "sort params: %s",
+            "\n\t".join(
+                f"{prop}, order: {'DESC' if order else 'ASC'}"
+                for prop, order in self._sort_by.items()
+            ),
+        )
 
     def parse_group_by(
         self,
@@ -58,6 +77,7 @@ class QueryParser:
         all_prop_list = set()
 
         if group_ranges:
+            logger.debug("creating filter functions for specified group ranges")
             for name, prop_list in group_ranges.items():
                 group_vals = tuple(prop_list)
                 self._group_mappings.update(
@@ -74,6 +94,7 @@ class QueryParser:
 
             # if ungrouped group wanted - filter for all not in any range specified
             if include_missing:
+                logger.debug("creating filter function for ungrouped group")
                 self._group_mappings.update(
                     {
                         "ungrouped results": lambda obj: self._prop_handler.get_prop(
@@ -100,6 +121,10 @@ class QueryParser:
         if self._group_by:
             # if group mappings not specified - make a group for each unique value found for prop
             if not self._group_mappings:
+                logger.info(
+                    "no group ranges specified - grouping by unique values of %s property",
+                    self._group_by.name,
+                )
                 self._group_mappings = self._build_unique_val_groups(
                     obj_list, group_by_prop=self._group_by
                 )
@@ -118,8 +143,13 @@ class QueryParser:
         :param sort_by_specs: A dictionary of property to sort by as key and value as boolean representing order
             - descending - True, ascending - False
         """
-
-        for sort_key, reverse in reversed(tuple(sort_by_specs.items())):
+        logger.debug("running multi-sort")
+        sort_num = len(sort_by_specs)
+        for i, (sort_key, reverse) in enumerate(
+            reversed(tuple(sort_by_specs.items())), 1
+        ):
+            logger.debug("running sort %s / %s", i, sort_num)
+            logger.debug("sorting by: %s, reverse=%s", sort_key, reverse)
             obj_list.sort(
                 key=lambda obj, sk=sort_key: self._prop_handler.get_prop(obj, sk),
                 reverse=reverse,
@@ -136,6 +166,11 @@ class QueryParser:
         unique_vals = OrderedDict(
             {self._prop_handler.get_prop(obj, group_by_prop): None for obj in obj_list}
         )
+        logger.debug(
+            "unique values found %s - each will become a group",
+            ",".join(f"{val}" for val in unique_vals),
+        )
+
         group_mappings = {}
 
         # build groups
--- a/lib/openstack_query/query_parser.py
+++ b/lib/openstack_query/query_parser.py
@@ -117,37 +117,26 @@ class QueryParser:
 
         # we sort first - assuming sorting is commutative to grouping
         if self._sort_by:
-            obj_list = self._run_sort(obj_list, self._sort_by)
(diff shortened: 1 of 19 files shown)


lib/openstack_query/query_parser.py
```diff
@@ -0,0 +1,197 @@
...
+                key=lambda obj, sk=sort_key: self._prop_handler.get_prop(obj, sk),
+                reverse=reverse,
+            )
+        return obj_list
+
+    def _build_unique_val_groups(
+        self,
+        obj_list: List[OpenstackResourceObj],
+        group_by_prop: PropEnum,
+    ) -> Dict[str, Callable[[OpenstackResourceObj], bool]]:
```
meoflynn, line 163: Could we add docstrings to this method?

lib/openstack_query/query_parser.py
```diff
@@ -0,0 +1,197 @@
...
+            )
+
+        return group_mappings
+
+    @staticmethod
+    def _run_group_by(
+        obj_list: List[OpenstackResourceObj],
+        group_mappings: Dict[str, Callable[[OpenstackResourceObj], bool]],
+    ) -> Dict[str, List[OpenstackResourceObj]]:
+        return {
```
meoflynn, line 194: Could we add a docstring to this method?

tests/structs/test_query_output_details.py
```diff
@@ -15,37 +15,48 @@ def setUp(self) -> None:
...
+        else:
+            self.assertEqual(
+                out.sort_by,
+                [
+                    (ServerProperties.SERVER_NAME, False),
+                    (ServerProperties.SERVER_ID, False),
+                ],
+            )
+
+        # TODO allow setting these using form kwargs
```
meoflynn, line 60: Is this for a future PR?

meoflynn (changes requested):

tests/structs/test_query_output_details.py
anish-mudaraddi: removed this comment and added an issue to make it clearer #102

lib/openstack_query/query_parser.py
```diff
@@ -0,0 +1,211 @@
...
+        if self._group_by:
+            # if group mappings not specified - make a group for each unique value found for prop
+            if not self._group_mappings:
+                logger.info(
+                    "no group ranges specified - grouping by unique values of %s property",
+                    self._group_by.name,
+                )
+                self._group_mappings = self._build_unique_val_groups(
+                    obj_list, group_by_prop=self._group_by
+                )
```
DavidFair, line 131: Nit: The `group_by` method should do this, not `run_parser`

lib/openstack_query/query_parser.py
```diff
@@ -0,0 +1,211 @@
...
+
+        logging.debug(
+            "sort params: %s",
+            "\n\t".join(
+                f"{prop}, order: {'DESC' if order else 'ASC'}"
+                for prop, order in self._sort_by.items()
+            ),
+        )
+
+    def parse_group_by(
```
DavidFair, line 62: There's enough grouping stuff I'd argue that it's worth just using the `refactor -> extract class` to extract out `OutputGrouper` or something similar from `query_parser`

lib/structs/query/query_output_details.py
```diff
@@ -1,34 +1,83 @@
...
-            output_type = QueryOutputTypes.from_string(kwargs["output_type"])
-        return QueryOutputDetails(properties_to_select=props, output_type=output_type)
+
+        # set the default for props to select as all available props
+        output_details = {
+            **{
+                "properties_to_select": list(prop_cls),
+            },
+            **kwargs,
+        }
```
DavidFair, line 58: Why do we get `properties_to_select` put into a dictionary to just unpack it anyway into the parent?

I would argue that:
`output_details = {"properties_to_select": list(prop_cls)}`

is easier to read, or am I misunderstanding what it's doing?

lib/structs/query/query_output_details.py
```diff
@@ -1,34 +1,83 @@
...
+
+        if properties_to_select:
+            output_details.update(
+                {
+                    "properties_to_select": [
+                        prop_cls.from_string(prop) for prop in properties_to_select
+                    ]
+                }
+            )
```
DavidFair, line 67: Do this as two steps to help readability:
```
new_props = [prop_cls.from_string() ....]
output_details["properties_to_select] = new_props
```

(Using `[]` is equivalent in this context to doing an update, since we overwrite it regardless)

tests/lib/openstack_query/managers/test_query_manager.py
```diff
@@ -86,32 +90,62 @@ def test_build_and_run_query_with_runner_params(
...
-        )
+        if not mock_output_details:
+            # just assume valid output details is made from calling from_kwargs
+            mock_query_output_details.from_kwargs.return_value = MOCKED_OUTPUT_DETAILS
+
+        with patch(
+            "openstack_query.managers.query_manager.QueryManager._populate_output_params"
+        ) as mock_populate_output_params:
+            with patch(
+                "openstack_query.managers.query_manager.QueryManager._get_query_output"
```
DavidFair, line 115: This many patches for a single method call is a very strong indicator it's doing too much

The class should break these down into public API steps with a top level one eg.

```
class QueryManager:

   # args omitted for brevity
    def build_query():
    def execute_query():
    def output_query():

    def run_full_query():
        built = self.build_query()
        result = self.execute_query(built)
        return self.output_query(result)
```

Since it's just as valid for a unit test, or an external caller to call self.output_query even if there's a helper method to stitch it up.

Our tests then simply test each step in isolation and test that the top level method calls all of those methods which we replace with mocks in `test_run_full_query`

tests/lib/openstack_query/managers/test_query_manager.py
```diff
@@ -86,32 +90,62 @@ def test_build_and_run_query_with_runner_params(
...
+                    )
+
+        if mock_preset_details:
+            self.query.where.assert_called_once_with(
+                preset=MOCKED_PRESET_DETAILS.preset,
+                prop=MOCKED_PRESET_DETAILS.prop,
+                **MOCKED_PRESET_DETAILS.args,
+            )
+        else:
+            self.query.where.assert_not_called()
```
DavidFair, line 146: If you see an if - else on a unit test then it's a sign you've got two test-cases smashed into one

tests/lib/openstack_query/managers/test_query_manager.py
```diff
@@ -129,23 +163,35 @@ def test_get_query_output_raises_error(self):
...
-            preset=MOCKED_PRESET_DETAILS.preset,
-            prop=MOCKED_PRESET_DETAILS.prop,
-            **MOCKED_PRESET_DETAILS.args,
-        )
+        if mock_output_details.sort_by:
+            self.query.sort_by.assert_called_once_with(*mock_output_details.sort_by)
+        else:
+            self.query.sort_by.assert_not_called()
+
+        if mock_output_details.group_by:
```
DavidFair, line 187: Ditto, there's three test cases here for each of the group_by / sort_by ...etc.
If there's common boilerplate extract it to a helper private method in the test. But tests like this tend to be very inflexible to future changes

tests/lib/openstack_query/test_query_methods.py
```diff
@@ -110,66 +112,81 @@ def test_where_with_kwargs(self):
...
+        mock_server_filters = self.mock_builder.server_side_filters
+        mock_parser_results = ["obj1", "obj2"]
+        self.mock_parser.run_parser.return_value = mock_parser_results
 
         if not mock_kwargs:
             mock_kwargs = {}
 
-        res = self.instance.run("test-account", mock_from_subset, **mock_kwargs)
-        mock_query_runner.run.assert_called_once_with(
+        _ = self.instance.run("test-account", mock_from_subset, **mock_kwargs)
```
DavidFair, line 129: We should check the return, as this test would pass with

```
def run(...)
   self.runner.run(...)
   return None
```

tests/lib/openstack_query/test_query_methods.py
```diff
@@ -110,66 +112,81 @@ def test_where_with_kwargs(self):
...
+        self.mock_output.generate_output.assert_called_once_with(["obj1", "obj2"])
 
-        self.assertEqual(res, self.instance)
+    @parameterized.expand(
+        [
+            ("parser outputs grouped results", True),
+            ("parser outputs list results", False),
+        ]
+    )
+    def test_run_with_parsing(self, _, parse_output_as_dict):
```
DavidFair, line 148: Ditto, these are two test cases and IMO the parameterized and if-elses match a duplicated test case since we only have two cases

Plus it's easier to reason with two tests called `test_run_with_grouped_parsing` and `test_run_with_list_results`

tests/lib/openstack_query/test_query_parser.py
```diff
@@ -0,0 +1,415 @@
+import unittest
+from unittest.mock import MagicMock, call, patch
+from parameterized import parameterized
+
+from openstack_query.query_parser import QueryParser
+from tests.lib.openstack_query.mocks.mocked_props import MockProperties
+
+# pylint:disable=protected-access
```
DavidFair, line 8: I'm tempted to start flagging these, if we're having to do protected access then either we have a hidden inner class doing processing that needs splitting out, or the API needs splitting (like I described with the run pipeline stuff)

tests/lib/openstack_query/test_query_parser.py
```diff
@@ -0,0 +1,415 @@
...
+        self.instance.parse_group_by(
+            group_by="prop1",
+            group_ranges=mock_group_ranges,
+            include_missing=include_missing,
+        )
+
+        mock_check_prop_valid.assert_called_once_with("prop1")
+        if not include_missing:
+            self.assertEqual(
+                self.instance._group_mappings.keys(), mock_group_ranges.keys()
```
DavidFair, line 88: We should just add a getter for this, I can't think of a reason why this needs to be a hidden implementation detail

tests/lib/openstack_query/test_query_parser.py
```diff
@@ -0,0 +1,415 @@
...
+    @patch("openstack_query.query_parser.QueryParser._run_sort")
+    @patch("openstack_query.query_parser.QueryParser._run_group_by")
+    def test_run_parser_with_sort_and_group(self, mock_run_group_by, mock_run_sort):
+        """
+        Tests that run_parser functions expectedly - when giving both group_by and sort_by
+        Should call run_sort method and then run_group_by on that output, then return
+        """
+        self.instance._sort_by = (MockProperties.PROP_1, False)
+        self.instance._group_by = MockProperties.PROP_1
+        self.instance._group_mappings = {"group1": "some-mapping_func"}
```
DavidFair, line 185: Ditto about sort_by and group_by having getters

For `group_mappings" we should call the public method `setup_groupings` instead of circumventing it. If it's a hassle to do this it either means our API needs updating or we need a helper method in the test or a fake object....etc.

I'm sure I've missed other places we do this, but it's more me trying to get us out of the habit of using protected by default and "dog food" the interface we're providing from objects in tests

tests/lib/openstack_query/test_query_parser.py
```diff
@@ -0,0 +1,415 @@
...
+        """
+
+        mock_obj_list = [
+            {"arg1": "a", "arg2": 2},
+            {"arg1": "c", "arg2": 4},
+            {"arg1": "d", "arg2": 3},
+            {"arg1": "b", "arg2": 1},
+        ]
+
+        res = self.instance._run_sort(mock_obj_list, sort_by_specs=mock_sort_by_specs)
```
DavidFair, line 256: `_run_sort` implies an inner class for output parsing, either we should add a `sort_results` to the query_parser or split this out

tests/lib/openstack_query/test_query_parser.py
```diff
@@ -0,0 +1,415 @@
...
+                [
+                    {"arg1": "b", "arg2": 1},
+                    {"arg1": "a", "arg2": 2},
+                    {"arg1": "d", "arg2": 3},
+                    {"arg1": "c", "arg2": 4},
+                ],
+            ),
+            (
+                "integer key descending",
+                {"arg2": True},
```
DavidFair, line 232: Having seen this in use I think we should change the API as `True` doesn't mean anything

IMO the args should be something like:

```
class SortOrder(auto)
   ASC = auto()
   DEC = auto()
```

```
def sort(result, prop_order: List[str], order: SortOrder)
```

Then for the call is just
```
sort(results, ["name", "age"], SortOrder.ASC)
```

And it's much clearer than:
```
sort(results, {"name": true}
```

tests/lib/openstack_query/test_query_parser.py
```diff
@@ -0,0 +1,415 @@
...
+    @parameterized.expand(
+        [
+            (
+                "string key ascending",
+                {"arg1": False},
+                [
+                    {"arg1": "a", "arg2": 2},
+                    {"arg1": "b", "arg2": 1},
+                    {"arg1": "c", "arg2": 4},
+                    {"arg1": "d", "arg2": 3},
```
DavidFair, line 207: Rather than pre-set these I would just do a sort in all of these tests too, something like

```
# Note I would change this to just a string or list of strings, so we can just do mock_sort[0]
prop_name = mock_sort_by_specs.keys()[0]
expected_results = sort(mock_input, key = lambda x: x.prop_name
```

Since it's a trivial bug if we get the sort order wrong to find and fix, and it removes a shed-load of boilerplate

tests/lib/openstack_query/test_query_parser.py
```diff
@@ -0,0 +1,415 @@
...
+        self.assertEqual(list(res.keys()), expected_group_names)
+
+        # check that groups are unique
+        for i, key in enumerate(res.keys()):
+            for j, test_val in enumerate(expected_unique_vals):
+                test_bool = res[key]({mock_group_by_prop: test_val})
+
+                # if the function is run with prop matching expected val - should return True
+                if i == j:
+                    self.assertTrue(test_bool)
```
DavidFair, line 390: This is a really complex way to test this, I would simply do:

```
assert len(set(res) == len(expected_group_names)
assert all(i for i in rest in expected_group_names)
```

Also since all the values don't overlap (e.g. a,b,c and 1,2,3) this wouldn't remove anything AFAIK?

tests/structs/test_query_output_details.py
```diff
@@ -15,37 +15,47 @@ def setUp(self) -> None:
...
-            ("properties and output type given", True, "TO_LIST"),
+            ("none given", {}),
+            ("props given", {"properties_to_select": ["server_id", "server_name"]}),
+            ("output type given", {"output_type": "TO_LIST"}),
+            ("group by given", {"group_by": "server_name"}),
+            ("sort by given", {"sort_by": ["server_name", "server_id"]}),
         ]
     )
-    def test_from_kwargs(self, _, set_mock_props, mock_output_type):
+    def test_from_kwargs(self, _, mock_kwargs):
```
DavidFair, line 25: Don't use parametrized to name test cases, instead move the boilerplate to a helper private method and create multiple test cases for it instead

```
def test_from_kwargs_group_by_given(self):
   self._run_from_kwargs_case({"group_by": "server_name"})

def test_from_kwargs_sort_given(self):
   self._run_from_kwargs_case({"sort_by": ["server_name", "server_id"]})
```

There's a few benefits:
- Easier to debug individual test cases
- Easier to read than following params through parametrized 
- The CLI will print the proper name for the test 
- and that name you can easily grep for as entire string, rather than a sub-string

tests/structs/test_query_output_details.py
```diff
@@ -15,37 +15,47 @@ def setUp(self) -> None:
...
+        if "sort_by" not in mock_kwargs.keys():
+            self.assertEqual(out.sort_by, None)
+        else:
+            self.assertEqual(
+                out.sort_by,
+                [
+                    (ServerProperties.SERVER_NAME, False),
+                    (ServerProperties.SERVER_ID, False),
+                ],
+            )
```
DavidFair, line 58: Ditto about this being cases, building on above just return and compare in the test case for only the thing your testing. I.e. don't check `group_by` in the `sort_by` cases

```
def test_from_kwargs_sort_given(self):
   out = self._run_from_kwargs_case({"sort_by": ["server_name", "server_id"]})
   
  self.assertEqual(out.sort_by, [ ... ])

def test_from_kwargs_sort_missing(self):
   out = self._run_from_kwargs_case({})
   
  self.assertEqual(out.sort_by, None)
```

If changing one implies another (e.g. SortBy changes GroupBy) add a third class of test cases to check this but it makes testing with state machines much easier

```
def test_setting_sort_by_resets_group_by
```

tests/lib/openstack_query/managers/test_query_manager.py
anish-mudaraddi: moving this into separate PR - issue #103

tests/lib/openstack_query/test_query_parser.py
anish-mudaraddi: so we need this because we need to set hidden attributes `_sort_by` and `_group_by`

These are set by the methods `parse_sort_by` and `parse_group_by` 
- called when users build the query using `sort_by`/`group_by` methods
 
Then these attributes are then used in `run_sort_by` and `run_group_by` 
- happens in the run() step 

access to these hidden attributes is needed to test that the "build step" methods actually set the internal attributes properly and also to set the hidden attributes to expected values to test "run step" methods

lib/openstack_query/query_parser.py
anish-mudaraddi: neither `parse_group_by` or `QueryMethods.group_by` methods can build unique value groups as the query needs to be run first. I can move this out into `run_group_by` method

lib/openstack_query/query_parser.py
anish-mudaraddi: I think it is fine for now - but I made a separate issue

lib/structs/query/query_output_details.py
anish-mudaraddi: it's packing both `properties_to_select` and any `kwargs` 
- this is to allow for other classes/methods to provide properties like `group_ranges` and `include_group_ranges` that we just pass through without needing any validation

tests/lib/openstack_query/test_query_parser.py
anish-mudaraddi: no the API works like this:
```
sort(  ("name", True), ("age", True)  )
```
you're passing tuples not dicts 
We need to pass tuples because we want to allow users to sort by multiple keys in different orders

Currently because stackstorm is a pain I've defaulted to sorting every key by DESC 

The reason we use True and False is to emulate python `sort` which takes a `reverse` boolean flag - I can change it to `SortOrder` - that makes sense - I'll create an issue for it

tests/lib/openstack_query/test_query_parser.py
anish-mudaraddi: seems dubious - you're relying on `i` being in the group name but not actually checking if the function that it maps to works

tests/lib/openstack_query/test_query_parser.py
anish-mudaraddi: a getter for what? The attribute `group_mappings`?
Why do we need a getter? We only use it in `ParseQuery`

tests/lib/openstack_query/test_query_methods.py
anish-mudaraddi: actually the `test_run_with_list_results` isn't needed since its the default and already being tested for free in other tests

tests/lib/openstack_query/test_query_parser.py
anish-mudaraddi: we can split this two parts - `QueryParserBuilder` and `QueryParserRunner`
but that will mean we should probably also split `QueryOutput` into `QueryOutputBuilder` and `QueryOutputRunner`

I'll create an issue for this

tests/lib/openstack_query/test_query_parser.py
DavidFair: Then along those lines, why can't we mark them as getters and use the public API?

I'm just concerned we're slipping into treating everything as public, as the private / public separation is another way to force people to consider their classes and how they're interacted with

I'll leave this for the moment and circle back to it in the future

tests/lib/openstack_query/test_query_parser.py
DavidFair: It's basically this in a nutshell:
https://softwareengineering.stackexchange.com/a/380296

It's not going to cause problems, but it's effectively us saying we don't need to consider how this object works from a public API point of view or the fact it's large enough that we can't observe it through a public method

tests/lib/openstack_query/test_query_parser.py
DavidFair: My concern is us doing two steps in our tests:

```
        res = self.instance._build_unique_val_groups(
            obj_list,
            mock_prop_enum,
        )
```

Then instead of directly checking `res` we appear to mutate it, I guess it's this syntax I'm really struggling with:

`test_bool = res[key]({mock_group_by_prop: test_val})`

I can see we iterate the key, so we have something like a group of things

But then it's this bit where we effectively do

`res[key]({"arg1: ["a", "b", "c"]})` that's causing my problems

IMO it's too clever for it's own good, I've took a few attempts to parse what's going on (for example what are we calling with the `(DICT)`, since a dictionary should return a data structure not another callable?

Whilst I understand it's an internal data structure I worry we're writing code that's so dense in meaning only a few people will be able to maintain it?

lib/openstack_query/query_parser.py
DavidFair: Yeah I'm thrown by us having the methods which prepare the grouping before the query, and the methods that actually do the grouping after in the same file

The ones for prepping are fine IMO (since it's a parser) but the ones that actively group should be elsewhere. But that's just my 2 cents

tests/lib/openstack_query/test_query_parser.py
anish-mudaraddi: that makes sense - never considered tests count as consumers of the class :+1:

tests/lib/openstack_query/test_query_parser.py
anish-mudaraddi: I've simplified this test - now we just run the group mappings and check the output matches expected

tests/lib/openstack_query/test_query_parser.py
anish-mudaraddi: I'll split this out into a separate PR - since we do this in lots of places #108

tests/lib/openstack_query/test_query_parser.py
anish-mudaraddi: see #104 for future PR

DavidFair (approved):

PR #109, opened 2023-08-31
MAINT: Add pytest to requirements

DavidFair:
### Description:
This will allow us to write against pytest instead of nose going forwards, so we have less to convert in the long-run


DavidFair:
Closing for smaller PRs

PR #112, opened 2023-08-31
MAINT: Switch away from nose in lib tests

DavidFair:
Nose is unsupported, and has broken with Python 3.11+, so switch away from this

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/email_api/emailer.py
+++ b/lib/email_api/emailer.py
@@ -30,9 +30,9 @@ class Emailer:
         Path(__file__).resolve().parent.parent.parent / "email_attachments"
     )
 
-    def __init__(self, smtp_account: SMTPAccount):
-        self._smtp_account = smtp_account
-        self._template_handler = TemplateHandler()
+    def __init__(self, smtp_account: SMTPAccount, template_handler=TemplateHandler()):
+        self.smtp_account = smtp_account
+        self._template_handler = template_handler
 
     def send_emails(self, emails: List[EmailParams]):
         """
@@ -43,7 +43,7 @@ class Emailer:
         logger.debug("connecting to SMTP server")
 
         with SMTP_SSL(
-            self._smtp_account.server, self._smtp_account.port, timeout=60
+            self.smtp_account.server, self.smtp_account.port, timeout=60
         ) as server:
             server.ehlo()
             logger.info("SMTP server connection established")
@@ -73,14 +73,14 @@ class Emailer:
                 server.sendmail(
                     email_params.email_from,
                     send_to,
-                    self._build_email(email_params).as_string(),
+                    self.build_email_header(email_params).as_string(),
                 )
 
             logger.info(
                 "sending complete - time elapsed: %s seconds", time.time() - start
             )
 
-    def _build_email(self, email_params: EmailParams) -> MIMEMultipart:
+    def build_email_header(self, email_params: EmailParams) -> MIMEMultipart:
         """
         Helper function to setup email as MIMEMultipart
         :param email_params: A dataclass holding parameters for building an email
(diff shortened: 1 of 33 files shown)


meoflynn (changes requested):

lib/email_api/emailer.py
```diff
@@ -73,34 +73,33 @@ def send_emails(self, emails: List[EmailParams]):
...
-                    self._build_email(email_params).as_string(),
+                    self.build_email_header(email_params).as_string(),
                 )
 
             logger.info(
                 "sending complete - time elapsed: %s seconds", time.time() - start
             )
 
-    def _build_email(self, email_params: EmailParams) -> MIMEMultipart:
+    def build_email_header(self, email_params: EmailParams) -> MIMEMultipart:
```
anish-mudaraddi, line 83: Why did you change this? 
Technically this is not just building email header since it calls `build_email_body` in the method

tests/lib/test_openstack_security_groups.py
```diff
@@ -115,42 +112,48 @@ def test_create_security_group_forwards_result(self):
...
-    def test_create_rule_throws_non_numeric(self, start_port, end_port):
+    def test_create_rule_throws_non_numeric(self):
         """
         Tests that non-numeric inputs throw
         """
         mocked_details = NonCallableMock()
-        mocked_details.port_range = (start_port, end_port)
-        with assert_raises(ValueError):
-            self.instance.create_security_group_rule(NonCallableMock(), mocked_details)
+        for start, end in (1, "a"), ("d", 2), ("*a", "a"), ("*", "*1"):
```
anish-mudaraddi, line 151: why not use `pytest.mark.parametrize`? 
so we can see which one fails if the test fails?

tests/lib/test_openstack_security_groups.py
```diff
@@ -115,42 +112,48 @@ def test_create_security_group_forwards_result(self):
...
-    def test_create_rule_throws_for_missing_port(self, start_port, end_port):
+    def test_create_rule_throws_for_missing_port(self):
         """
         Tests that missing port starting values throw
         """
         mocked_details = NonCallableMock()
-        mocked_details.port_range = (start_port, end_port)
-        with assert_raises(ValueError):
-            self.instance.create_security_group_rule(NonCallableMock(), mocked_details)
+        for start, end in (0, None), (None, 0), (None, None):
```
anish-mudaraddi, line 130: why not use pytest.mark.parametrize?
so we can see which one fails if the test fails?

tests/lib/test_openstack_identity.py
```diff
@@ -198,27 +198,26 @@ def test_delete_project_handles_resource_not_found(self):
...
-    @raises(MissingMandatoryParamError)
     def test_find_project_raises_missing_project_identifier(self):
         """
         Tests that a missing mandatory parameter is missing if a whitespace
         or empty string is passed
         """
-
-        self.instance.find_project("set", " ")
+        with self.assertRaises(MissingMandatoryParamError):
+            self.instance.find_project("set", "")
```
anish-mudaraddi, line 220: This used to be passing `" "` - I'm assuming passing `""` is the same - just checking

lib/email_api/emailer.py
DavidFair: Originally I wanted to split it up, but I've reverted it and called it `build_email` (so we're not going in via protected)

anish-mudaraddi:
closing this as it's been done

PR #115, opened 2023-09-05
bug: create internal project fix

anish-mudaraddi:
this action is required for create internal project orquesta workflow to complete

erroneously removed (deprecated by moving to `todo` directory)  and not caught because stackstorm doesn't delete actions upon a re-install

This action should be deprecated in favour of query library - but this is a later job


DavidFair (approved):
Chatted with Anish, he's tested this against 3.8 and shown it working

PR #116, opened 2023-09-11
BUG: fix project create yaql type issue

anish-mudaraddi:
creating a project fails when admin or stfc user box left empty - null isn't a list so the workflow fails - change to using an empty list as default value
sorry - didn't catch this in the previous bugfix PR

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/workflow.project.create.internal.yaml
+++ b/actions/workflow.project.create.internal.yaml
@@ -30,14 +30,14 @@ parameters:
     type: boolean
   admin_user_list:
     required: false
-    default: null
+    default: []
     type: array
     items:
       type: string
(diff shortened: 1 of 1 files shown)


DavidFair (approved):
LGTM, don't apologise that's the whole point of iterative design and feedback

One small thing to check, does this use a factory to generate the list or return the same list each time? 

If the latter, because of the fun way Python does references, assigning anything to the default empty list from Stackstorm means all the following lists will also contain those items too

Similar to the pattern here: https://stackoverflow.com/a/366430

meoflynn (approved):

PR #120, opened 2023-09-15
MAINT: Port tests for lib/openstack_query/handlers

DavidFair:
### Description:

As part of the ongoing pytest migration, port
lib/openstack_query/handlers to use pytest

The test for `handler_base` has been removed, as there was nothing to test following an earlier PR which changed it to effectively an interface

### DIFF
--- a/tests/lib/openstack_query/handlers/test_client_side_handler.py
+++ b/tests/lib/openstack_query/handlers/test_client_side_handler.py
@@ -14,9 +14,16 @@ from tests.lib.openstack_query.mocks.mocked_props import MockProperties
 
 @pytest.fixture(name="mock_filter_func")
 def filter_func_fixture():
-    # Required to mock the filter_func
+    """
+    Stubs out something callable for the method
+    under test. We don't want any side effects from
+    this function, so we just pass.
+    """
+
     # pylint:disable=unused-argument
     def _mock_filter_func(prop, *args, **kwargs):
+        # We're not testing filter functions in this test, so
+        # pass instead of returning a value
         pass
 
     return _mock_filter_func
@@ -24,6 +31,11 @@ def filter_func_fixture():
 
 @pytest.fixture(name="instance")
 def instance_fixture(mock_filter_func):
+    """
+    Returns an instance with a mocked filter function
+    injects, and various properties to query presets
+    pre-defined
+    """
     _filter_function_mappings = {
         MockQueryPresets.ITEM_1: ["*"],
         MockQueryPresets.ITEM_2: [MockProperties.PROP_1, MockProperties.PROP_2],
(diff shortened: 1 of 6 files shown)


tests/lib/openstack_query/handlers/test_client_side_handler.py
```diff
@@ -12,293 +12,309 @@
 # pylint:disable=protected-access
 
 
-class ClientSideHandlerBaseTests(unittest.TestCase):
+@pytest.fixture(name="mock_filter_func")
+def filter_func_fixture():
+    # Required to mock the filter_func
+    # pylint:disable=unused-argument
+    def _mock_filter_func(prop, *args, **kwargs):
+        pass
```
meoflynn, line 20: Why does this method not return a value?

tests/lib/openstack_query/handlers/test_client_side_handler.py
```diff
@@ -12,293 +12,309 @@
 # pylint:disable=protected-access
 
 
-class ClientSideHandlerBaseTests(unittest.TestCase):
+@pytest.fixture(name="mock_filter_func")
+def filter_func_fixture():
```
meoflynn, line 16: Could we add a docstring to this method?

tests/lib/openstack_query/handlers/test_client_side_handler.py
```diff
@@ -12,293 +12,309 @@
...
+    # Required to mock the filter_func
+    # pylint:disable=unused-argument
+    def _mock_filter_func(prop, *args, **kwargs):
+        pass
+
+    return _mock_filter_func
+
+
+@pytest.fixture(name="instance")
+def instance_fixture(mock_filter_func):
```
meoflynn, line 26: Could we add a docstring to this method?

tests/lib/openstack_query/handlers/test_client_side_handler.py
```diff
@@ -12,293 +12,309 @@
...
+)
+def test_check_filter_func_valid_filters(input_val, instance):
+    """
+    Tests that check_filter_func method works expectedly - for filter_func which takes extra params
+    returns True only if args match expected args required by client-side filter function
+    """
+
+    # pylint:disable=unused-argument
+    def _mock_filter_func(prop, arg1: int, arg2: str = "some-default", **kwargs):
+        return None
```
meoflynn, line 238: Why does this method return `None`?

tests/lib/openstack_query/handlers/test_client_side_handler_datetime.py
```diff
@@ -33,147 +30,138 @@ def setUp(self):
...
+            NonCallableMock(),
+            NonCallableMock(),
+            NonCallableMock(),
+            NonCallableMock(),
+        )
+        mock_datetime.strptime.return_value.timestamp.return_value = (
+            prop_time  # Prop timestamp
+        )
+        mock_time_get_timestamp.return_value = (
+            user_time  # The user's selected timeframe
```
meoflynn, line 65: Doesn't `prop_time` and `user_time` need to be `NonCallableMock()` as well?

meoflynn (changes requested):

tests/lib/openstack_query/handlers/test_client_side_handler_datetime.py
DavidFair: This is a small static method to remove the duplication in the tests. 

Since it runs for both the younger than or older than we need to pick real values for `<` or `>` , since `NonCallableMock > NonCallableMock` potentially doesn't make sense

I had to double check it was intended and not a bug / thing we could improve, but I can't think how to avoid it :thinking:

meoflynn (approved):

PR #122, opened 2023-09-15
Prop enum and *Query refactor

DavidFair:
### Description:
- Unify `from_string` into a single place in prop_enum
- Simplify init and add tests to *Query Objects

Another PR will extract `QueryWrapper` out to a factory, but I want to keep these small

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/enums/query/props/prop_enum.py
+++ b/lib/enums/query/props/prop_enum.py
@@ -1,6 +1,10 @@
+from __future__ import annotations
+
 from abc import abstractmethod
 from enum import Enum
-from typing import Callable, Any, Optional
+from typing import Callable, Any, Optional, Type
+
+from exceptions.parse_query_error import ParseQueryError
 
 PropFunc = Callable[[Any], Any]
 
@@ -9,6 +13,19 @@ class PropEnum(Enum):
     """
     An enum base class for all openstack resource properties - for type annotation purposes
     """
+
+    @staticmethod
+    def _from_string_impl(val: str, prop_cls: Type[PropEnum]):
+        """
+        Converts a given string in a case-insensitive way to the enum values
+        """
+        try:
+            return prop_cls[val.upper()]
+        except KeyError as err:
+            raise ParseQueryError(
+                f"Could not find property {val}. "
+                f"Available properties are {','.join([prop.name for prop in prop_cls])}"
+            ) from err
 
     @staticmethod
     @abstractmethod
--- a/lib/openstack_query/queries/server_query.py
+++ b/lib/openstack_query/queries/server_query.py
@@ -33,10 +33,16 @@ class ServerQuery(QueryWrapper):
 
     @property
     def prop_mapping(self) -> Type[ServerProperties]:
+        """
+        Returns the enum this query class uses to map property names to openstack server properties
+        """
         return ServerProperties
 
     @property
     def query_runner(self) -> ServerRunner:
+        """
+        Instantiates and returns a ServerRunner object with the correct property mapping pre-defined
+        """
         return ServerRunner(self.prop_mapping)
 
     def _get_server_side_handler(self) -> ServerSideHandler:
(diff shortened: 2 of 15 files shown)


lib/openstack_query/queries/server_query.py
```diff
@@ -22,17 +24,20 @@
...
 
-    prop_enum_cls = ServerProperties
-    runner_cls = ServerRunner
+    @property
+    def prop_mapping(self) -> Type[ServerProperties]:
+        return ServerProperties
+
+    @property
+    def query_runner(self) -> ServerRunner:
+        return ServerRunner(self.prop_mapping)
```
meoflynn, line 40: Could we add docstrings for these two methods?

lib/openstack_query/runners/user_runner.py
```diff
@@ -25,14 +24,6 @@ class UserRunner(QueryRunner):
...
     DEFAULT_DOMAIN = UserDomains.STFC
 
-    @staticmethod
-    def _get_marker(obj: User) -> str:
-        """
-        This method returns the marker value for a User object, required for pagination
-        :param obj: An Server object to get the marker property for
-        """
-        return UserProperties.get_prop_mapping("id")(obj)
-
```
meoflynn, line 35: Why has this method been removed?

meoflynn (changes requested):

lib/openstack_query/query_base.py
```diff
@@ -1,9 +1,12 @@
 from abc import ABC, abstractmethod
-from structs.query.query_client_side_handlers import QueryClientSideHandlers
+from typing import Type
+from typing import TypeVar
```
anish-mudaraddi, line 3:
```suggestion
from typing import Type, TypeVar
```

lib/openstack_query/runners/query_runner.py
```diff
@@ -155,7 +154,7 @@ def _run_paginated_query(
                 if i == self._LIMIT_FOR_PAGINATION - 1:
                     # restart the for loop with marker set
                     paginated_filters.update(
-                        {"marker": self._page_marker_prop_func(resource)}
+                        {"marker": self._prop_enum.get_prop_mapping(resource)}
```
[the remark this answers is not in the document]
anish-mudaraddi: Plesae change the tests accordingly

lib/openstack_query/runners/user_runner.py
anish-mudaraddi: Looks to be a redundant function that I should have deleted

anish-mudaraddi (changes requested):
Looks good - except for the issue with how pagination marker function is set. This should really have some acceptance testing done in a dev environment

lib/enums/query/props/prop_enum.py
```diff
@@ -1,6 +1,10 @@
+from __future__ import annotations
```
anish-mudaraddi, line 1: Are you sure this works on Python 3.8? 
I believe I opted not to use this because of version dependencies when using python 3.6

DavidFair: We've dropped 3.6 with the upgrade, and I've just checked 3.7+ https://docs.python.org/3/library/__future__.html

lib/openstack_query/runners/query_runner.py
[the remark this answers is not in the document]
DavidFair: Attempting to do this change results in the following:

```
                if i == self._LIMIT_FOR_PAGINATION - 1:
                    # restart the for loop with marker set
                    paginated_filters.update(
>                       {"marker": self._prop_enum.get_prop_mapping()(resource)}
                    )
E                   TypeError: QueryRunnerTests.setUp.<locals>.<lambda>() missing 1 required positional argument: 'mock_obj'

lib/openstack_query/runners/query_runner.py:157: TypeError
```

Since it looks like we need an object here, as we're now calling the lambda instead simply returning it?

DavidFair:
Closing as stale

PR #125, opened 2023-09-22
BUG: Not using CloudDomain to check validity of cloud_account in run() method

anish-mudaraddi:
### Description:

Recent changes made to run() method signiture - where `cloud_account` was changed from `CloudDomain` enum to a string removed use of `CloudDomain` entirely - this was necessary to check the validity of the `cloud_account` being passed in - and convert it to a value that would be accepted by Openstacksdk

This fix adds that back in 

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/tests/lib/openstack_query/managers/test_query_manager.py
+++ b/tests/lib/openstack_query/managers/test_query_manager.py
@@ -41,8 +41,11 @@ class QueryManagerTests(unittest.TestCase):
 
         self.query = MagicMock()
         self.prop_cls = MagicMock()
+        self.cloud_account = MagicMock()
+        self.cloud_account.name = "test_account"
+
         self.instance = QueryManager(
-            cloud_account="test_account", query=self.query, prop_cls=self.prop_cls
+            cloud_account=self.cloud_account, query=self.query, prop_cls=self.prop_cls
         )
 
     def _run_build_and_run_query_case(
--- a/tests/lib/openstack_query/test_query_methods.py
+++ b/tests/lib/openstack_query/test_query_methods.py
@@ -121,14 +121,17 @@ class QueryMethodsTests(unittest.TestCase):
         mock_client_filter_func = self.mock_builder.client_side_filter
         mock_server_filters = self.mock_builder.server_side_filters
         mock_parser_results = ["obj1", "obj2"]
+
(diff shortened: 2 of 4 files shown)


tests/lib/openstack_query/test_query_methods.py
```diff
@@ -145,12 +148,25 @@ def test_run_with_parsing_grouped(self):
...
+        # should implicitly convert it to a valid enum in the function
+        self.instance.run("DEV")
+
+        self.mock_runner.run.assert_called_once_with(
+            "dev", mock_client_filter_func, mock_server_filters, None
+        )
+
+        self.mock_parser.run_parser.assert_called_once_with(
+            self.mock_runner.run.return_value
+        )
```
DavidFair, line 168: I think we can drop this, since the prod case shows that it works whilst this exercises the same code path anyway?

tests/lib/openstack_query/managers/test_query_manager.py
```diff
@@ -41,8 +41,11 @@ def setUp(self) -> None:
 
         self.query = MagicMock()
         self.prop_cls = MagicMock()
+        self.cloud_account = MagicMock()
+        self.cloud_account.name = "test_account"
+
         self.instance = QueryManager(
-            cloud_account="test_account", query=self.query, prop_cls=self.prop_cls
+            cloud_account=self.cloud_account, query=self.query, prop_cls=self.prop_cls
```
DavidFair, line 48: We should just use the mock and not set a string, that way we can check it calls the right thing:

```python
self.run.assert_called_with(self.cloud_account.name.lower.return_value)
```
This makes it clear what we're doing to our enum, rather than have a string pop out

DavidFair (changes requested):

tests/lib/openstack_query/test_query_methods.py
anish-mudaraddi: No - this test checks that `run` can handle when the results returned by `run_parser` are grouped - i.e the method returns a `dict` rather than a list

DavidFair (approved):

DavidFair:
CI is failing due to coverage submission, I'm going to merge this anyway as I can see we haven't added any untested code

PR #127, opened 2023-09-22
ENH: clean workflow action outputs

anish-mudaraddi:
### Description:

We noticed that `create_project_internal` and `create_project_external` were both providing too much information that would lead to confusion - especially that it would output the admin `project_id` which could be confused with the actual `project_id`. 
(Almost deleted the admin project because of this :grimacing: ) - hence we removed this 

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/workflows/project.create.external.yaml
+++ b/actions/workflows/project.create.external.yaml
@@ -26,11 +26,16 @@ vars:
   - project_uuid: null
   - router_uuid: null
   - subnet_uuid: null
-  - stdout: null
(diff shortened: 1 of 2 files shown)


DavidFair (approved):

DavidFair (approved):
@anish-mudaraddi We'll need to re-run pack registration on the CLI, as I find that params are wishy-washy if they update

PR #128, opened 2023-09-26
ENH: Remove text for keys when building unique groups

meoflynn:
### Description:
Removing the text before the key value when we group query results so that we can then use the group keys in other operations

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:


anish-mudaraddi (dismissed):
LGTM

anish-mudaraddi:
ah yea you may want to update that unittest before merging :)

anish-mudaraddi (approved):

PR #129, opened 2023-09-26
MAINT: move run() method into separate class

anish-mudaraddi:
run method was getting too big - moved it to separate class - QueryExecute

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:


lib/openstack_query/queries/query_wrapper.py
```diff
@@ -17,7 +18,10 @@ class QueryWrapper(QueryMethods, QueryBase):
 
     def __init__(self):
         self._query_results = []
-        self.runner = self.runner_cls(self.prop_enum_cls.get_marker_prop_func())
+        self.executer = QueryExecuter(
+            self.prop_enum_cls,
+            self.runner_cls(self.prop_enum_cls.get_marker_prop_func()),
```
DavidFair, line 23: Since we already have to prop enum cls we should just do this in the child class

lib/openstack_query/query_executer.py
```diff
@@ -0,0 +1,113 @@
...
+
+class QueryExecuter:
+    """
+    Helper class to handle executing the query - primarily performing 'run()' method
+    """
+
+    def __init__(self, prop_enum_cls: Type[PropEnum], runner: QueryRunner):
+        self._prop_enum_cls = prop_enum_cls
+        self.runner = runner
+        self._parser_func = None
```
DavidFair, line 26: ... I think we can simply do `prop_enum_cls.get_marker_prop_func()` here instead?

lib/openstack_query/query_executer.py
```diff
@@ -0,0 +1,113 @@
...
+        client_side_filter_func: Optional[ClientSideFilterFunc] = None,
+        server_side_filters: Optional[ServerSideFilters] = None,
+    ):
+        """
+        Setter method for setting run filters
+        :param client_side_filter_func: An Optional function used to limit the results after querying openstacksdk
+        :param server_side_filters: An Optional set of filter kwargs to pass to openstacksdk
+        """
+        self._client_side_filter_func = client_side_filter_func
+        self._server_side_filters = server_side_filters
```
DavidFair, line 42: Use properties and directly set these instead:

```
@property
def client_side_filter_func(self):
   return self._client_side_filter_func()

@property.setter
def client_side_filter_func(self, val):
   self._client_side_filter_func = val
```

Then later on we can easily add extra validation, but we can use these as properties such as

```
executor.server_side_filters = [filter1, filter2]
executor.client_side_filters = [client_side_filter]
```

This also make the unit testing for the callers easier, since we can simply

```
assert self.mocked_executor.client_side_filters == [expected]
```

lib/openstack_query/query_executer.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def set_parse_func(
+        self,
+        parser_func: Optional[
+            Callable[[List[OpenstackResourceObj]], Union[Dict, List]]
```
DavidFair, line 47: I still really don't like this, the type basically screams to me we have a class or method here that we should be using rather than passing lambda / methods around

It's not for this PR, but we're adding more references to this so it's going to get harder to refactor :thinking:

lib/openstack_query/query_executer.py
DavidFair: Ditto on all the setters and getters below

lib/openstack_query/query_executer.py
```diff
@@ -0,0 +1,113 @@
...
+        :param from_subset: A subset of openstack resources to run query on instead of querying openstacksdk
+        :param kwargs: An extra set of kwargs to pass to internal _run_query method that changes what/how the
+        openstacksdk query is run
+            - valid kwargs to _run_query is specific to the runner object - see docstrings for _run_query() on the
+            runner of interest.
+        """
+
+        start = time.time()
+        results = self.runner.run(
+            cloud_account=CloudDomains.from_string(cloud_account).name.lower(),
```
DavidFair, line 102: This should just take the string as-is

Otherwise someone with `Openstack` as their cloud running locally won't be able to use that, as it's not in the cloud domains list

lib/openstack_query/query_methods.py
```diff
@@ -135,60 +133,25 @@ def run(
     ):
         """
         Public method that runs the query provided and outputs
-        :param cloud_account: An string for the account from the clouds configuration to use
+        :param cloud_account: An Enum for the account from the clouds configuration to use
```
DavidFair, line 136: Ditto about using the string to not limit people running locally to `Prod` / `Dev` since this is a St2 specific thing

tests/lib/openstack_query/test_query_executer.py
```diff
@@ -0,0 +1,141 @@
+from unittest.mock import MagicMock, patch
+import pytest
+from openstack_query.query_executer import QueryExecuter
+from tests.lib.openstack_query.mocks.mocked_props import MockProperties
+
+# pylint:disable=protected-access
```
DavidFair, line 6: Can we drop this? I'm really not keen on protected-access as it's usually a sign the class is doing too much, since we need to observer intermediate state

DavidFair (changes requested):

lib/openstack_query/query_executer.py
anish-mudaraddi: then there's no point in CloudDomains Enum at all

anish-mudaraddi: Although perhaps we shouldn't allow aliases here - as a security measure - the user must know the cloud domain in order to use it?

tests/lib/openstack_query/test_query_executer.py
anish-mudaraddi: How would you test whether a setter works then?

DavidFair: Just exercise both at the same time, if you set a value do you get it back
The setter getter should be doing validation at most (e.g. `isinstance(list)`), not manipulating the value

If we need to manipulate the value in a complex way then it's maybe a sign the value should be part of a class

lib/openstack_query/query_executer.py
```diff
@@ -97,9 +111,12 @@ def run_query(
             runner of interest.
         """
 
+        if isinstance(cloud_account, str):
+            cloud_account = CloudDomains.from_string(cloud_account)
```
DavidFair, line 115: This still would force us to use either `PROD` or `DEV`?

lib/openstack_query/query_methods.py
```diff
@@ -127,25 +128,23 @@ def group_by(
 
     def run(
         self,
-        cloud_account: str,
+        cloud_account: Union[str, CloudDomains],
```
DavidFair, line 131: If we've changed this type to be an enum either, I can't see a name.lower()?

tests/lib/openstack_query/test_query_executer.py
```diff
@@ -1,51 +1,93 @@
...
+    assert instance.client_side_filter_func == mock_client_filter
+
+
+def test_set_server_side_filters(instance):
+    """
+    Tests that server_side_filters setter method works as expected
+    """
+    mock_client_filter = MagicMock()
+    instance.server_side_filters = mock_client_filter
+    assert instance._server_side_filters == mock_client_filter
```
DavidFair, line 45: Ditto on my previous comment, just test a property instead of a getter / setter:

```
def test_client_side_filter_prop(instance):
    mock_client_filter = MagicMock()
    instance.client_side_filter_func = mock_client_filter
    assert instance.client_side_filter_func == mock_client_filter
```

Then we can add validation later, e.g.

```
def test_server_side_filter_func_takes_iterable():
    mocked_server_filter = NonCallableMock()
    # Intentionally don't pack this into a list
    with self.assertRaises(...):
         self.instance.server_filter_funcs = mocked_server_filter
```

lib/openstack_query/query_executer.py
anish-mudaraddi: I'm not comfortable removing it - I'll leave it for another PR

lib/openstack_query/query_methods.py
anish-mudaraddi: we do the conversion in `run_query` - because we want to limit what we do in QueryMethods

lib/openstack_query/query_executer.py
```diff
@@ -126,5 +126,6 @@ def run_query(
         logger.debug("run completed - time elapsed: %s seconds", time.time() - start)
 
         if self._parse_func:
-            results = self.parse_func(results)
+            parse_func = self._parse_func
+            results = parse_func(results)
```
DavidFair, line 130: Nit, we can just `results = self._parse_func(results)`

anish-mudaraddi: I thought that would fix it but apparently not

DavidFair: Hmm, could we revert and possible suppress?
Usually though the warnings are pretty good and point at something real, if it's mis-typing or just param mis-matches

anish-mudaraddi: I think its complaining because it thinks we're calling a method - but we're instead taking the output of the property and calling that since the output is a function

DavidFair: Yeah looking at the warning:
`self.parse_func is not callable`

The typing just needs to be included

```
self.parse_func: Optional[Callable] = None
```

DavidFair (approved):

lib/openstack_query/query_executer.py
anish-mudaraddi: yea I just used the internal attribute instead

PR #132, opened 2023-10-02
Openstack query/type fixes

anish-mudaraddi:
### Description:
very simple type fixes

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/openstack_query/managers/user_manager.py
+++ b/lib/openstack_query/managers/user_manager.py
@@ -28,7 +28,7 @@ class UserManager(ManagerWrapper):
             prop_cls=UserProperties,
         )
 
-    def search_by_datetime(self, **_) -> QueryReturn:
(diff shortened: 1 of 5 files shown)


lib/openstack_query/managers/user_manager.py
```diff
@@ -28,7 +27,7 @@ def __init__(self, cloud_account: CloudDomains):
             prop_cls=UserProperties,
         )
 
-    def search_by_datetime(self, **_) -> QueryReturn:
+    def search_by_datetime(self, **_) -> None:
```
DavidFair, line 30: Is this returning anything?

DavidFair: This simply raises an error now

DavidFair (approved):

PR #133, opened 2023-10-02
Openstack query/re write query block tests

anish-mudaraddi:
Rewrite "Query Block" tests to use fixtures

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:


anish-mudaraddi:
Too big - splitting

PR #136, opened 2023-10-02
TST: Rewrite QueryParser Tests

anish-mudaraddi:
rewrite tests to use pytest fixtures and conventions - i.e. use functions instead of test classes

### DIFF
--- a/lib/openstack_query/query_blocks/query_parser.py
+++ b/lib/openstack_query/query_blocks/query_parser.py
@@ -1,8 +1,13 @@
-from typing import List, Dict, Tuple, Union, Optional, Callable
+from typing import List, Dict, Tuple, Union, Optional, Type
 import logging
 from collections import OrderedDict
 from enums.query.props.prop_enum import PropEnum
-from custom_types.openstack_query.aliases import OpenstackResourceObj, PropValue
+from custom_types.openstack_query.aliases import (
+    OpenstackResourceObj,
+    PropValue,
+    GroupMappings,
+    GroupRanges,
+)
 from exceptions.parse_query_error import ParseQueryError
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


lib/openstack_query/query_blocks/query_parser.py
```diff
@@ -14,23 +19,61 @@ class QueryParser:
...
+    @property
+    def group_mappings(self):
+        """
+        a getter method to set group_mappings options
+        """
+        return self._group_mappings
+
+    @group_mappings.setter
+    def group_mappings(self, group_mappings: GroupMappings):
+        self._group_mappings = group_mappings
```
meoflynn, line 68: Could we add a docstring to this method?

tests/lib/openstack_query/query_blocks/test_query_parser.py
```diff
@@ -1,468 +1,466 @@
...
+        return MagicMock(wraps=lambda obj, prop=mock_prop: obj[prop.name.lower()])
+
+    return _mock_get_prop_mapping
+
+
+@pytest.fixture(name="run_sort_by_tests")
+def run_sort_by_tests_fixture(instance, mock_get_prop_mapping):
+    def _run_sort_by_tests(obj_list, sort_by_specs, expected_list):
+        instance.sort_by = sort_by_specs
         with patch.object(
```
meoflynn, line 36: Could we add a docstring to this method?

Why is the method within the `run_sort_by_tests_fixture` method?

tests/lib/openstack_query/query_blocks/test_query_parser.py
```diff
@@ -1,468 +1,466 @@
...
+        instance._parse_group_ranges(group_ranges=mock_group_ranges)
+        mock_get_prop_func.assert_called_once_with(mock_group_by)
+
+    assert instance.group_mappings
+
+    assert instance.group_mappings["group1"]({"prop_1": "val1"})
+    assert not instance.group_mappings["group2"]({"prop_1": "val1"})
+
+
+def test_add_include_missing_group(instance, mock_get_prop_mapping):
```
meoflynn, line 160: Could we add a docstring to this unit test?

meoflynn (changes requested):

tests/lib/openstack_query/query_blocks/test_query_parser.py
anish-mudaraddi: This fixture is basically setup to return some shared boilerplate code to run sort by test cases - so I return a function which each test can run internally. 

I think its cleaner since if we want any setup code in the fixture (like here this fixture is using another `instance`) fixture - this can be separated from the actual test cases. In other words - the stuff in the internal function is stuff concerned with running the test case, the stuff outside the internal function is around setting up tests

meoflynn (approved):

PR #141, opened 2023-10-02
TST: Refactor query_runner tests to pytest

anish-mudaraddi:
changed QueryRunner tests so that they are in line with pytest conventions
Included shared fixtures in conftest.py

### DIFF
--- a/tests/lib/openstack_query/runners/conftest.py
+++ b/tests/lib/openstack_query/runners/conftest.py
@@ -1,0 +1,18 @@
+from unittest.mock import MagicMock
+import pytest
+
+
+@pytest.fixture(scope="function", name="mock_connection")
+def mock_connection_fixture():
+    """
+    Returns a mocked OpenstackConnection class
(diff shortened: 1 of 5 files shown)


tests/lib/openstack_query/runners/conftest.py
```diff
@@ -0,0 +1,18 @@
...
+
+@pytest.fixture(scope="module", name="mock_connection")
+def mock_connection_fixture():
+    """
+    Returns a mocked OpenstackConnection class
+    """
+    return MagicMock()
+
+
+@pytest.fixture(scope="module", name="mock_openstack_connection")
```
DavidFair, line 13: These should be `function` scope at most, else the same mocks will pass later tests for an `assert_called`

DavidFair (changes requested):

DavidFair (approved):

DavidFair (approved):

PR #144, opened 2023-10-03
TST: Refactor struct module to use pytest

anish-mudaraddi:
refactor struct module tests to use pytest fixtures and conventions

### DIFF
--- a/tests/structs/test_smtp_account.py
+++ b/tests/structs/test_smtp_account.py
@@ -1,84 +1,89 @@
-import unittest
-
+from typing import List, Dict
 import pytest
-
 from structs.email.smtp_account import SMTPAccount
 
 
-class TestEmailParams(unittest.TestCase):
+@pytest.fixture(name="mock_smtp_accounts")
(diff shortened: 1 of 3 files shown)


meoflynn (dismissed):

tests/structs/test_smtp_account.py
```diff
@@ -1,84 +1,85 @@
...
-    def test_from_dict(self):
-        """
-        Tests that from_dict() static method works properly
-        this method should build a SMTPAccount dataclass from a valid dictionary
-        """
-
-        mock_valid_kwargs = {
+@pytest.fixture(name="mock_smtp_accounts")
+def mock_smtp_accounts_fixture():
+    return [
```
DavidFair, line 7: Docstring - this should mention that it includes multiple smtp accounts as a dict

We should also add a type annotation here too as we have a data struct with a similar name

DavidFair (changes requested):

DavidFair (approved):

PR #146, opened 2023-10-03
MAINT: Refactor to use factory pattern

anish-mudaraddi:
change to now use a factory pattern to create queries 
decouple query mappings and query components from queries

This will make chaining from one query to another easier


anish-mudaraddi:
Needs tests for RunnerWrapper

lib/openstack_query/handlers/server_side_handler.py
```diff
@@ -95,12 +101,12 @@ def _check_filter_mapping(
...
-                "\n\t".join(
-                    [f"{key}: '{value}'" for key, value in filter_params.items()]
-                ),
-            )
+            # logger.debug(
+            #     "checking server-side filter function against provided parameters\n\t%s",
+            #     "\n\t".join(
+            #         [f"{key}: '{value}'" for key, value in filter_params.items()]
+            #     ),
+            # )
```
DavidFair, line 109: Just so we don't forget this, we should add this back or resolve the issues we were seeing with None types

lib/openstack_query/api/query_objects.py
```diff
@@ -0,0 +1,28 @@
+from typing import Type
+from openstack_query.api.query_api import QueryAPI
+from openstack_query.mappings.mapping_interface import MappingInterface
+from openstack_query.query_factory import QueryFactory
+from openstack_query.mappings.user_mapping import UserMapping
+from openstack_query.mappings.server_mapping import ServerMapping
+
+
+def get_common(query_mapping: Type[MappingInterface]) -> QueryAPI:
+    return QueryAPI(QueryFactory.build_query_deps(query_mapping))
```
DavidFair, line 10: Missing docstring

lib/openstack_query/handlers/server_side_handler.py
```diff
@@ -26,6 +26,12 @@ class ServerSideHandler(HandlerBase):
     def __init__(self, kwarg_mappings: ServerSideFilterMappings):
         self._server_side_filter_mappings = kwarg_mappings
 
+    def get_supported_props(self, preset: QueryPresetsGeneric) -> List[PropEnum]:
+        """
+        Gets a list of all supported properties for a given preset
+        """
+        return list(self._server_side_filter_mappings[preset].keys())
```
DavidFair, line 33: I didn't manage to write a test for this, we should add one

In this case I think a really simple stub would work to just check we get the top level keys from a nested dict of NonCallableMocks

DavidFair (changes requested):

lib/openstack_query/query_factory.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class QueryFactory:
+    """
+    This class is used to construct a query object using a factory pattern.
+    Decouples the API from the implementation
+    """
+
+    @staticmethod
+    def build_query_deps(mapping_cls: Type[MappingInterface]) -> QueryComponents:
```
DavidFair, line 20: We can probably just move this to a module method, rather than having a class for it

I think it's a hang-over from when I was trying to do it with metaclasses

tests/lib/openstack_query/mappings/test_server_mapping.py
```diff
@@ -0,0 +1,50 @@
...
+        ServerProperties.PROJECT_ID: "project_id",
+    }
+
+    supported_props = handler.get_supported_props(QueryPresetsGeneric.EQUAL_TO)
+    assert all(key_to_check in supported_props for key_to_check in mappings)
+    for prop, expected in mappings.items():
+        server_filter = handler.get_filters(
+            QueryPresetsGeneric.EQUAL_TO, prop, {"value": "test"}
+        )
+        assert server_filter == {expected: "test"}
```
DavidFair, line 50: We don't have any test for client side filters similarly?

tests/lib/openstack_query/mappings/test_user_mapping.py
```diff
@@ -0,0 +1,47 @@
...
+        UserProperties.USER_ID: "id",
+    }
+
+    supported_props = handler.get_supported_props(QueryPresetsGeneric.EQUAL_TO)
+    assert all(key_to_check in supported_props for key_to_check in mappings)
+    for prop, expected in mappings.items():
+        user_filter = handler.get_filters(
+            QueryPresetsGeneric.EQUAL_TO, prop, {"value": "test"}
+        )
+        assert user_filter == {expected: "test"}
```
DavidFair, line 47: Since this mechanism is the same, lets create a helper file with this as a method where we take an expected dict and the instance under test

Since we'll have to duplicate this for every mapping file

DavidFair (changes requested):

lib/openstack_query/handlers/server_side_handler.py
anish-mudaraddi: there is a test for this

lib/openstack_query/query_factory.py
anish-mudaraddi: I may need this class to do chaining between queries

anish-mudaraddi:
I've added client-side mapping tests - and altered server-side tests to ensure that there is an equivalent client-side test (to allow queries using `from_subset`)

lib/structs/query/query_components.py
```diff
@@ -0,0 +1,14 @@
+from dataclasses import dataclass
+
+from openstack_query.query_blocks.query_builder import QueryBuilder
+from openstack_query.query_blocks.query_output import QueryOutput
+from openstack_query.query_blocks.query_parser import QueryParser
+from openstack_query.query_blocks.query_executer import QueryExecuter
+
+
+@dataclass
+class QueryComponents:
```
DavidFair, line 10: Missing doc string on this one

DavidFair (changes requested):

DavidFair (approved):

PR #148, opened 2023-10-03
ENH:  Allow multiple queries on same resource

anish-mudaraddi:
refactored builder and runner methods to allow multiple where() methods to be run. This allows you to define multiple presets to filter by.

This change also fixes ANY_IN - now using ANY_IN will run multiple server-side EQUAL_TO queries and aggregate the results - acting essentially as OR

server side and client side fiilters now accept a list instead of just one

### DIFF
--- a/lib/enums/query/query_presets.py
+++ b/lib/enums/query/query_presets.py
@@ -13,6 +13,8 @@ class QueryPresetsGeneric(QueryPresets):
 
     EQUAL_TO = auto()
     NOT_EQUAL_TO = auto()
+    ANY_IN = auto()
+    NOT_ANY_IN = auto()
 
     @staticmethod
     def from_string(val: str):
@@ -81,9 +83,7 @@ class QueryPresetsString(QueryPresets):
     Enum class which holds string comparison operators
     """
 
-    ANY_IN = auto()
     MATCHES_REGEX = auto()
-    NOT_ANY_IN = auto()
 
     @staticmethod
     def from_string(val: str):
--- a/lib/openstack_query/handlers/client_side_handler.py
+++ b/lib/openstack_query/handlers/client_side_handler.py
@@ -158,69 +158,27 @@ class ClientSideHandler(HandlerBase):
         func: FilterFunc, func_kwargs: Optional[FilterParams] = None
     ) -> Tuple[bool, str]:
         """
-        Method that checks a given function can accept a set of kwargs as arguments.
+        Method checks a given function can accept a set of kwargs as arguments - using the EAFP principle.
+        returns false and the error string if an error is raised - else true
         :param func: function to test
         :param func_kwargs: kwargs to test
         """
-
         signature = inspect.signature(func)
+        prop_param = list(signature.parameters.values())[0]
 
-        # skip first parameter for the filter function as this is always the a positional arg which takes
-        # an Openstack resource property as the input for our filter function to compare against
-        parameters = list(signature.parameters.values())[1:]
+        # a hack to get EAFP working - set a default value for prop just to see if filter function works
+        prop_val = {
+            typing.Any: "",
+            typing.Union[int, float]: 0,
+        }.get(prop_param.annotation, prop_param.annotation())
 
-        has_varargs = any(param.kind == param.VAR_POSITIONAL for param in parameters)
-        has_varkwargs = any(param.kind == param.VAR_KEYWORD for param in parameters)
+        if not func_kwargs:
+            func_kwargs = {}
+        try:
+            func(prop=prop_val, **func_kwargs)
+            return True, ""
 
-        logger.debug(
-            "checking client-side filter function against provided parameters\n\t%s",
-            "\n\t".join([f"{key}: '{value}'" for key, value in func_kwargs.items()]),
-        )
-
-        all_params = [param.name for param in parameters]
-        log_all_params = "<none>"
-        if all_params:
-            log_all_params = "\n\t".join(all_params)
-        logger.debug(
-            "client-side filter function '%s' has the following params:\n\t%s",
-            func.__name__,
-            log_all_params,
-        )
-
-        req_params = [
-            param.name
-            for param in parameters
-            if param.VAR_POSITIONAL and param.default == inspect.Parameter.empty
-        ]
-        log_all_req_params = "<none>"
-        if req_params:
-            log_all_req_params = "\n\t".join(req_params)
-
-        logger.debug(
-            "of those, the following params are required:\n\t%s", log_all_req_params
-        )
-
-        for param in parameters:
-            if param.kind == param.POSITIONAL_OR_KEYWORD:
-                param_name = param.name
-                if param_name in func_kwargs:
-                    kwargs_value = func_kwargs[param_name]
-                    param_type = param.annotation
-                    if param_type not in [
-                        Any,
-                        inspect.Parameter.empty,
-                    ] and not isinstance(kwargs_value, param_type):
-                        return (
-                            False,
-                            f"{param_name} given has incorrect type, "
-                            f"expected {param_type}, found {type(kwargs_value)}",
-                        )
-                elif param.default is inspect.Parameter.empty:
-                    return False, f"{param_name} expected but not given"
-
-        if not has_varargs and not has_varkwargs:
-            # Check for extra items in kwargs not present in the function signature
-            unexpected_vals = set(func_kwargs.keys()) - set(p.name for p in parameters)
-            if unexpected_vals:
-                return False, f"unexpected arguments: '{unexpected_vals}'"
-        return True, ""
+        # we're catching all possible exceptions - and fail noisily
+        # pylint:disable=broad-exception-caught
+        except Exception as exp:
+            return False, str(exp)
--- a/lib/openstack_query/handlers/client_side_handler.py
+++ b/lib/openstack_query/handlers/client_side_handler.py
@@ -168,8 +168,8 @@ class ClientSideHandler(HandlerBase):
 
         # a hack to get EAFP working - set a default value for prop just to see if filter function works
         prop_val = {
-            typing.Any: "",
-            typing.Union[int, float]: 0,
+            Any: "",
+            Union[int, float]: 0,
         }.get(prop_param.annotation, prop_param.annotation())
 
         if not func_kwargs:
--- a/lib/openstack_query/handlers/client_side_handler.py
+++ b/lib/openstack_query/handlers/client_side_handler.py
@@ -1,6 +1,7 @@
 import logging
 import inspect
-from typing import Optional, Tuple, Any, List, Union
+import typing
+from typing import Optional, Tuple, List, Union
 
 from openstack_query.handlers.handler_base import HandlerBase
 from custom_types.openstack_query.aliases import (
@@ -167,10 +168,12 @@ class ClientSideHandler(HandlerBase):
         prop_param = list(signature.parameters.values())[0]
 
         # a hack to get EAFP working - set a default value for prop just to see if filter function works
-        prop_val = {
-            Any: "",
-            Union[int, float]: 0,
-        }.get(prop_param.annotation, prop_param.annotation())
+        if prop_param.annotation == typing.Any:
+            prop_val = ""
+        elif prop_param.annotation == typing.Union[int, float]:
+            prop_val = 0
+        else:
+            prop_val = prop_param.annotation()
 
         if not func_kwargs:
             func_kwargs = {}
--- a/lib/openstack_query/handlers/client_side_handler.py
+++ b/lib/openstack_query/handlers/client_side_handler.py
@@ -3,6 +3,10 @@ import inspect
 import typing
 from typing import Optional, Tuple, List, Union
 
+from exceptions.missing_mandatory_param_error import MissingMandatoryParamError
+from exceptions.parse_query_error import ParseQueryError
+from exceptions.query_preset_mapping_error import QueryPresetMappingError
+
 from openstack_query.handlers.handler_base import HandlerBase
 from custom_types.openstack_query.aliases import (
     FilterFunc,
@@ -15,7 +19,6 @@ from custom_types.openstack_query.aliases import (
 
 from enums.query.query_presets import QueryPresets
 from enums.query.props.prop_enum import PropEnum
-from exceptions.query_preset_mapping_error import QueryPresetMappingError
 
 logger = logging.getLogger(__name__)
 
@@ -181,7 +184,10 @@ class ClientSideHandler(HandlerBase):
             func(prop=prop_val, **func_kwargs)
             return True, ""
 
-        # we're catching all possible exceptions - and fail noisily
-        # pylint:disable=broad-exception-caught
-        except Exception as exp:
+        except (
+            TypeError,
+            NameError,
+            ParseQueryError,
+            MissingMandatoryParamError,
+        ) as exp:
             return False, str(exp)
--- a/lib/openstack_query/query_blocks/query_builder.py
+++ b/lib/openstack_query/query_blocks/query_builder.py
@@ -7,11 +7,14 @@ from openstack_query.handlers.server_side_handler import ServerSideHandler
 from enums.query.query_presets import QueryPresets
 from enums.query.props.prop_enum import PropEnum
 
-from exceptions.parse_query_error import ParseQueryError
 from exceptions.query_preset_mapping_error import QueryPresetMappingError
 from exceptions.query_property_mapping_error import QueryPropertyMappingError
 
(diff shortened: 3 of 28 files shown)


anish-mudaraddi:
tests for RunnerWrapper.run() need to be re-written

anish-mudaraddi:
waiting on #146

lib/enums/query/query_presets.py
```diff
@@ -81,9 +83,7 @@ class QueryPresetsString(QueryPresets):
     Enum class which holds string comparison operators
     """
 
-    ANY_IN = auto()
     MATCHES_REGEX = auto()
-    NOT_ANY_IN = auto()
```
DavidFair, line 86: Why have we dropped this?
Whilst you can use regex, doing multiple groups with regex is painful such as `ANY_IN ("foo", "bar")` is easier to read / parse?

DavidFair:
@anish-mudaraddi could you rebase this on main?

anish-mudaraddi:
@DavidFair it already is - we can merge it fine

lib/openstack_query/handlers/client_side_handler.py
```diff
@@ -158,69 +159,29 @@ def _check_filter_func(
...
         signature = inspect.signature(func)
+        prop_param = list(signature.parameters.values())[0]
+
+        # a hack to get EAFP working - set a default value for prop just to see if filter function works
+        if prop_param.annotation == typing.Any:
+            prop_val = ""
+        elif prop_param.annotation == typing.Union[int, float]:
+            prop_val = 0
+        else:
+            prop_val = prop_param.annotation()
```
DavidFair, line 176: We shouldn't be using the typing info at runtime, this feels like prop_val (and associated) should be encapsulated in a dedicated class which then uses properties to validate on the setter

lib/openstack_query/handlers/client_side_handler.py
```diff
@@ -158,69 +159,29 @@ def _check_filter_func(
...
-        if not has_varargs and not has_varkwargs:
-            # Check for extra items in kwargs not present in the function signature
-            unexpected_vals = set(func_kwargs.keys()) - set(p.name for p in parameters)
-            if unexpected_vals:
-                return False, f"unexpected arguments: '{unexpected_vals}'"
-        return True, ""
+        # we're catching all possible exceptions - and fail noisily
+        # pylint:disable=broad-exception-caught
+        except Exception as exp:
+            return False, str(exp)
```
DavidFair, line 187: We should only catch the expected ones ....e.g. type error
Currently this will also catch things like OutOfMemoryError, or KeyboardInterrupt for example

lib/openstack_query/handlers/client_side_handler.py
```diff
@@ -158,69 +159,29 @@ def _check_filter_func(
...
-                        )
-                elif param.default is inspect.Parameter.empty:
-                    return False, f"{param_name} expected but not given"
-
-        if not has_varargs and not has_varkwargs:
-            # Check for extra items in kwargs not present in the function signature
-            unexpected_vals = set(func_kwargs.keys()) - set(p.name for p in parameters)
-            if unexpected_vals:
-                return False, f"unexpected arguments: '{unexpected_vals}'"
-        return True, ""
```
DavidFair, line 226: Nice tidy-up and simplification

lib/openstack_query/query_blocks/query_builder.py
```diff
@@ -66,6 +70,23 @@ def server_side_filters(self, server_filters: ServerSideFilters):
...
+    @server_filter_fallback.setter
+    def server_filter_fallback(self, fallback_filters: ClientSideFilters):
+        """
+        a setter method to set client-side filters to fallback on for each server-side filter if
+        server-side filters are not to be used
+        :param fallback_filters: a set of client-side filters
+        """
+        self._server_filter_fallback = fallback_filters
+
     def parse_where(
```
DavidFair, line 90: Can we drop this and simply append server-side filters to the client-side list on subset?
That way we still only have two modes (client / server) for filtering, since the effect rather than why we have it is the important part?

lib/openstack_query/query_blocks/query_builder.py
```diff
@@ -104,17 +119,20 @@ def parse_where(
...
             filter_func_kwargs=preset_kwargs,
         )
 
-        self.server_side_filters = self._server_side_handler.get_filters(
+        server_side_filters = self._server_side_handler.get_filters(
             preset=preset, prop=prop, params=preset_kwargs
         )
-        if not self.server_side_filters:
+        if server_side_filters and not isinstance(server_side_filters, list):
+            server_side_filters = [server_side_filters]
```
DavidFair, line 133: Do this in the setter, user's of the class should get something sane back according to the type we promised

lib/openstack_query/query_blocks/query_builder.py
```diff
@@ -123,14 +141,70 @@ def parse_where(
...
+        )
+
+        self._add_filter(
+            client_side_filter=client_side_filter,
+            server_side_filters=server_side_filters,
+        )
+
+    def _add_filter(
+        self,
+        client_side_filter: ClientSideFilterFunc,
```
DavidFair, line 166: What if there's no ClientSideFilters? Should this be optional too?

lib/openstack_query/query_blocks/query_builder.py
```diff
@@ -123,14 +141,70 @@ def parse_where(
...
+        """
+
+        # add as client_side_filter if no server_side_filter
+        if not server_side_filters:
+            self.client_side_filters.append(client_side_filter)
+            return
+
+        # we convert to singleton list for aggregating into server_side_filter
+        if not isinstance(server_side_filters, list):
+            server_side_filters = [server_side_filters]
```
DavidFair, line 183: Ditto about the property

lib/openstack_query/query_blocks/query_builder.py
```diff
@@ -123,14 +141,70 @@ def parse_where(
...
+
+        # check each any new server filter being added has overlapping keys with any server filters currently added
+        # if so add as client_side_filter
+        for current_server_filter in self.server_side_filters:
+            for new_server_filter in server_side_filters:
+                if set(new_server_filter.keys()) & set(current_server_filter.keys()):
```
DavidFair, line 189: Nit: use `and` instead of `&` 

Also your comments are straying into what it's doing rather than why, why do we append a client side filter if the new and old server side filter keys match?

lib/openstack_query/query_blocks/query_builder.py
```diff
@@ -123,14 +141,70 @@ def parse_where(
...
+                    self.client_side_filters.append(client_side_filter)
+                    return
+
+        # before adding server-side filter - set fallback
+        self.server_filter_fallback.append(client_side_filter)
+
+        # if server side filter not set - set it
+        if not self.server_side_filters:
+            self.server_side_filters = server_side_filters
+            return
```
DavidFair, line 199: Why? I can see if it's not set we set it, but not why

lib/openstack_query/runners/runner_wrapper.py
```diff
@@ -34,80 +34,112 @@ def __init__(self, marker_prop_func: PropFunc, connection_cls=OpenstackConnectio
...
+
+        if from_subset and server_side_filters:
+            logger.error(
+                "Error: Received server-side filters - suggesting that the query requires getting results via "
+                "openstacksdk, but also openstack objects directly - suggesting that we want to filter a set already "
+                "given. Not sure what to do - aborting."
+            )
+            raise RuntimeError(
+                "Ambiguous Query: received both server-side filters and values passed directly"
+            )
```
DavidFair, line 63: We should extract this to something like `self._validate_params` to prevent the run growing

lib/openstack_query/runners/runner_wrapper.py
```diff
@@ -34,80 +34,112 @@ def __init__(self, marker_prop_func: PropFunc, connection_cls=OpenstackConnectio
...
-                logger.debug("server-side query found %s items", len(resource_objects))
 
-        if client_side_filter_func and not server_side_filters:
+        if client_side_filters:
             logger.info("applying client side filters")
-            resource_objects = self._apply_client_side_filter(
-                resource_objects, client_side_filter_func
+            resource_objects = self._apply_client_side_filters(
+                items=resource_objects, filters=client_side_filters
             )
```
DavidFair, line 96: Is there any way we can make this more Pythonic, e.g.:
```
for filter in client_side_filters:
   result = self._apply_client_side_filter(filter)
```

That way we don't care if there's 0 or >=1 elements in the list

lib/openstack_query/runners/runner_wrapper.py
```diff
@@ -34,80 +34,112 @@ def __init__(self, marker_prop_func: PropFunc, connection_cls=OpenstackConnectio
...
+                "\n%s "
+                "\n\n meta kwargs: "
+                "\n\t%s",
+                filters_log_str,
+                "None" if not meta_param_log_str else meta_param_log_str,
+            )
+
+            resource_objects.extend(self._run_query(conn, query_filters, **kwargs))
+
         return resource_objects
```
DavidFair, line 143: IMO this should be in it's own class or the executor, as it bloats the runner with implementation details that can be abstracted / tested elsewhere

lib/openstack_query/runners/runner_wrapper.py
```diff
@@ -173,17 +205,21 @@ def _run_paginated_query(
...
-        :param filter_func: An Optional function that we can use to limit the results after querying openstacksdk,
-            - function takes an openstack resource object and returns True if it passes the filter, false if not
+        :param filters: filter functions that we can use to limit the results after querying openstacksdk,
+            - each function takes an openstack resource object and returns True if it passes the filter, false if not
         :return: List of items that match the1 given query
         """
-        return [item for item in items if filter_func(item)]
+        filtered_items = []
+        for item in items:
+            if all(filter_func(item) for filter_func in filters):
```
DavidFair, line 220: Hmm, I don't like this, it took me three attempts to realise it's asking us the caller to filter despite being a filter function

Can we not get filter_func to do it for us:

```
def _apply(input_items: List):
  for filter in client_side_filters:
       input_items = filter_func(input_items)

# Where filter_func is implemented as
def filter_func(...):
    return filter(input, lambda)
```

tests/lib/openstack_query/api/test_query_api.py
```diff
@@ -40,14 +43,19 @@ def _run_with_test_case(data_subset, mock_kwargs):
...
+        # - this differs based on if from_subset is given
+        if data_subset:
+            client_filters = (
+                instance.builder.client_side_filters
+                + instance.builder.server_filter_fallback
+            )
+            server_filters = None
+        else:
+            client_filters = instance.builder.client_side_filters
+            server_filters = instance.builder.server_side_filters
```
DavidFair, line 55: This if-else should be two different test cases
If this is a PITA to test where we don't want to write new cases then it's a bad code smell

tests/lib/openstack_query/mappings/conftest.py
DavidFair, line 1: Nit: the file should be called something like `mapping_test_helpers` or similar, since this doesn't directly test anything

tests/lib/openstack_query/query_blocks/test_query_builder.py
```diff
@@ -42,61 +41,95 @@ def instance_fixture(mock_client_side_handlers, mock_server_side_handler):
...
+            "openstack_query.query_blocks.query_builder.QueryBuilder._get_preset_handler"
+        ) as mock_get_preset_handler:
+            with patch(
+                "openstack_query.query_blocks.query_builder.QueryBuilder._add_filter"
+            ) as mock_add_filter:
+                mock_get_preset_handler.return_value = mock_client_side_handler
+                with patch.object(MockProperties, "get_prop_mapping") as mock_prop_func:
+                    instance.parse_where(
+                        MockQueryPresets.ITEM_1, MockProperties.PROP_1, mock_kwargs
+                    )
```
DavidFair, line 77: Why do we need to patch so much stuff for `get_filter_func`, this shouts to me the method is doing way too much

It should be fairly pure, e.g. `get_filter_func` gets pre-set filter functions, rather than generating them (which can be a separate step)

tests/lib/openstack_query/runners/test_runner_wrapper.py
```diff
@@ -75,146 +75,260 @@ def _run_paginated_query_test(number_iterations):
...
+        """
+        method to run run() test case with provided input values
+        The individual methods that are called in run must be patched out by the test function
+        prior to this and any asserts need to be done by the test function
+        """
 
-    res = instance.run(
-        cloud_account=mock_cloud_domain,
-        client_side_filter_func=mock_client_side_filter_func,
+        with patch(
```
DavidFair, line 96: I want to come in and refactor run in the future, so we'll get this in as-is and I'll throw together a prototype refactor :thinking:

DavidFair (changes requested):

lib/openstack_query/query_blocks/query_builder.py
anish-mudaraddi: we only register subsets at time of running - hence we need to hold a separate list here in query builder

lib/openstack_query/query_blocks/query_builder.py
anish-mudaraddi: nope - there's always a client-side filter. Otherwise we raise an error when we try and get the filter

lib/openstack_query/runners/runner_wrapper.py
anish-mudaraddi: for a future PR

lib/openstack_query/runners/runner_wrapper.py
anish-mudaraddi: future PR - I agree the wrapper is quite big - this is the case with a lot of components atm - I'll have a think about this

tests/lib/openstack_query/mappings/conftest.py
anish-mudaraddi: can't have fixtures in any other file but conftest.py i'm afraid

tests/lib/openstack_query/query_blocks/test_query_builder.py
anish-mudaraddi: I'll fix this later

tests/lib/openstack_query/runners/test_runner_wrapper.py
anish-mudaraddi: please do

lib/openstack_query/handlers/client_side_handler.py
anish-mudaraddi: I don't understand? I'll leave this for a future PR

lib/openstack_query/query_blocks/query_builder.py
anish-mudaraddi: & does something else here - i changed to intersection

anish-mudaraddi:
I made various code review changes - to address the server-side list typing issue - I enforce all server-side filters to be a list after getting it using the server-side-handler

lib/openstack_query/handlers/client_side_handler.py
DavidFair: Basically we shouldn't use type annotations as they're exactly that annotations

Really this should just be done in a dedicated class e.g.
```
class PropParam
   def __init__(prop_type: Type):
        self.prop_type: Type = prop_type
        self._held_val = None

   @property
   def val(self):
       return self._held_val

  @property.setter(val)
     def val(self, val):
          # cast to the set property type, this will fail with a type error if it's not correct
          self._held_val = self.prop_type(val)
```

Then all the checking is done for free, and Python gives a nice error such as `string "helllo" is not convertible to int`

DavidFair: One for the future

lib/openstack_query/handlers/client_side_handler.py
```diff
@@ -183,5 +186,10 @@ def _check_filter_func(
 
         # we're catching all possible exceptions - and fail noisily
         # pylint:disable=broad-exception-caught
```
DavidFair, line 188: This pylint disable can go away now

DavidFair (changes requested):

lib/openstack_query/handlers/client_side_handler.py
anish-mudaraddi: yea - I guess move ClientFIlter into a separate class (Same with ServerFilter) and handle checking and stuff there

DavidFair (approved):

meoflynn (approved):

PR #149, opened 2023-10-05
BUG: Make sure email addresses to email are provided as a list

meoflynn:
### Description:

If an email address has something like an apostrophe, it will split the email address into individual characters. Email params has been updated to take email addresses as a list of addresses

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/custom_types/email_api/aliases.py
+++ b/lib/custom_types/email_api/aliases.py
@@ -1,7 +1,7 @@
-from typing import Dict, Tuple
+from typing import Dict, List
 
 EmailAddress = str
-EmailAddresses = Tuple[EmailAddress]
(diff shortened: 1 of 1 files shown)


lib/structs/email/email_params.py
```diff
@@ -22,11 +22,11 @@ class EmailParams:
     """
 
     subject: str
-    email_from: EmailAddress
-    email_to: EmailAddresses
+    email_from: [EmailAddress]
```
DavidFair, line 25: This shouldn't have the `[]` chars

DavidFair (approved):

PR #150, opened 2023-10-09
ENH: add flavor query

anish-mudaraddi:
Draft for getting query by flavor functionality

### DIFF
--- a/lib/enums/query/props/flavor_properties.py
+++ b/lib/enums/query/props/flavor_properties.py
@@ -1,0 +1,68 @@
+from enum import auto
+from enums.query.props.prop_enum import PropEnum
+from exceptions.parse_query_error import ParseQueryError
+from exceptions.query_property_mapping_error import QueryPropertyMappingError
+
+
+class FlavorProperties(PropEnum):
+    """
+    An enum class for all flavor properties
+    """
+
+    FLAVOR_DESCRIPTION = auto()
+    FLAVOR_DISK = auto()
+    FLAVOR_EPHEMERAL = auto()
+    FLAVOR_ID = auto()
+    FLAVOR_IS_DISABLED = auto()
+    FLAVOR_IS_PUBLIC = auto()
+    FLAVOR_NAME = auto()
+    FLAVOR_RAM = auto()
+    FLAVOR_SWAP = auto()
+    FLAVOR_VCPU = auto()
+
+    @staticmethod
+    def from_string(val: str):
+        """
+        Converts a given string in a case-insensitive way to the enum values
+        """
+        try:
+            return FlavorProperties[val.upper()]
+        except KeyError as err:
+            raise ParseQueryError(
+                f"Could not find Flavor Property {val}. "
+                f"Available properties are {','.join([prop.name for prop in FlavorProperties])}"
+            ) from err
+
+    @staticmethod
+    def get_prop_mapping(prop):
+        """
+        Method that returns the property function if function mapping exists for a given FlavorProperty Enum
+        how to get specified property from an openstacksdk Flavor object is documented here:
+        https://docs.openstack.org/openstacksdk/latest/user/resources/compute/v2/flavor.html#openstack.compute.v2.flavor.Flavor
+        :param prop: A FlavorProperty Enum for which a function may exist for
+        """
+        mapping = {
+            FlavorProperties.FLAVOR_DESCRIPTION: lambda a: a["description"],
+            FlavorProperties.FLAVOR_DISK: lambda a: a["disk"],
+            FlavorProperties.FLAVOR_EPHEMERAL: lambda a: a["ephemeral"],
+            FlavorProperties.FLAVOR_ID: lambda a: a["id"],
+            FlavorProperties.FLAVOR_IS_DISABLED: lambda a: a["is_disabled"],
+            FlavorProperties.FLAVOR_IS_PUBLIC: lambda a: a["is_public"],
+            FlavorProperties.FLAVOR_NAME: lambda a: a["name"],
+            FlavorProperties.FLAVOR_RAM: lambda a: a["ram"],
+            FlavorProperties.FLAVOR_SWAP: lambda a: a["swap"],
+            FlavorProperties.FLAVOR_VCPU: lambda a: a["vcpus"],
+        }
+        try:
+            return mapping[prop]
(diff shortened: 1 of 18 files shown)


lib/enums/query/props/flavor_properties.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        Converts a given string in a case-insensitive way to the enum values
+        """
+        try:
+            return FlavorProperties[val.upper()]
+        except KeyError as err:
+            raise ParseQueryError(
+                f"Could not find Flavor Property {val}. "
+                f"Available properties are {','.join([prop.name for prop in FlavorProperties])}"
+            ) from err
```
DavidFair, line 34: Can we move this into the base class? As we're now up to three copies of the same method with just a slightly different error

lib/openstack_query/api/query_objects.py
```diff
@@ -31,3 +32,10 @@ def UserQuery() -> QueryAPI:
     Simple helper function to setup a query using a factory
     """
     return get_common(UserMapping)
+
+
+def FlavorQuery() -> QueryAPI:
```
DavidFair, line 37: We need to add this to the top level `__init__` too

lib/openstack_query/runners/flavor_runner.py
```diff
@@ -0,0 +1,66 @@
...
+            - see https://docs.openstack.org/api-ref/compute/#list-flavors-with-details
+        """
+        if not filter_kwargs:
+            # return all info
+            filter_kwargs = {"details": True}
+        logger.debug(
+            "running openstacksdk command conn.compute.flavors(%s)",
+            ",".join(f"{key}={value}" for key, value in filter_kwargs.items()),
+        )
+        return self._run_paginated_query(conn.compute.flavors, filter_kwargs)
```
DavidFair, line 52: Hmmm I'm not keen on the fact you (as a developer) have to know you need to call this from the base class

IMO, the parent of this should do this call:

```
def run():
    ret = self._run_query()
    self._run_paginated_query(ret)
```

Or similar would solve this from the base class

lib/openstack_query/runners/flavor_runner.py
```diff
@@ -0,0 +1,66 @@
...
+        """
+        This method is a helper function that will check a list of flavors to ensure that they are valid Flavor
+        objects
+        :param subset: A list of openstack Flavor objects
+        """
+
+        # should check validity that each flavor still exists but takes too long
+        if any(not isinstance(i, Flavor) for i in subset):
+            raise ParseQueryError("'from_subset' only accepts Flavor openstack objects")
+        return subset
```
DavidFair, line 66: Again this should be done in the base class:

```
@abstract_method
def get_openstack_object() -> Type:
   pass

def subset_valid():
   type = self.get_openstack_object()
    if any(not isinstance(i, Flavor) for i in subset):
            raise ParseQueryError(f"'from_subset' only accepts {type} openstack objects")
```

DavidFair (changes requested):

lib/openstack_query/runners/flavor_runner.py
anish-mudaraddi: well - sometimes you won't want to do a paginated query - for instance when we look for users - ideally run_paginated_query needs to be in a separate `RunnerUtils` class since `RunnerWrapper` is quite big already.

This is the case for many Wrapper classes already - so leaving this for another PR

lib/enums/query/props/flavor_properties.py
anish-mudaraddi: https://github.com/stfc/st2-cloud-pack/pull/122 - you want to basically me to do this right? I'll make a new PR

lib/openstack_query/runners/flavor_runner.py
anish-mudaraddi: I'll do this in another PR

lib/openstack_query/runners/runner_wrapper.py
```diff
@@ -246,16 +248,23 @@ def _run_query(
...
 
-    @abstractmethod
     def _parse_subset(
-        self, conn: OpenstackConnection, subset: List[OpenstackResourceObj]
+        self, _, subset: List[OpenstackResourceObj]
     ) -> List[OpenstackResourceObj]:
         """
         This method is a helper function that will check a subset of openstack objects and check their validity
-        :param conn: An OpenstackConnection object - used to connect to openstacksdk
+        :param _: An OpenstackConnection object - not used right now
```
meoflynn, line 256: If this isn't being used by the method, should we remove this parameter?

tests/lib/openstack_query/runners/test_flavor_runner.py
```diff
@@ -0,0 +1,85 @@
...
+    res = instance._parse_subset(
+        mock_openstack_connection, [mock_flavor_1, mock_flavor_2]
+    )
+    assert res == [mock_flavor_1, mock_flavor_2]
+
+
+def test_parse_subset_invalid(instance, mock_openstack_connection):
+    """
+    Tests _parse_subset works expectedly
+    method raises error when provided value which is not of User type
```
meoflynn, line 81: Should be referencing Flavor, not User type

meoflynn (changes requested):

tests/lib/openstack_query/api/test_query_objects.py
```diff
@@ -1,5 +1,13 @@
 from unittest.mock import patch, MagicMock
-from openstack_query.api.query_objects import UserQuery, ServerQuery, get_common
+
+import pytest
+
+from openstack_query.api.query_objects import (
+    UserQuery,
+    ServerQuery,
+    FlavorQuery,
+    get_common,
```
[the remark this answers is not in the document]
anish-mudaraddi: its a helper function which calls the factory pattern 
- this is a linting fix - this PR doesn't alter it

lib/openstack_query/runners/runner_wrapper.py
anish-mudaraddi: I guess we can remove it - I was thinking that we'd want to run some sort of validation on each resrouce - something like `find_<resource>` before using it - but I guess it rather slow

meoflynn (approved):

PR #153, opened 2023-10-13
ENH: add project query

anish-mudaraddi:
allows querying on openstack projects
Also fix a bug with client-side filters

### DIFF
--- a/lib/enums/query/props/project_properties.py
+++ b/lib/enums/query/props/project_properties.py
@@ -1,0 +1,62 @@
+from enum import auto
+from enums.query.props.prop_enum import PropEnum
+from exceptions.parse_query_error import ParseQueryError
+from exceptions.query_property_mapping_error import QueryPropertyMappingError
+
+
+class ProjectProperties(PropEnum):
+    """
+    An enum class for all server properties
+    """
+
+    PROJECT_DESCRIPTION = auto()
+    PROJECT_DOMAIN_ID = auto()
+    PROJECT_ID = auto()
+    PROJECT_IS_DOMAIN = auto()
+    PROJECT_IS_ENABLED = auto()
+    PROJECT_NAME = auto()
+    PROJECT_PARENT_ID = auto()
+
+    @staticmethod
+    def from_string(val: str):
+        """
+        Converts a given string in a case-insensitive way to the enum values
+        """
+        try:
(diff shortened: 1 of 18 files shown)


lib/enums/query/props/project_properties.py
```diff
@@ -0,0 +1,62 @@
...
+        """
+        mapping = {
+            ProjectProperties.PROJECT_DESCRIPTION: lambda a: a["description"],
+            ProjectProperties.PROJECT_DOMAIN_ID: lambda a: a["domain_id"],
+            ProjectProperties.PROJECT_ID: lambda a: a["id"],
+            ProjectProperties.PROJECT_IS_DOMAIN: lambda a: a["is_domain"],
+            ProjectProperties.PROJECT_IS_ENABLED: lambda a: a["is_enabled"],
+            ProjectProperties.PROJECT_NAME: lambda a: a["name"],
+            ProjectProperties.PROJECT_PARENT_ID: lambda a: a["parent_id"],
+        }
```
DavidFair, line 49: IMO this should be a class level variable, since calling this method will generate a new set of lambdas each time

Something like `_CLASS_MAPPING`

DavidFair: Or alternatively, maybe a dedicated getter on the interface, similar to `ServerProperties`

lib/openstack_query/runners/project_runner.py
```diff
@@ -0,0 +1,59 @@
...
+    ProjectRunner encapsulates running any openstacksdk Project commands
+    """
+
+    RESOURCE_TYPE = Project
+
+    def _parse_meta_params(self, _: OpenstackConnection, **__):
+        logger.error(
+            "ProjectQuery doesn't take any meta-params, if you think it should,"
+            "please raise an issue with the repo maintainers"
+        )
```
DavidFair, line 29: I would drop this error, `FlavorQuery has no meta-params available` is pretty descriptive enough tbh

DavidFair (changes requested):

lib/enums/query/props/project_properties.py
anish-mudaraddi: we can't make it a class_level variable - since the keys are based on enums - we get  `unresolved reference to ServerProperties`

DavidFair (approved):

PR #154, opened 2023-10-13
ENH: Add IP Adresses Property 

anish-mudaraddi:
Add IP addresses to ServerProperties

### DIFF
--- a/lib/enums/query/props/server_properties.py
+++ b/lib/enums/query/props/server_properties.py
@@ -73,6 +73,10 @@ class ServerProperties(PropEnum):
 
     @staticmethod
     def get_ips(obj):
+        """
+        A method to parse IPs from a server object and output as a comma-spaced string
+        :param obj: a openstacksdk server object
+        """
         return ", ".join(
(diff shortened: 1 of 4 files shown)


lib/enums/query/props/server_properties.py
```diff
@@ -54,6 +55,9 @@ def get_prop_mapping(prop):
             ServerProperties.FLAVOR_ID: lambda a: a["flavor_id"],
             ServerProperties.IMAGE_ID: lambda a: a["image_id"],
             ServerProperties.PROJECT_ID: lambda a: a["location"]["project"]["id"],
+            ServerProperties.ADDRESSES: lambda a: ", ".join(
+                [ip["addr"] for network in a["addresses"].values() for ip in network]
+            ),
```
DavidFair, line 60: This should be in a separate method, else we can't easily unit test it and it's non-trivial

DavidFair (changes requested):

DavidFair (approved):

PR #157, opened 2023-10-16
MAINT: separate out executing and parsing

anish-mudaraddi:
methods which handle sorting, grouping and generating output now decoupled with actually running the query - now these things can be changed at anytime and will only be calculated at the time of output without needing to re-run the query

I.e. you can do things like:
```
q1 = ServerQuery().where(...).run(...)
print(q1.group_by(a...).select(a, b, c).to_list())
print(q1.select(d, e, f).group_by(b..).sort_by(c..).to_string())
```

This will also have the hidden caveat that calling `select` will overwrite what was selected from previous calls - which means that a single select call must be used to set properties when outputting - which seems an okay trade-off
i.e. This is not possible anymore
```
q1.select(a).select(b).to_string()
```
you must do 
```
q1.select(a, b).to_string()
```

### DIFF
--- a/lib/openstack_query/api/query_api.py
+++ b/lib/openstack_query/api/query_api.py
@@ -23,11 +23,6 @@ class QueryAPI:
         self.output = query_components.output
         self._query_results = None
         self._query_results_as_objects = None
-
-        # setup executer functions
-        self.executer.parse_func = self.parser.run_parser
-        self.executer.output_func = self.output.generate_output
-
         self._query_run = False
 
     def select(self, *props: PropEnum):
@@ -164,7 +159,9 @@ class QueryAPI:
         :param flatten: boolean which will flatten results if true
         :param groups: a list group to limit output by
         """
-        result_as_objects, selected_results = self.executer.parse_results()
+        result_as_objects, selected_results = self.executer.parse_results(
+            parse_func=self.parser.run_parser, output_func=self.output.generate_output
+        )
         results = result_as_objects if as_objects else selected_results
 
         if groups:
@@ -191,7 +188,10 @@ class QueryAPI:
         :param groups: a list group to limit output by
         :param kwargs: kwargs to pass to generate table
         """
-        _, selected_results = self.executer.parse_results()
+        _, selected_results = self.executer.parse_results(
+            parse_func=self.parser.run_parser, output_func=self.output.generate_output
+        )
+
         return self.output.to_string(selected_results, title, groups, **kwargs)
 
     def to_html(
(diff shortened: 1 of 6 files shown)


lib/openstack_query/api/query_api.py
```diff
@@ -24,6 +24,12 @@ def __init__(self, query_components: QueryComponents):
         self._query_results = None
         self._query_results_as_objects = None
 
+        # setup executer functions
+        self.executer.parse_func = self.parser.run_parser
+        self.executer.output_func = self.output.generate_output
```
DavidFair, line 29: This is a smell, the executor should return whatever for this level to forward onto `self.parser.run_parser` rather than make them implicitly linked through the executor

The parse method call in executor should be moved up to here. It also means the executor's methods don't just become god methods in themselves

lib/openstack_query/query_blocks/query_executer.py
```diff
@@ -79,6 +84,9 @@ def parse_func(
 
     @property
     def output_func(self):
+        """
+        A getter for output function
+        """
         return self._output_func
```
DavidFair, line 90: Ditto output what I said about, where the `output_func` should not be in the executor, this should only run the query then return to the `QueryAPI` to marshall onto the class / method that handles output related things

DavidFair (changes requested):

lib/openstack_query/api/query_api.py
meoflynn: Why are the attributes for the parser and output been linked together under `self.executer`? I'm a little confused what `self.executer` does and what it's now doing

anish-mudaraddi: @meoflynn 
We need to perform two steps after generating raw results. Raw results in this case are openstack objects

1. First - group/sort them - handled by function `run_parser` 
2. Second  - convert openstack objects into dictionaries - to get only the properties we care about - handled by the function `generate_output` 

We pass a reference to these functions to the executer - so the executer can perform these steps when running `executer.parse_results` 

I did this mainly because I want the executer to be the only place to store a reference to the `raw_results`

anish-mudaraddi: @DavidFair I guess I can refactor `run_parser`, and `generate_output` to be parameters for `parse_results` instead

lib/openstack_query/query_blocks/query_executer.py
anish-mudaraddi: refactoring run_parser, and generate_output to be parameters for parse_results instead

lib/openstack_query/api/query_api.py
DavidFair: Alternatively, in the future we can move these to be child classes of the `executor` if you're worried about encapsulation, rather than passing methods as args

For the moment, I'm happy to take them as params just because the PR is an improvement and we can refactor in the future :+1:

DavidFair (approved):

meoflynn (approved):

PR #158, opened 2023-10-17
ENH: Add Chaining 

anish-mudaraddi:
### Description:

A rough draft of how we would use the query library to send decom flavor emails

add append_from and then public api methods

`then()` method allows creating a new query from a current query - this allows the query to essentially "pick-up" the results and run another query in a different paradigm

`append_from()` method runs a new query internally and shunts the output into the current query

And various Bugfixes

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/email_api/email_template_schemas.yaml
+++ b/lib/email_api/email_template_schemas.yaml
@@ -32,3 +32,11 @@ shutoff_error:
     errored_table: "No Errored VMs"
   html_filepath: "html/shutoff_error.html.j2"
   plaintext_filepath: "plaintext/shutoff_error.txt.j2"
+
+decom_email:
+  schema:
+    username: null
+    affected_flavors: "No Flavors"
+    decom_table: "No VMs affected"
+  html_filepath: "html/decom_email.html.j2"
+  plaintext_filepath: "plaintext/decom_email.txt.j2"
--- a/lib/openstack_query/api/query_api.py
+++ b/lib/openstack_query/api/query_api.py
@@ -7,7 +7,6 @@ from enums.query.props.prop_enum import PropEnum
 from enums.query.query_presets import QueryPresets
 from enums.query.query_types import QueryTypes
 from exceptions.parse_query_error import ParseQueryError
-from exceptions.query_chaining_error import QueryChainingError
 from structs.query.query_components import QueryComponents
 
 logger = logging.getLogger(__name__)
@@ -24,8 +23,6 @@ class QueryAPI:
         self.parser = query_components.parser
         self.output = query_components.output
         self.chainer = query_components.chainer
-        self._query_results = None
-        self._query_results_as_objects = None
         self._query_run = False
 
     def select(self, *props: PropEnum):
--- a/lib/openstack_query/api/query_api.py
+++ b/lib/openstack_query/api/query_api.py
@@ -23,11 +23,6 @@ class QueryAPI:
         self.output = query_components.output
         self._query_results = None
         self._query_results_as_objects = None
-
-        # setup executer functions
-        self.executer.parse_func = self.parser.run_parser
-        self.executer.output_func = self.output.generate_output
-
         self._query_run = False
 
     def select(self, *props: PropEnum):
@@ -164,7 +159,9 @@ class QueryAPI:
         :param flatten: boolean which will flatten results if true
         :param groups: a list group to limit output by
         """
-        result_as_objects, selected_results = self.executer.parse_results()
+        result_as_objects, selected_results = self.executer.parse_results(
+            parse_func=self.parser.run_parser, output_func=self.output.generate_output
+        )
         results = result_as_objects if as_objects else selected_results
 
         if groups:
@@ -191,7 +188,10 @@ class QueryAPI:
         :param groups: a list group to limit output by
         :param kwargs: kwargs to pass to generate table
         """
-        _, selected_results = self.executer.parse_results()
+        _, selected_results = self.executer.parse_results(
+            parse_func=self.parser.run_parser, output_func=self.output.generate_output
+        )
+
         return self.output.to_string(selected_results, title, groups, **kwargs)
 
     def to_html(
@@ -203,5 +203,7 @@ class QueryAPI:
         :param groups: a list group to limit output by
         :param kwargs: kwargs to pass to generate table
         """
-        _, selected_results = self.executer.parse_results()
+        _, selected_results = self.executer.parse_results(
+            parse_func=self.parser.run_parser, output_func=self.output.generate_output
+        )
         return self.output.to_html(selected_results, title, groups, **kwargs)
--- a/lib/openstack_query/api/query_api.py
+++ b/lib/openstack_query/api/query_api.py
@@ -4,8 +4,11 @@ from typing import Union, List, Optional, Dict, Tuple
 from custom_types.openstack_query.aliases import OpenstackResourceObj, PropValue
 from enums.cloud_domains import CloudDomains
 from enums.query.props.prop_enum import PropEnum
-from enums.query.query_presets import QueryPresets
+from enums.query.query_presets import QueryPresets, QueryPresetsGeneric
+from enums.query.query_types import QueryTypes
 from exceptions.parse_query_error import ParseQueryError
+from exceptions.query_chaining_error import QueryChainingError
+from openstack_query.query_factory import QueryFactory
 from structs.query.query_components import QueryComponents
 
 logger = logging.getLogger(__name__)
@@ -21,6 +24,7 @@ class QueryAPI:
         self.executer = query_components.executer
         self.parser = query_components.parser
         self.output = query_components.output
+        self.chainer = query_components.chainer
         self._query_results = None
         self._query_results_as_objects = None
         self._query_run = False
@@ -113,7 +117,6 @@ class QueryAPI:
         not specified in group mappings - ignored if group ranges not given
         """
         self.parser.parse_group_by(group_by, group_ranges, include_ungrouped_results)
-
         return self
 
     def run(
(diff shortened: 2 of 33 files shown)


lib/email_api/email_template_schemas.yaml
```diff
@@ -32,3 +32,11 @@ shutoff_error:
     errored_table: "No Errored VMs"
   html_filepath: "html/shutoff_error.html.j2"
   plaintext_filepath: "plaintext/shutoff_error.txt.j2"
+
+decom_email:
+  schema:
+    username: null
+    affected_flavors: "No Flavors"
+    decom_table: "No VMs affected"
```
DavidFair, line 40: Surely these should be null, rather than email people saying "nothing to do"?

lib/openstack_query/api/query_api.py
```diff
@@ -199,4 +206,79 @@ def to_html(
...
+            raise QueryChainingError(
+                f"Query Chaining Error: Could not find a way to chain current query into {query_type}"
+            )
+
+        curr_query_results = self.group_by(link_props[0]).to_list()
+        if not curr_query_results:
+            raise QueryChainingError(
+                "Query Chaining Error: No values found after running query set in {curr_query_obj_cls} aborting. "
+                "Have you run the query first?"
+            )
```
DavidFair, line 240: If we're going to ask the user to run the query, why not just emit a warning and run it on their behalf?

lib/openstack_query/api/query_api.py
```diff
@@ -199,4 +206,79 @@ def to_html(
...
+        Public method to chain current query into another query of a different type
+        and return the new query so that it will work only on the results of the original query.
+        NOTE - query must be run first for this to work
+        NOTE - a shared common property must exist between this query and the new query
+            - i.e. both ServerQuery and UserQuery share the 'USER_ID' property so chaining is possible
+                - see Mappings for more chaining options
+
+        :param query_type: an enum representing the new query to chain into
+        :param forward_outputs: (default=True) boolean that if true - will forward outputs from t
+        his query (and previous chained queries) onto new query.
```
DavidFair, line 223: We should add something about `forward_outputs == False` simply runs another query based on the previous results as a filter without adding the fields to make it clear

lib/openstack_query/mappings/flavor_mapping.py
```diff
@@ -27,6 +29,22 @@ class FlavorMapping(MappingInterface):
...
+        """
+        Should return a dictionary containing property pairs mapped to query mappings.
+        This is used to define how to chain results from this query to other possible queries
+        """
+        # local import prevents circular import issue
+        from enums.query.query_types import QueryTypes
+
+        return {
+            QueryTypes.SERVER_QUERY: (
+                FlavorProperties.FLAVOR_ID,
```
DavidFair, line 43: Do we need `QueryTypes.SERVER_QUERY`, surely we can do something like:

```
def get_chain_mappings():
    return {ServerProperties.FLAVOR_ID: FlavourProperties.FLAVOR_ID}


# Chaining module
original_properties = query_output.get_props()
mappings = flavor.get_chain_mappings()
found_key = next(prop for prop in original_properties if prop in mappings, None)

if found_key:
   matching_key = mappings[found_key]
```

This only has to be written once by the chaining module, and it completely decouples property mappings from QueryTypes, since we don't really care at this level

This means that the chain_mapping in all the classes don't bloat with having to know what other query types link in too

lib/openstack_query/query_blocks/query_output.py
```diff
@@ -154,9 +176,28 @@ def generate_output(
...
+                    # this "should not" error because forwarded outputs should always be a super-set
+                    # but sometimes resolving the property fails for whatever reason
+                    # in this case just copy all the keys in one of the dictionaries in the list
+                    # since they should be equal and set the values to "Not Found"
+                    try:
+                        output_list = outputs[prop_val]
+                        # should only ever contain one element since it's grouped by a unique id
+                        prop_list.update(output_list[0])
+                    except KeyError:
+                        prop_list.update({key: "Not Found" for key in outputs.keys()})
```
DavidFair, line 195: Could we not fail noisy? I can see `Not Found` keys being really confusing for end-users

lib/workflows/email/send_decom_flavor_email.py
```diff
@@ -0,0 +1,141 @@
...
+    :param all_projects: A boolean which if selected will search in all projects
+    :param send_email: Actually send the email instead of printing what will be sent
+    :param as_html: Send email as html
+    :param kwargs: see EmailParams dataclass class docstring
+    """
+
+    if project_list and all_projects:
+        raise RuntimeError(
+            "given both project list and all_projects flag - please choose only one"
+        )
```
DavidFair, line 42: IMO: we should have a `validate` method in the workflow files, where this goes to as a convention

lib/workflows/email/send_decom_flavor_email.py
```diff
@@ -0,0 +1,141 @@
...
+
+    # flavor query
```
DavidFair, line 44: Nit: Redundant comment, why are we only doing public flavors would be better IMO

lib/workflows/email/send_decom_flavor_email.py
```diff
@@ -0,0 +1,141 @@
...
+    q1 = FlavorQuery().where(
```
DavidFair, line 45: Nit `q1` -> `flavor_query`

lib/workflows/email/send_decom_flavor_email.py
```diff
@@ -0,0 +1,141 @@
...
+        QueryPresetsGeneric.EQUAL_TO, FlavorProperties.FLAVOR_IS_PUBLIC, value=True
+    )
+    q1.where(
+        QueryPresetsGeneric.ANY_IN,
+        FlavorProperties.FLAVOR_NAME,
+        values=decom_flavor_list,
+    )
+    q1.run(cloud_account)
+    q1.sort_by((FlavorProperties.FLAVOR_ID, False))
```
DavidFair, line 54: We really need to kill off the bool before this gets merged, now we're writing things that use the public API

lib/workflows/email/send_decom_flavor_email.py
```diff
@@ -0,0 +1,141 @@
...
+
+    # server query
+    q2 = q1.then(query_type="SERVER_QUERY", forward_outputs=True).run(
```
DavidFair, line 57: Ditto on `q2` -> `server_query`, since I want this to be our example we start people with as it has a bit of everything

DavidFair: `forward_outputs` feels a bit clunky here, and I can't put my finger on why....

To bikeshed `then ... keep_previous_results = True` makes more sense from just speaking the API
rather than `then forward_results = True` which requires you to explain what a forward is

lib/workflows/email/send_decom_flavor_email.py
```diff
@@ -0,0 +1,141 @@
...
+        ServerProperties.ADDRESSES,
+    )
+
+    # user query
+    q3 = q2.then("USER_QUERY", forward_outputs=True).run(cloud_account)
+    q3.select(UserProperties.USER_NAME)
+    q3.group_by(UserProperties.USER_EMAIL)
+    decom_results = q3.to_list()
+
+    for email_addr, outputs in decom_results.items():
```
DavidFair, line 73: Nit: can we move this down into a method e.g. `email_on_demon_results(decom_results)`

lib/workflows/email/send_decom_flavor_email.py
```diff
@@ -0,0 +1,141 @@
...
+                            },
+                        ),
+                        EmailTemplateDetails(
+                            template_name="footer", template_params={}
+                        ),
+                    ],
+                    "email_to": email_addr,
+                    "as_html": as_html,
+                },
+                **kwargs,
```
DavidFair, line 110: I really don't like this, yes it does it in a single step but it's unweildly to look at
Can we do it as three statements

Also why do we make two different email template details? This feels really weird from an API so I think I'd want to change it to take a single `EmailTemplateDetails` object

Lastly, we should use kwargs such as `email_to` directly and not (ab)use the fact we can decompile a dict into kwargs, it's not very future developer friendly to do this way

DavidFair (changes requested):

lib/email_api/email_template_schemas.yaml
anish-mudaraddi: either way this won't be seen unless we're running some obscure tests

lib/openstack_query/api/query_api.py
anish-mudaraddi: which cloud account do we use? The user has to provide that info somehow - right now we only accept that at `run()`

anish-mudaraddi: underlying question here is:

1. Do we ask for the `cloud_account` upon creating the query object 
     - so the query is set only to work with those credentials (making it a little inflexible)
     
2. Or do we add `cloud_account` param to all possible methods like; 
    - `to_list`, `to_html`, `to_string`, `then` and possibly more as they get created

3. Or do we just force users to run `run()` first

something to think about

lib/openstack_query/mappings/flavor_mapping.py
anish-mudaraddi: it will have to be instead
```
found_key = next(prop for prop in original_properties if prop.name in mappings, None)
```
The distinction here is `prop.name` - we create a implicit equality here that the **names** should match between enums 

This is something I'd rather do explicitly - i.e. mappings the way I set out.

lib/openstack_query/query_blocks/query_output.py
anish-mudaraddi: Sure I guess?

This happens when there's a mismatch between the way we get properties between two link properties - this may be pretty common or pretty rare - I don't know as of yet

lib/workflows/email/send_decom_flavor_email.py
anish-mudaraddi: I mean we don't really need this check since it'll get caught in the query level

lib/workflows/email/send_decom_flavor_email.py
anish-mudaraddi: I assumed because non-public flavors are already decommissioned?

lib/workflows/email/send_decom_flavor_email.py
anish-mudaraddi: I agree - I'll change this is and clean it up in the actual PR

lib/workflows/email/send_decom_flavor_email.py
anish-mudaraddi: happy with that :+1:

lib/workflows/email/send_decom_flavor_email.py
anish-mudaraddi: 1. Email is constructed from multiple Templates - the header (decom_email) and footer (footer) 

we _can_ have a define a single dataclass hold multiple Templates - either way we need to define each template separately - as they take different params

2. Yea it's a mess I agree - I'll clean it up

lib/openstack_query/api/query_api.py
DavidFair: Ah I'd forgotten about the cloud account at the run portion, let's leave it as a limitation then :+1:

lib/openstack_query/mappings/flavor_mapping.py
DavidFair: Why do we need to prop name? Since the existing query will have a `ServerProperties.FLAVOR_ID` set somewhere in the select for it to work 

We just need to retrieve the keys passed to the select?

lib/openstack_query/query_blocks/query_output.py
DavidFair: Can we make it noisy and obvious to figure out if it is common or rare
We can make it more permissive in the future by re-adding this, but we can't easily restrict it once people rely on this

lib/workflows/email/send_decom_flavor_email.py
DavidFair: Still I'd like us to have explicit validation per workflow rather than relying on implicit validation in the query library

lib/workflows/email/send_decom_flavor_email.py
DavidFair: Worth an ask, might be just out of scope to make our lives easier or just out of scope period

lib/workflows/email/send_decom_flavor_email.py
DavidFair: Can't we just have a header and footer in `EmailTemplateDetails` - I can't think of a time we'd ever need `template_params` for the footer? So it can just be hardcoded to None/Empty dict

lib/openstack_query/mappings/flavor_mapping.py
anish-mudaraddi: Ignore - I was misunderstanding - I'm happy with this change :+1:

DavidFair: Also separately to this, if you don't do a `select` does that implicitly put all the `ServerProperties` enum fields into a n internally tracked list of properties or similar? As that would be a requirement of this

lib/workflows/email/send_decom_flavor_email.py
anish-mudaraddi: Happy to hardcode header and footer - but this doesn't fix your issue

The underlying question to ask here is 
will we ever require creating an email with multiple templates in the body? 

If yes - we'll need to define templates separately like we've done here - and build EmailParams datastructure from multiple templates. 

If no - I'd argue that we should support having multiple templates anyway - since we can make those templates DRY

DavidFair: Let's see what it's like with a cleaner call

I'm sure if it looked like
```
body_template = EmailTemplateDetails(...)
footer_template = EmailTemplateDetails(...)

email_api.send_email (
   email_templates = [body_template, footer_template],
   email_to = email_addr
)
```
I wouldn't have noticed / flagged it

lib/openstack_query/query_blocks/query_chainer.py
```diff
@@ -12,25 +16,75 @@ class QueryChainer:
...
         """
-        if not self.check_supported(query):
-            return None
-        return self._chain_mappings[query]
+
+        new_query_props = list(next_query.value.get_prop_mapping())
+        for start_chain_prop, check_prop in self._chain_mappings.items():
+            for target_prop in new_query_props:
+                if target_prop == check_prop:
+                    return start_chain_prop, target_prop
```
DavidFair, line 45: Why do we validate that `target_prop` == `check_prop`
Surely if the key is in the chaining dict, we can just assume that the target is valid? Rather than iterate both dictionaries, since this effectively means we could drop the mapping all-together?

lib/workflows/email/send_decom_flavor_email.py
anish-mudaraddi: @jacob-ward  says that they are - any outliers will be found separately if there are any

lib/workflows/email/send_decom_flavor_email.py
anish-mudaraddi: i've split up the method in a different way

lib/workflows/email/send_decom_flavor_email.py
anish-mudaraddi: @jacob-ward  says that they are - any outliers will be found separately if there are any

anish-mudaraddi:
closing this as its been done

PR #159, opened 2023-10-24
BUG: Fix all_tenants not being used as filter kwargs

anish-mudaraddi:
### Description:

Fixed a bug where "all_tenants" isn't being converted properly from meta_param into a filter kwarg 

This meant that we couldn't find all servers from all projects and it would print out the VMs on the project attached to the clouds.yaml file over and over :/

### DIFF
--- a/lib/openstack_query/runners/server_runner.py
+++ b/lib/openstack_query/runners/server_runner.py
@@ -118,6 +118,16 @@ class ServerRunner(RunnerWrapper):
                 "- or you are not running as admin"
             )
 
+        if "all_tenants" in meta_params:
+            filter_kwargs["all_tenants"] = meta_params["all_tenants"]
+
+        if "projects" not in meta_params:
+            logger.debug(
+                "running openstacksdk command conn.compute.servers (%s)",
+                ", ".join(f"{key}={value}" for key, value in filter_kwargs.items()),
+            )
+            return self._run_paginated_query(conn.compute.servers, dict(filter_kwargs))
+
         query_res = []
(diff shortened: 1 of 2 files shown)


lib/openstack_query/runners/server_runner.py
```diff
@@ -118,6 +118,16 @@ def _run_query(
...
 
+        if "all_tenants" in meta_params:
+            filter_kwargs["all_tenants"] = meta_params["all_tenants"]
+
+        if "projects" not in meta_params:
+            logger.debug(
+                "running openstacksdk command conn.compute.servers (%s)",
+                ", ".join(f"{key}={value}" for key, value in filter_kwargs.items()),
+            )
+            return self._run_paginated_query(conn.compute.servers, dict(filter_kwargs))
```
DavidFair, line 129: We should use the same code path as below, even for 1 project rather than one project being the exception and a different code path

Is there a way to do that, or set the project name to an empty string to get the same effect

DavidFair (changes requested):

lib/openstack_query/runners/server_runner.py
anish-mudaraddi: The code path you highlighted is when we are not provided any projects. 
In this case we will run the query on: 

a) all projects if provided `all_tenants` in the filter kwargs
b) on project specified in clouds.yaml if `all_tenants` not specified

this is much quicker than getting all projects and using the code path below

DavidFair (approved):

PR #160, opened 2023-10-24
BUG: change flavor id property mapping

anish-mudaraddi:
change ServerQuery property mapping for flavor-id - previous one seems to be None far too often - the new mapping is more reliable


dev-0pz (approved):
LGTM

PR #161, opened 2023-10-24
BUG: domain_id key error bug

anish-mudaraddi:
When not providing meta_param `domain_id` when running a `UserQuery` - an error occurs - this fix will set a default domain-id if no meta_param given


Koldstream (approved):
Looks good to me!

DavidFair:
@anish-mudaraddi Can we add a simple test that if `domain_id` isn't provided it sets a default param as a follow up PR?

PR #163, opened 2023-10-24
ENH: Add then() and append_from() API Methods

anish-mudaraddi:
Add these methods for external chaining functionality
see #158 for more info
This PR concerns itself with just external chaining functionality - split from the draft PR

### DIFF
--- a/lib/openstack_query/api/query_api.py
+++ b/lib/openstack_query/api/query_api.py
@@ -24,8 +24,6 @@ class QueryAPI:
         self.parser = query_components.parser
         self.output = query_components.output
         self.chainer = query_components.chainer
-        self._query_results = None
-        self._query_results_as_objects = None
         self._query_run = False
 
     def select(self, *props: PropEnum):
--- a/lib/openstack_query/api/query_api.py
+++ b/lib/openstack_query/api/query_api.py
@@ -4,8 +4,9 @@ from typing import Union, List, Optional, Dict, Tuple
 from custom_types.openstack_query.aliases import OpenstackResourceObj, PropValue
 from enums.cloud_domains import CloudDomains
 from enums.query.props.prop_enum import PropEnum
-from enums.query.query_presets import QueryPresets
 from enums.query.sort_order import SortOrder
+from enums.query.query_presets import QueryPresets
+from enums.query.query_types import QueryTypes
 from exceptions.parse_query_error import ParseQueryError
 from structs.query.query_components import QueryComponents
 
@@ -22,6 +23,7 @@ class QueryAPI:
         self.executer = query_components.executer
         self.parser = query_components.parser
         self.output = query_components.output
+        self.chainer = query_components.chainer
         self._query_results = None
         self._query_results_as_objects = None
         self._query_run = False
@@ -114,7 +116,6 @@ class QueryAPI:
         not specified in group mappings - ignored if group ranges not given
         """
         self.parser.parse_group_by(group_by, group_ranges, include_ungrouped_results)
-
         return self
 
     def run(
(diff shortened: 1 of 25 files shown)


anish-mudaraddi:
missing a test for query_types enum - but cannot do it properly due to a circular import issue 

This circular import is due to enums directory being outside openstack_query module meaning it needs to resolve imports located in `openstack_query.__init__.py` - not sure how to fix really

lib/openstack_query/api/query_api.py
```diff
@@ -208,3 +207,54 @@ def to_html(
...
+        """
+        if isinstance(query_type, str):
+            query_type = QueryTypes.from_string(query_type)
+
+        new_query = self.then(query_type, keep_previous_results=False)
+        new_query.select(*props)
+        new_query.run(cloud_account)
+
+        link_props = self.chainer.get_link_props(query_type)
+        new_query.group_by(link_props[1])
```
DavidFair, line 257: Shouldn't the `chainer` class handle this, since we're calling multiple method of this anyway and it's a form of chaining

lib/openstack_query/mappings/server_mapping.py
```diff
@@ -30,6 +34,19 @@ class ServerMapping(MappingInterface):
...
+    def get_chain_mappings():
+        """
+        Should return a dictionary containing property pairs mapped to query mappings.
+        This is used to define how to chain results from this query to other possible queries
+        """
+        return {
+            ServerProperties.USER_ID: UserProperties.USER_ID,
+            ServerProperties.PROJECT_ID: ProjectProperties.PROJECT_ID,
+            ServerProperties.FLAVOR_ID: FlavorProperties.FLAVOR_ID
+            # TODO hypervisor mapping, image mapping etc
```
DavidFair, line 47: Nit: drop this TODO, since we'll get it anyway as we need the other forms of chaining

lib/openstack_query/query_blocks/query_chainer.py
```diff
@@ -0,0 +1,92 @@
...
+    def get_link_props(
+        self, next_query: QueryTypes
+    ) -> Optional[Tuple[PropEnum, PropEnum]]:
+        """
+        Get the link property pair to use to convert from current query to new query
+        (if mapping exists)
+        :param next_query: next Query QueryTypes
+        """
+
+        new_query_props = list(next_query.value.get_prop_mapping())
```
DavidFair, line 39: Don't convert this to a list if we're going to iterate it, a generator would work fine here too

lib/openstack_query/query_blocks/query_chainer.py
```diff
@@ -0,0 +1,92 @@
...
+        if isinstance(query_type, str):
+            query_type = QueryTypes.from_string(query_type)
+
+        link_props = curr_query.chainer.get_link_props(query_type)
+        if not link_props:
+            raise QueryChainingError(
+                f"Query Chaining Error: Could not find a way to chain current query into {query_type}"
+            )
+
+        curr_query_results = curr_query.group_by(link_props[0]).to_list()
```
DavidFair, line 74: We should add a comment here why the group_by for link_props is important, that we're going to group the chained query results by this key

lib/openstack_query/query_blocks/query_output.py
```diff
@@ -21,15 +20,38 @@ class QueryOutput:
     def __init__(self, prop_enum_cls: Type[PropEnum]):
         self._prop_enum_cls = prop_enum_cls
         self._props = set()
+        self._forwarded_outputs = {}
```
DavidFair, line 23: We should have a type info for this, as it's not obvious what we map from / to without studying the setters

lib/openstack_query/query_blocks/query_output.py
```diff
@@ -153,9 +175,51 @@ def generate_output(
...
-    def _parse_property(
+    def _parse_forwarded_outputs(
+        self, openstack_resource: OpenstackResourceObj
+    ) -> Dict[str, str]:
+        """
+        Generates a dictionary of forwarded outputs for the item given
+        :param openstack_resource: openstack resource item to parse forwarded outputs for
+        """
+        forwarded_output_dict = {}
+        if self.forwarded_outputs:
```
DavidFair, line 193: This should always be a dict, even if it's empty. Lets drop this `if` and let the for handle the case where there's 0 elements

lib/openstack_query/query_blocks/query_output.py
```diff
@@ -153,9 +175,51 @@ def generate_output(
...
+                        "This is due to a mismatch in property mappings - likely an Openstack issue"
+                    ) from exp
+
+                    # get keys from first set of forwarded outputs - and set it all to "Not Found"
+                    # (get first item from first group - keys should all be the same)
+                    # TODO: uncomment this block if the error is common
+                    # first_output_entry = list(outputs.values())[0][0]
+                    # forwarded_output_dict.update(
+                    #     {key: "Not Found" for key in list(first_output_entry.keys())}
+                    # )
```
DavidFair, line 219: Lets put this into an issue / docs and not leave commented code since it tends to rot

tests/enums/test_query_types.py
```diff
@@ -0,0 +1,46 @@
...
+from enums.query.query_types import QueryTypes
+from exceptions.parse_query_error import ParseQueryError
+
+
+@pytest.mark.parametrize("query_type", ["Flavor_Query", "FlAvOr_Query", "flavor_query"])
+def test_flavor_query_serialization(query_type):
+    """
+    Tests that variants of FLAVOR_QUERY can be serialized
+    """
+    assert QueryTypes.from_string(query_type) is QueryTypes.FLAVOR_QUERY
```
DavidFair, line 12: We only need a single case, since they all use the same code path and will throw:

```
@pytest.mark.parameterize("query_type", ["Flavor_Query", "SERVER_QUERY"]
```

Would cover all of our cases, since we shouldn't need to validate `enum` works (as it's not our code)

DavidFair (changes requested):

lib/openstack_query/api/query_api.py
anish-mudaraddi: huge headache dealing with circular imports

lib/openstack_query/query_blocks/query_chainer.py
anish-mudaraddi: made #164

tests/enums/test_query_types.py
anish-mudaraddi: I'm leaving this as is - its the same way we tested other enums with `from_string` method. 
These tests can be adapted in the future if we decide to add aliases for query types same as aliases for properties/presets

DavidFair (dismissed):

tests/lib/openstack_query/test_query_factory.py
```diff
@@ -1,45 +1,83 @@
...
+    @patch("openstack_query.query_factory.QueryExecuter")
+    @patch("openstack_query.query_factory.QueryChainer")
+    @patch("openstack_query.query_factory.QueryComponents")
+    def _run_build_query_deps_test_case(
+        mock_forwarded_outputs,
+        mock_query_components,
+        mock_chainer,
+        mock_executer,
+        mock_parser,
+        mock_output,
```
meoflynn, line 24: Could we add docstrings to these two methods?

meoflynn (changes requested):

meoflynn (approved):

PR #167, opened 2023-10-27
DOC: Add Query API README.md

anish-mudaraddi:
add README for query API along with how to use it

### DIFF
--- a/lib/openstack_query/README.md
+++ b/lib/openstack_query/README.md
@@ -1,0 +1,375 @@
+### Note: This is under active development
+### TODO: The Query Library needs to be moved out of StackStorm and into a standalone repo
+
+# Openstack Query Library
+
+A Python Package to query for Openstack Resources.
+
+Built on top of the [openstacksdk](https://docs.openstack.org/openstacksdk/latest/) to allow for more complex queries.
+
+Our overall goal is to provide a better, easier-to-use interface to run openstack queries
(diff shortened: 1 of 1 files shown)


lib/openstack_query/README.md
```diff
@@ -0,0 +1,375 @@
...
+```python
+print(query.to_list(flatten=False))
+
+[
+
+  # first result
+  { "project_name": "foo", "project_id": "bar" },
+
+  # second result
+  { "project_name": "foo1", "project_id": "bar2" },
```
khalford, line 223: Second result: "bar2" looks like it should be "bar1" to match "foo1".

lib/openstack_query/README.md
```diff
@@ -0,0 +1,375 @@
...
+If the results are grouped:
+
+Instead of returning:
+```python
+print(grouped_query.to_list(flatten=False))
+
+{
+    "group1": [
+        { "project_name": "foo", "project_id": "bar" },
+        { "project_name": "foo1", "project_id": "bar2" },
```
khalford, line 247: "bar2" looks like it should be "bar1" to match "foo1".

khalford (changes requested):

lib/openstack_query/README.md
anish-mudaraddi: heh yea - changing

DavidFair (approved):

PR #168, opened 2023-10-30
MAINT: update to_list API method into to_objects, to_props

anish-mudaraddi:
since to_list() can output a dictionary (if results are grouped) so the method may be misleading

change to:
- to_objects() - to return openstacksdk objects
- to_props() - to return selected properties

### DIFF
--- a/lib/openstack_query/api/query_api.py
+++ b/lib/openstack_query/api/query_api.py
@@ -150,20 +150,41 @@ class QueryAPI:
         self._query_run = True
         return self
 
-    def to_list(
-        self, as_objects=False, flatten=False, groups: Optional[List[str]] = None
-    ) -> Union[Dict[str, List], List[Dict[str, str]]]:
+    def to_objects(
+        self, groups: Optional[List[str]] = None
+    ) -> Union[Dict[str, List], List]:
         """
-        Public method to return results as a list (ungrouped) or dict (if grouped/flattened)
-        :param as_objects: if true return result as openstack objects
-        :param flatten: boolean which will flatten results if true
-        :param groups: a list group to limit output by
+        Public method to return results as openstack objects
+        :param groups: a list of group keys to limit output by
+        :return: List of openstacksdk resource objects or Dictionary of grouped openstacksdk resource objects
         """
-        result_as_objects, selected_results = self.executer.parse_results(
+        results, _ = self.executer.parse_results(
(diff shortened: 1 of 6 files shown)


lib/openstack_query/api/query_api.py
```diff
@@ -150,20 +150,41 @@ def run(
...
+                    f"Group(s) given are invalid - valid groups {list(results.keys())}"
+                )
+            return {group_key: results[group_key] for group_key in groups}
+        return results
 
+    def to_props(
+        self, flatten: bool = False, groups: Optional[List[str]] = None
+    ) -> Union[Dict[str, List], List]:
+        """
+        Public method to return results as openstack objects
```
khalford, line 180:
```suggestion
        """
        Public method to return results as openstack properties
```

tests/lib/openstack_query/query_blocks/test_query_chainer.py
```diff
@@ -44,7 +44,7 @@ def _run_parse_then_query_valid(
         mock_curr_query.chainer.get_link_props.return_value = ("curr-prop", "new-prop")
 
         mock_curr_query_results = {"prop-val1": "out1", "prop-val2": "out2"}
-        mock_curr_query.group_by.return_value.to_list.return_value = (
+        mock_curr_query.group_by.return_value.to_props.return_value = (
```
khalford, line 47: "curr" is not self-explanatory unless I am missing something.

khalford (changes requested):

lib/openstack_query/api/query_api.py
```diff
@@ -150,20 +150,41 @@ def run(
...
-    ) -> Union[Dict[str, List], List[Dict[str, str]]]:
+    def to_objects(
+        self, groups: Optional[List[str]] = None
+    ) -> Union[Dict[str, List], List]:
         """
-        Public method to return results as a list (ungrouped) or dict (if grouped/flattened)
-        :param as_objects: if true return result as openstack objects
-        :param flatten: boolean which will flatten results if true
-        :param groups: a list group to limit output by
+        Public method to return results as openstack objects
```
DavidFair, line 157: `this is either returned as a list if no groups are specified, or as a dict if they grouping was requested`

DavidFair (changes requested):

khalford (approved):

DavidFair (approved):

PR #169, opened 2023-10-31
BUG: Duplicates are no longer removed implicitly

anish-mudaraddi:
Queries now don't implicitly remove duplicates. Especially when chaining.

This fixes issue with keep_previous_results not working with Many-to-One type queries.

flatten() does not remove duplicates any more

### DIFF
--- a/lib/openstack_query/query_blocks/query_chainer.py
+++ b/lib/openstack_query/query_blocks/query_chainer.py
@@ -71,23 +71,36 @@ class QueryChainer:
                 f"Query Chaining Error: Could not find a way to chain current query into {query_type}"
             )
 
-        # we group the current results by the link property as this is the way we forward results
-        current_query_results = current_query.group_by(link_props[0]).to_props()
-        if not current_query_results:
+        if not current_query.to_props():
             raise QueryChainingError(
                 "Query Chaining Error: No values found after running this query - aborting. "
                 "Have you run the query first?"
             )
 
+        selected_props = current_query.output.selected_props
+
+        # grab all link prop values - including duplicates
+        search_values = current_query.select(link_props[0]).to_props(flatten=True)[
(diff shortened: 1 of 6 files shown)


meoflynn (dismissed):

lib/openstack_query/query_blocks/query_chainer.py
```diff
@@ -71,23 +71,36 @@ def parse_then(
...
+
+        # reset select
+        current_query.select(*selected_props)
+
         to_forward = None
         if keep_previous_results:
-            to_forward = (link_props[1], current_query_results)
+            # we group the current results by the link property as this is the way we forward results
+            to_forward = (
+                link_props[1],
```
DavidFair, line 94: Is the link prop always going to be the second element

lib/openstack_query/query_blocks/query_output.py
```diff
@@ -302,8 +306,5 @@ def _flatten_list(data: List[Dict]) -> Dict:
         res = {}
         for key in keys:
             vals = [d[key] for d in data]
-            # converting into a set preserving order
-            # https://stackoverflow.com/a/53657523
-            vals = list(dict.fromkeys(vals))
-            res[key] = vals
+            res[key] = list(vals)
```
DavidFair, line 309: Nit: this can be done as
```
res[key] = [d[key] for d in data]
```

Since you ask for a list then cast to a list again

DavidFair (changes requested):

lib/openstack_query/query_blocks/query_output.py
anish-mudaraddi: it's already a list because list comprehension above

DavidFair (approved):

meoflynn (approved):

PR #170, opened 2023-10-31
BUG: fix one-to-many chaining bug

anish-mudaraddi:
When going back from UserQuery to ServerQuery - we get a singleton list for each user-id we fine - each of these can match to multiple servers - i.e. a single entry shunting into many other entries - one-to-many.

This is a hacky fix where if grouped properties only contains one item - we don't delete it.
This works because:

- many-to-one relationships will contain a list with multiple items - each will be assigned to a duplicate as the output is parsed - hence deleting each one after being read is fine

- one-to-one, one-to-many relationships will contain a single list - each can be assigned to any number of outputs - hence we need to keep them around - we know it only contains one item - so not deleting it is fine

I'm happy to refactor this if it doesn't make sense to people - it's just to get it to work

### DIFF
--- a/lib/openstack_query/query_blocks/query_output.py
+++ b/lib/openstack_query/query_blocks/query_output.py
@@ -206,7 +206,12 @@ class QueryOutput:
                 # forwarded properties might contain more than one value
                 # then() will keep duplicates so each one in the list will be shunted into an output
                 forwarded_output_dict.update(output_list[0])
-                del output_list[0]
+
+                # a hacky way to prevent one-to-many chaining erroring - keep at least one value
+                # one-to-many/one-to-one will only ever contain one value per grouped_value
+                # many-to-one will contain multiple values per grouped values
+                if len(output_list) > 1:
+                    del output_list[0]
             except KeyError as exp:
                 raise QueryChainingError(
                     "Error: Chaining failed. Could not attach forwarded outputs.\n"
--- a/lib/openstack_query/query_blocks/query_chainer.py
+++ b/lib/openstack_query/query_blocks/query_chainer.py
@@ -77,15 +77,14 @@ class QueryChainer:
                 "Have you run the query first?"
             )
 
-        selected_props = current_query.output.selected_props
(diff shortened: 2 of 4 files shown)


lib/openstack_query/query_blocks/query_output.py
```diff
@@ -206,7 +206,12 @@ def _parse_forwarded_outputs(
                 # forwarded properties might contain more than one value
                 # then() will keep duplicates so each one in the list will be shunted into an output
                 forwarded_output_dict.update(output_list[0])
-                del output_list[0]
+
+                # a hacky way to prevent one-to-many chaining erroring - keep at least one value
+                # one-to-many/one-to-one will only ever contain one value per grouped_value
+                # many-to-one will contain multiple values per grouped values
+                if len(output_list) > 1:
+                    del output_list[0]
```
DavidFair, line 214: Can we add a small test to check a list `len ==1` doesn't throw, since this is something small that can be easily missed

anish-mudaraddi: already exists the following tests test for this: 
`test_parse_forwarded_outputs_one_set` - tests one set of forwarded outputs with a singleton list

and
`test_parse_forwarded_outputs_with_duplicates` - tests one set of forwarded outputs with list > 1

DavidFair: Ah but this didn't fail when we deleted the only element? Since nothing failed prior to this bug is my concern

anish-mudaraddi: I've modified the tests for parse_forwarded_output accordinly to check.
`test_parse_forwarded_outputs_one_to_many` - testing that they aren't deleted (singleton list)
`test_parse_forwarded_outputs_many_to_one` - testing that they are deleted
`test_parse_forwarded_outputs_multiple_sets` - testing when given both one_to_many set and many_to_one set

meoflynn (approved):

DavidFair (approved):

PR #171, opened 2023-10-31
ENH: Add reset() to API

anish-mudaraddi:
function simply reset query parameters.

### DIFF
--- a/lib/openstack_query/api/query_api.py
+++ b/lib/openstack_query/api/query_api.py
@@ -285,3 +285,28 @@ class QueryAPI:
 
         self.output.update_forwarded_outputs(link_props[0], new_query.to_props())
         return self
+
+    def reset(self, hard=False):
+        """
+        function that when called will reset the query back to original state.
+        This means that any pre-configured info like select(), where(), group_by() and sort_by() will be deleted
+        :param hard: boolean (default False). If true will delete any cached results from previous run() command
+        """
+        # reset select
+        self.output.selected_props = set()
+
+        # reset where
+        self.builder.client_side_filters = []
+        self.builder.server_side_filters = []
+        self.builder.server_filter_fallback = []
+
+        # reset group_by
+        self.parser.group_by = None
+        self.parser.group_mappings = {}
+
+        # reset sort_by
+        self.parser.sort_by = {}
+
+        if hard:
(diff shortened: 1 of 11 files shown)


lib/openstack_query/api/query_api.py
```diff
@@ -285,3 +285,28 @@ def append_from(
 
         self.output.update_forwarded_outputs(link_props[0], new_query.to_props())
         return self
+
+    def reset(self, hard=False):
```
DavidFair, line 289: Bike shedding, reset should reset everything

lib/openstack_query/api/query_api.py
```diff
@@ -285,3 +285,28 @@ def append_from(
...
+        # reset group_by
+        self.parser.group_by = None
+        self.parser.group_mappings = {}
+
+        # reset sort_by
+        self.parser.sort_by = {}
+
+        if hard:
+            self.executer.raw_results = []
+        return self
```
DavidFair, line 312: This means that `query_api` now knows how the internal of a different class works. Can we not use the existing factory to regenerate the builder / parser / etc... since it means a reset is the same as an init?

DavidFair (changes requested):

Koldstream (approved):

lib/openstack_query/api/query_api.py
anish-mudaraddi: alternatively I could implement a reset() on each block that the API method calls

anish-mudaraddi: I don't want the Query API accessing the Factory - all kinds of circular import headaches

lib/openstack_query/api/query_api.py
anish-mudaraddi: disagree - we can use reset() for this use-case:

you run a query and group-by/sort-by 

you want to output same query results but without grouping and sorting - `reset(hard=False)` means they don't have to run the query again

DavidFair:
@anish-mudaraddi as per our in-person discussion, did you want to tweak it to get `init` to call reset so they use the same code path?

anish-mudaraddi:
I'm closing this for now - I'll come back to this once #182 is in

PR #173, opened 2023-11-02
MAINT: improve query chaining

anish-mudaraddi:
A draft for improving how forwarded results are handled

These are a series of changes on how we calculate and store forwarded outputs from previous queries

Previously we evaluated these at the time of outputting. 
However, this complicates the code and slows down outputting significantly
- Also the data-structure was very complex to store forwarded outputs

Now we store a dictionary of forwarded outputs alongside the objects generated at the run() step.
    - we now parse the forwarded outputs at the earliest possible step - right after running the query
    - this couples forwarded outputs to the openstack object - and makes it much quicker to output

We've also simplified the datastructure holding forwarded outputs to just a dictionary of key-value pairs - so much easier to work with.

### DIFF
--- a/lib/openstack_query/query_blocks/query_chainer.py
+++ b/lib/openstack_query/query_blocks/query_chainer.py
@@ -1,5 +1,6 @@
 from typing import Tuple, List, Optional, Union
 
+from enums.cloud_domains import CloudDomains
 from enums.query.props.prop_enum import PropEnum
 from enums.query.query_presets import QueryPresetsGeneric
 from enums.query.query_types import QueryTypes
@@ -107,4 +108,33 @@ class QueryChainer:
             QueryPresetsGeneric.ANY_IN,
             link_props[1],
             values=search_values,
+        )
+
+    @staticmethod
+    def parse_append_from(
+        current_query,
+        query_type: Union[str, QueryTypes],
+        cloud_account: Union[str, CloudDomains],
+        *props: PropEnum,
+    ):
+        """
+        Public static method to parse an append_from call - which appends specific properties from another query onto
+        current query.
+        :param current_query: current QueryAPI object
+        :param query_type: an enum representing the new query we want to append properties from
+        :param cloud_account: A String or a CloudDomains Enum for the clouds configuration to use
+        :param props: one or more properties to collect described as enum
+        """
+        if isinstance(query_type, str):
+            query_type = QueryTypes.from_string(query_type)
+
+        new_query = current_query.then(query_type, keep_previous_results=False)
+        new_query.select(*props)
+        new_query.run(cloud_account)
+
+        link_props = current_query.chainer.get_link_props(query_type)
+
+        current_query.executer.append_forwarded_results(
+            new_query.group_by(link_props[1]).to_props(),
+            lambda obj: current_query.output.parse_property(link_props[0], obj),
         )
--- a/lib/openstack_query/query_blocks/query_executer.py
+++ b/lib/openstack_query/query_blocks/query_executer.py
@@ -1,3 +1,4 @@
+import copy
 import time
 import logging
 from typing import Tuple, Callable, Optional, Dict, List, Any, Union, Type
@@ -10,6 +11,7 @@ from custom_types.openstack_query.aliases import (
     OpenstackResourceObj,
     ServerSideFilters,
     ClientSideFilters,
+    PropValue,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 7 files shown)


lib/openstack_query/query_blocks/query_chainer.py
```diff
@@ -108,3 +109,32 @@ def parse_then(
...
+        new_query = current_query.then(query_type, keep_previous_results=False)
+        new_query.select(*props)
+        new_query.run(cloud_account)
+
+        link_props = current_query.chainer.get_link_props(query_type)
+
+        current_query.executer.append_forwarded_results(
+            new_query.group_by(link_props[1]).to_props(),
+            lambda obj: current_query.output.parse_property(link_props[0], obj),
+        )
```
DavidFair, line 140: IMO: This should return the result to the query api which should do:

```
def append_from(...):
   results = self.chainer.parse_append_from(self, query_type, cloud_account, *props)
   self.executor.append_forwarded_results(results[1] ...)
   return self
```

Since we're gradually ending up with the marshalling causing the various models to tightly couple up

lib/openstack_query/query_blocks/query_executer.py
```diff
@@ -29,7 +31,11 @@ def __init__(self, prop_enum_cls: Type[PropEnum], runner_cls: Type[RunnerWrapper
         self.runner = runner_cls(self._prop_enum_cls.get_marker_prop_func())
         self._client_side_filters = None
         self._server_side_filters = None
-        self._raw_results = []
+        self._results = []
+
+        # for chaining
+        self._forwarded_values = {}
+        self._link_prop_func = lambda a: None
```
DavidFair, line 38: Can't we just store this? Instead of passing a lambda in lets just do

```
        current_query.executer.append_forwarded_results(
            new_query.group_by(link_props[1]).to_props(),
            self._link_prop = parse_property(link_props[0], obj),
        )
```

lib/openstack_query/query_blocks/query_executer.py
```diff
@@ -29,7 +31,11 @@ def __init__(self, prop_enum_cls: Type[PropEnum], runner_cls: Type[RunnerWrapper
         self.runner = runner_cls(self._prop_enum_cls.get_marker_prop_func())
         self._client_side_filters = None
         self._server_side_filters = None
-        self._raw_results = []
+        self._results = []
+
+        # for chaining
+        self._forwarded_values = {}
```
DavidFair, line 37: Hmm I'm not keen on this because now we're leaking the concept of chaining throughout our codebase

IMO, the chaining should be isolated and effectively run the executor in a way it can't tell if the user or chainer has invoked this path rather than it being a special case?

lib/openstack_query/query_blocks/query_executer.py
```diff
@@ -29,7 +31,11 @@ def __init__(self, prop_enum_cls: Type[PropEnum], runner_cls: Type[RunnerWrapper
         self.runner = runner_cls(self._prop_enum_cls.get_marker_prop_func())
         self._client_side_filters = None
         self._server_side_filters = None
-        self._raw_results = []
+        self._results = []
```
DavidFair, line 34: Is this a list of tuples now? I think we should have some typing on this

lib/openstack_query/query_blocks/query_executer.py
```diff
@@ -96,37 +127,89 @@ def run_query(
...
             client_side_filters=self.client_side_filters,
             server_side_filters=self.server_side_filters,
             from_subset=from_subset,
             **kwargs,
         )
+
+        forwarded_results = copy.deepcopy(self._forwarded_values)
+        self.results = self.apply_forwarded_results(
+            results, forwarded_results, self.link_prop_func
+        )
```
DavidFair, line 141: I don't see why we need to do the copies if we have a list of objects?
I was thinking you simply have something like:

```
if not results:
   # First query
   self.results = [() for _ in len(query_results)

for dest_tuple, openstack_obj in zip(self.results, query_results)
   dest_tuple.append(openstack_obj)
```

since `append_forward_results` is a bit of a beast currently:

DavidFair: To handle the removal case which would look something like:

```
new_results = []

for dest_tuple, openstack_obj in zip(self.results, query_results)
   was_removed = False
   for obj in dest_tuple:
        # check if any of the fields match, if none do we can assume this tuple was removed
        if not any(getattr(field_name, obj) == getattr(field_name, openstack_obj) for field_name in openstack_obj):
             continue

   dest_tuple.append(openstack_obj)
   new_results.append(dest_tuple)
self.results = new_results
```

lib/openstack_query/query_blocks/query_executer.py
anish-mudaraddi: so pass forwarded values as a parameter to `run_query`?
and store forwarded_values in `QueryChainer?`
The Query needs to store forwarded outputs somewhere so if the user calls `run()` again we need to evaluate it again

lib/openstack_query/query_blocks/query_executer.py
anish-mudaraddi: I think i've addressed this since we're not storing link_prop in executer anymore

PR #174, opened 2023-11-02
ENH: Add email workflow for shutoff VMs

meoflynn:
### Description:

Adds a workflow and unit tests for automating emails to notify users about shutoff VMs

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:


lib/workflows/email_shutoff.py
```diff
@@ -0,0 +1,117 @@
...
+    """
+
+    query_obj = ServerQuery()
+
+    query_obj.select(
+        ServerProperties.SERVER_NAME
+    ).where(
+        QueryPresetsGeneric.EQUAL_TO,
+        ServerProperties.SERVER_STATUS,
+        value="SHUTOFF")
```
anish-mudaraddi, line 26: IMO: I think you should add an extra `where` clause to limit search to servers which haven't been updated in a while rather than all of them

```
.where(
    QueryPresetsDateTime.OLDER_THAN,
    ServerProperties.SERVER_LAST_UPDATED_DATE,
    days=int(days_threshold)
)
```
you can add `days_threshold` as an extra param to your workflow main function

lib/workflows/email_shutoff.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+
+    if project_id:
+        user_query.run(
+            "openstack",
+            # from_projects=project_id
+        )
+
+    else:
+        user_query.run("openstack", as_admin=True, all_projects=True)
```
anish-mudaraddi, line 58: you don't need to provide as_admin=True or all_projects=True for user_query. They are ignored here anyway

lib/workflows/email_shutoff.py
```diff
@@ -0,0 +1,117 @@
...
+    """
+
+    if not user_email:
+        user_email = "cloud-support@stfc.ac.uk"
+
+    test_params = EmailParams(
+        subject=subject_text,
+        email_from='cloud-support@stfc.ac.uk',
+        email_to=[user_email],
+        email_templates=[]
```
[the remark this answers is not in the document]
anish-mudaraddi: You need to write an email template using jinja2 - I'll document how to do that soon

lib/workflows/email_shutoff.py
```diff
@@ -0,0 +1,117 @@
...
+    :param: project_id: (Optional) A list of project IDs to query from
+    """
+
+    # get VMs shutoff, grouped by user, include user email in results
+    shutoff_vm = query_shutoff_vms(project_id)
+
+    for user in shutoff_vm.values():
+        user_email, server_name_list = prepare_user_server_email(user) # shutoff_vm[user] is list of dictionaries
+        print('output: ', user_email, server_name_list)
+        # prepare and send blank test email
```
anish-mudaraddi, line 115: comment can be removed

anish-mudaraddi (changes requested):
first pass through - looking at Query Library/StackStorm related stuff mostly

lib/workflows/email_shutoff.py
DavidFair: I think this can be changed simply to 
```
    user_query.select(
        UserProperties.USER_EMAIL
    ).group_by(
        UserProperties.USER_NAME
    )
    user_query.run("openstack")
```

(As per the previous comments, the hardcoded "openstack" needs removing too

lib/workflows/email_shutoff.py
```diff
@@ -0,0 +1,117 @@
...
+        subject=subject_text,
+        email_from='cloud-support@stfc.ac.uk',
+        email_to=[user_email],
+        email_templates=[]
+    )
+
+    Emailer(smtp_account).send_emails([test_params])
+
+
+def main(smtp_account: SMTPAccount, project_id: Optional[List[str]] = None):
```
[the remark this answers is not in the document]
DavidFair: Nit: `limit_by_project` would be better as `project_ids` since we use that terminology regularly in conversation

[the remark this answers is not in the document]
DavidFair: This is a separate discussion, but I was wondering if we could hardcode a name for our entry point (don't mind what) since this forces a future developer to find the file and the fact the `main` is actually `send_shutoff_server_email`

In another file it might be `delete_old_errored_vms` for example, so they would always have to refer to the yaml even if they found the workflow file

lib/workflows/email_shutoff.py
DavidFair: Ditto about the print

tests/lib/workflows/test_email_shutoff.py
```diff
@@ -0,0 +1,181 @@
...
+        subject=mock_subject,
+        email_from='cloud-support@stfc.ac.uk',
+        email_to=[mock_email],
+        email_templates=[]
+    )
+
+    email_object.send_emails.assert_called_once_with([expected_email_params])
+
+
+@patch('workflows.email_shutoff.Emailer')
```
DavidFair, line 120: Nit this one uses an `@patch` but the one above uses a `with` for the same code path, could we change this to use a `with` context manager like the last one?

DavidFair (changes requested):
Re: formatting this can easily be done locally with

```
pip install --user pre-commit
cd <path_to_this_repo>
pre-commit install
pre-commit run --all-files
```

This will re-format all your files, it also makes it so that `git commit` will re-format your files each time you go to commit

lib/workflows/email_shutoff.py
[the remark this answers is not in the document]
anish-mudaraddi: well query library allows you to specify both project names or ids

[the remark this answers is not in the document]
anish-mudaraddi: I'm fine with hardcoding a name - keeping it as "main" makes sense then  :+1:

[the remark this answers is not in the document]
DavidFair: `project_ids` and `project_names` then? More I just want to match the lingua franca used by the ops team since they're our main "customer"

lib/workflows/email_shutoff.py
DavidFair: I would ask @jacob-ward

lib/workflows/email_shutoff.py
```diff
@@ -0,0 +1,124 @@
...
+
+    # if project ID provided as parameter
+    if project_id:
+        query_obj.run(cloud_account, as_admin=True, from_projects=project_id)
+
+    else:
+        query_obj.run(cloud_account, as_admin=True, all_projects=True)
+
+    user_query = query_obj.then("USER_QUERY", keep_previous_results=True)
+
```
anish-mudaraddi, line 43: I found that this introduces a bug I had to fix in #184
This will ignore ERROR and SHUTOFF VMs which have a user_id which can't be found by running a UserQuery - either the users don't exist anymore or there's some issue in openstack. 

In these cases we should not ignore these values, but instead send these to cloud-support email

anish-mudaraddi (changes requested):

lib/workflows/email_shutoff.py
anish-mudaraddi: have a look at https://github.com/stfc/st2-cloud-pack/blob/main/lib/workflows/email/send_decom_flavor_email.py#L159-L174

I split out finding user info from the original query. 
1. Group the server query by user id
2. for each user id run a separate UserQuery
3. if the UserQuery returns nothing - set "default" values for user info (set cloud-support@stfc.ac.uk) as user_email
4. if UserQuery returns something - use that

anish-mudaraddi:
I'm going to take this one over and open up a new PR when I'm happy with it

PR #178, opened 2023-11-06
MAINT: Add Result and ResultContainer Classes

anish-mudaraddi:
#173 - a ready-to-review version of this

improves how query results and forwarded results are handled and stored.  The datastructure holding these parts together were getting too complicated

add Result class - query results are now stored in separate classes - keeping results encapsulated away from query blocks. 

Result class holds openstack item, and forwarded prop to attach
Modify QueryParser run functions to use list of Results instead 

add results container class which will store results and handle adding forwarded results and selecting results instead of QueryOutput.


anish-mudaraddi:
this is too big - closing and reopening as smaller PRs

PR #179, opened 2023-11-06
MAINT: add Results and ResultsQuery classes

anish-mudaraddi:
This PR does adds the following - part of a larger PR - see #173 for the full picture 

**`Result` class** 
which is used to store a single query result item. 
It stores the openstack item as well as any forwarded properties.
This class is responsible for outputting the value stored as_object or as_props. 
When as_props() is selected, it will concatenate any forwarded properties stored automatically 

**`ResultContainer` class** 
which is responsible for storing multiple Result objects and returning the total output to_objects/to_props. 
It is responsible for creating and managing each Result item once a query is run
Currently neither class is being used - Query Executer will make use of  this in future - this PR is being split up so it can be reviewed quicker

### DIFF
--- a/tests/lib/openstack_query/query_blocks/test_result_container.py
+++ b/tests/lib/openstack_query/query_blocks/test_result_container.py
@@ -299,3 +299,11 @@ def test_get_forwarded_result_one_to_many():
     )
 
     assert all(len(val) == 1 for val in forwarded_values.values())
+
+
+def test_get_forwarded_result_empty():
+    """
+    Tests get_forwarded_results static method, where forwarded results are empty
+    should return an empty dict
+    """
+    assert ResultsContainer._get_forwarded_result("grouped_value1", {}) == {}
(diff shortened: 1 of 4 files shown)


tests/lib/openstack_query/query_blocks/test_result_container.py
```diff
@@ -0,0 +1,300 @@
...
+    return _setup_instance_with_results
+
+
+def test_to_props_results_empty(setup_instance_with_results):
+    """
+    Test to_props method when results are empty - return empty list
+    """
+    instance = setup_instance_with_results([])
+    instance.to_props(MockProperties.PROP_1, MockProperties.PROP_2)
+
```
khalford, line 30: What is the pass/fail condition in this test? I cannot see if there is an assert in this or either called methods

khalford (approved):

tests/lib/openstack_query/query_blocks/test_result_container.py
anish-mudaraddi: good catch - I forgot to add an assertion

AspectVI (dismissed):
I don't see any issues with these changes, the logic seems right to me.

Unfortunatly I can't comment on the execution too much, due to not not knowing how to code half of this myself. But is does work so... that seem alright.

Anyway I recommend 1 more review from someone that can; review the implementation of the code, then this pr should be good.

khalford (dismissed):

khalford (approved):

AspectVI (approved):

PR #182, opened 2023-11-08
MAINT: Use Results and ResultsContainer Classes

anish-mudaraddi:
refactor query blocks to use results classes

QueryParser now expects a list of Result classes instead when running run_sort_by/run_group_by

QueryOutput no longer handles parsing selected properties - instead calls the function in ResultsContainer to get selected properties - simplifies this block quite a bit

QueryExecuter now creates and manages a ResultsContainer object instead of holding raw results itself

QueryChainer stores forwarded_properties from previous queries

### DIFF
--- a/lib/openstack_query/query_blocks/query_output.py
+++ b/lib/openstack_query/query_blocks/query_output.py
@@ -1,14 +1,9 @@
-from copy import deepcopy
 from typing import List, Dict, Union, Type, Set, Optional
 from tabulate import tabulate
 from enums.query.props.prop_enum import PropEnum
-from custom_types.openstack_query.aliases import (
-    OpenstackResourceObj,
-    PropValue,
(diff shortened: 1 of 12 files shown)


lib/openstack_query/query_blocks/query_output.py
```diff
@@ -59,43 +38,80 @@ def selected_props(self, props=Set[PropEnum]):
         """
         self._props = props
 
+    @staticmethod
+    def _validate_groups(
+        results: Union[List, Dict], groups: Optional[List[str]] = None
+    ):
```
khalford, line 44: doc

khalford (changes requested):

khalford (approved):

DavidFair (approved):

PR #183, opened 2023-11-09
BUG: various bugfixes on send email decom flavors

anish-mudaraddi:
catch cases when flavor given is invalid + when no servers found with the flavor and output useful errors

add use_override flag to the stackstorm action

### DIFF
--- a/actions/email.users.with.decom.flavors.yaml
+++ b/actions/email.users.with.decom.flavors.yaml
@@ -47,6 +47,11 @@ parameters:
     description: "tick to search in all projects - default True"
     required: true
     default: true
+  use_override:
+    type: boolean
+    description: "TESTING: Set this flag so ALL emails will be redirected to override email"
+    required: true
+    default: False
   as_html:
     type: boolean
     description: "Send email body as HTML"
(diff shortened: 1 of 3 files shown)


khalford (dismissed):
LGTM

actions/email.users.with.decom.flavors.yaml
```diff
@@ -47,6 +47,11 @@ parameters:
     description: "tick to search in all projects - default True"
     required: true
     default: true
+  use_override:
+    type: boolean
+    description: "TESTING: Set this flag so ALL emails will be redirected to override email"
```
DavidFair, line 52: Nit: can we drop `TESTING` it makes it look this doesn't work

Also don't we have override email? If so we should just use that instead of another bool for an optional field

DavidFair (changes requested):

actions/email.users.with.decom.flavors.yaml
anish-mudaraddi: `override_email_address` is the email we use when a VM doesn't have a email set 
`use-override` is a flag which makes it so every email gets sent to one set in `override_email_address`

DavidFair (approved):

khalford (approved):

PR #187, opened 2023-11-28
DOC: adds docs for query library stuff

anish-mudaraddi:
Adds docs for query library 

~~TODO: I need to add links from README.md to the docs and between each doc, but they can only be added once this is merged~~ This is done now


lib/openstack_query/docs/PRESETS.md
```diff
@@ -0,0 +1,245 @@
...
+    ```python
+    class QueryPresetsGeneric(QueryPresets):
+        """
+        Enum class which holds generic query comparison operators
+        """
+
+        EQUAL_TO = auto()
+        ...
+        NEW_PRESET = auto() # <- we add this line to repesent a new preset enum belonging to the 'Generic' group
+    ```
```
meoflynn, line 105: This code block needs to shift to the left as it's not displaying correctly

lib/openstack_query/docs/QUERIES.md
```diff
@@ -0,0 +1,232 @@
...
+           # server-side filters and be used when calling the specific openstacksdk query which gets the resources being queried for
+           ...
+
+       def _run_query(
+           self,
+           conn: OpenstackConnection,
+           filter_kwargs: Optional[ServerSideFilters] = None,
+           **meta_params,
+       )
+           # Deine logic here that will setup keyword arguments - merging server-side kwargs (filter_kwargs) and meta_param
```
meoflynn, line 141: Typo here for `Define`

meoflynn (changes requested):
I've done an initial review of the docs, they look good so far 😄

lib/openstack_query/docs/QUERIES.md
```diff
@@ -0,0 +1,232 @@
...
+        return ResouceProperties
+
+    @staticmethod
+    def get_server_side_handler() -> ServerSideHandler:
+       ...
+        # here you need to define server-side filters. See PRESETS.md for details
+
+    @staticmethod
+    def get_client_side_handlers() -> ServerSideHandler:
+       ...
```
meoflynn, line 197: Could you add docstrings to these static methods like the methods above?

meoflynn (changes requested):

lib/openstack_query/docs/QUERIES.md
anish-mudaraddi: These aren't actual methods - its pseudo-code - basically instructions of methods you need to edit. 

I don't want people to copy-paste without reading so I didn't copy the entire method

lib/openstack_query/docs/PROPERTIES.md
```diff
@@ -0,0 +1,117 @@
...
+|---------------------|----------------------------------------------------------------------------------|
+| `ServerProperties`  | Properties related to Openstack Servers (VMs). Used by `ServerQuery` Query class |
+| `UserProperties`    | Properties related to Openstack Users. Used by `UserQuery` Query class           |
+| `ProjectProperties` | Properties related to Openstack Projects. Used by `ProjectQuery` Query class     |
+| `FlavorProperties`  | Properties related to Openstack Flavors. Used by `FlavorQuery` Query class       |
+
+
+## Property Reference
+
+`ServerProperties`
```
DavidFair, line 25: IMO each of these should be in their own files:

```
lib/openstack_query/docs/properties/server_properties.md
lib/openstack_query/docs/properties/user_properties.md
```

That way I can easily get to the relevant docs in vim/emacs/code/GH based on a file name rather than open a massive file and have to jump around

It also makes it trivial to link someone to the right docs directly when we're giving support

lib/openstack_query/docs/PROPERTIES.md
```diff
@@ -0,0 +1,117 @@
...
+| FLAVOR_ID          | `None`  | Unique ID Openstack has assigned the flavor.                                                    |
+| FLAVOR_IS_DISABLED | `None`  | Indicates whether flavor is disabled. <br/>True if disabled, False if not                       |
+| FLAVOR_IS_PUBLIC   | `None`  | Indicates if flavor is available to all projects. <br/>True if publicly available, False if not |
+| FLAVOR_NAME        | `None`  | Name of the flavor                                                                              |
+| FLAVOR_RAM         | `None`  | The amount of RAM (in MB) this flavor offers. Type: int                                         |
+| FLAVOR_SWAP        | `None`  | Size of the swap partitions.                                                                    |
+| FLAVOR_VCPU        | `None`  | The number of virtual CPUs this flavor offers. Type: int                                        |
+
+
+## Adding a property related to an existing resource
```
DavidFair, line 86: This can stay in the original file if we split out the above sections though as it refers to the generic sense

lib/openstack_query/docs/QUERIES.md
```diff
@@ -0,0 +1,232 @@
...
+This method allows us to configure server-side key-word arguments to pass directly to Openstack API servers so that they
+can execute the query at the server-side. This allows for much quicker results than using client-side filter functions.
+See [PRESETS.md](PRESETS.md) for details
+
+
+#
+#### get_client_side_handler
+
+This method is used to define valid preset-property pairs that can be passed to `where()`.
+See PRESETS.md for details
```
DavidFair, line 56: These are abstract methods to enforce their implementation - let the ABC force future devs to do the right thing

If these descriptions are better than the existing doc strings we should move them there into the code (which you'll be directly interacting with) rather than here

I think saying, implement the abstract class X, then letting the user read the interface is acceptable and keeps the docstrings near the codebase to save the person switching between this page and the new class they're creating

DavidFair (changes requested):
Additionally, could we split apart the user and developer docs into two locations?

Since "users" writing a workflow don't need to know how to add a new resource type for example, and it's something I see many doc pages do:

E.g.
https://wiki.blender.org/wiki/Developer_Intro
https://wiki.blender.org/wiki/Python

Which maps onto us quite nicely, they have users of Python who need to develop using the API, and contributors who develop the API

lib/openstack_query/docs/QUERIES.md
meoflynn: Ahh I see, okay 👍

lib/openstack_query/docs/API.md
```diff
@@ -0,0 +1,324 @@
...
+
+#
+### run
+
+`run()` will run the query. `run()` will apply all predefined conditioned set by `where` calls
+
+
+```python
+def run(cloud_account: Union[str, CloudDomains],
+        from_subset: Optional[List[OpenstackResourceObj]] = None,
```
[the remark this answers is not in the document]
anish-mudaraddi: I want to remove it - I'm going leaving it in just in case - I'll put a note in the docs

lib/openstack_query/docs/API.md
```diff
@@ -0,0 +1,324 @@
...
+
+`where()` requires a preset-property pair to work.
+- A preset is a special enum that defines the logic to query by
+- A property is what the preset will be used on
+- (optional) A set of key-word arguments that the preset-property pair require - like a value to compare against
+
+This can be called multiple times to define multiple conditions for the query - acts as Logical AND.
+
+```python
+def where(preset: QueryPresets, prop: PropEnum, **kwargs)
```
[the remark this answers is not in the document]
DavidFair: Sorry I meant put the signature then the example below
I wrote this comment before the other one

lib/openstack_query/docs/CHAINING.md
```diff
@@ -0,0 +1,136 @@
...
+## Reference
+This section shows what queries can be chained together. Applies for both `then()` and `append_from()`
+
+`ServerQuery` available chains
+
+| Prop 1                      | Prop 2                       | Type        | Maps                            |
+|-----------------------------|------------------------------|-------------|---------------------------------|
+| ServerProperties.USER_ID    | UserProperties.USER_ID       | Many-to-One | `ServerQuery` to `UserQuery`    |
+| ServerProperties.PROJECT_ID | ProjectProperties.PROJECT_ID | Many-to-One | `ServerQuery` to `ProjectQuery` |
+| ServerProperties.FLAVOR_ID  | FlavorProperties.FLAVOR_ID   | Many-to-One | `ServerQuery` to `FlavorQuery`  |
```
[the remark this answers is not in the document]
anish-mudaraddi: I get that - I'll delete it - but if someone asks for it - you can write it next time haha

lib/openstack_query/docs/FILTER_MAPPINGS.md
```diff
@@ -0,0 +1,197 @@
...
+| EQUAL_TO | ServerProperties.USER_ID                                             | `{"user_id": <value-given>}`       |
+| EQUAL_TO | ServerProperties.SERVER_ID                                           | `{"uuid": <value-given>}`          |
+| EQUAL_TO | ServerProperties.SERVER_NAME                                         | `{"hostname": <value-given>}`      |
+| EQUAL_TO | ServerProperties.SERVER_DESCRIPTION                                  | `{"description": <value-given>}`   |
+| EQUAL_TO | ServerProperties.SERVER_STATUS                                       | `{"status": <value-given>}`        |
+| EQUAL_TO | ServerProperties.SERVER_CREATION_DATE                                | `{"created_at": <value-given>}`    |
+| EQUAL_TO | ServerProperties.FLAVOR_ID                                           | `{"flavor": <value-given>}`        |
+| EQUAL_TO | ServerProperties.IMAGE_ID                                            | `{"image": <value-given>}`         |
+| EQUAL_TO | ServerProperties.PROJECT_ID                                          | `{"project_id": <value-given>}`    |
```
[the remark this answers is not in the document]
anish-mudaraddi: I'll delete it

lib/openstack_query/docs/FILTER_MAPPINGS.md
```diff
@@ -0,0 +1,197 @@
...
+| EQUAL_TO | all other ServerProperties <br/>(see [PROPERTIES.md](PROPERTIES.md)) | None                               |
+
+
+### `QueryPresetsGeneric.ANY_IN` preset:
+
+Same as `EQUAL_TO` - since it `ANY_IN` just runs multiple `EQUAL_TO` queries
```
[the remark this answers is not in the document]
anish-mudaraddi: if you want me to delete FILTER_MAPPINGS - then this can be ignored?

lib/openstack_query/docs/FILTER_MAPPINGS.md
```diff
@@ -0,0 +1,197 @@
...
+
+
+### `QueryPresetsDateTime` presets:
+
+*see `/openstack_query/time_utils.py` - for `convert-to-timestamp` method details*
+
+| Preset                   | Property(ies)                             | server-side key-word args (if any)                                        |
+|--------------------------|-------------------------------------------|---------------------------------------------------------------------------|
+| YOUNGER_THAN_OR_EQUAL_TO | ServerProperties.SERVER_LAST_UPDATED_DATE | `{"changes-since": convert-to-timestamp(days, hours, minutes, seconds)}`  |
+| YOUNGER_THAN_OR_EQUAL_TO | ServerProperties.SERVER_CREATION_DATE     | None                                                                      |
```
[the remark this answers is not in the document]
anish-mudaraddi: UTC - I'll make a note

anish-mudaraddi:
I'm closing this the changes @DavidFair wanted me to make has changed this quite a lot! 
I'm going to make a new PR for the updated docs

Issue #88, opened 2023-07-27
Query Library: Simplify using the underlying Query Mechanism


anish-mudaraddi:
currently using the underlying query mechanism requires using enums as input for Properties and Query Presets. 

This is overly complicated - we should include the option of accepting strings and handle the conversion to enums within the function rather than requesting enums as arguments 

Makes it easier for users to use the Query Library once it's a standalone package

instead of doing:
```
from openstack_query.queries.server_query import ServerQuery
from openstack_query.enums.cloud_domains import CloudDomains
from openstack_query.enums.props.server_properties import ServerProperties

# create query object
query_obj = ServerQuery(CloudDomains.from_string("prod"))

# set what to output
to_select = [ServerProperties.from_string(prop) for prop in ["server_id", "server_name", "server_status"]]
query_obj.select(*to_select)

# set query preset to use
query_obj.where(QueryPresets.from_string("any_in"), ServerProperties.from_string("server_status"), ["error", "shutoff"])

# run the query
query_obj.run()

# get output
print(query_obj.to_str())
```

we should allow:
```
from openstack_query.enums.props.server_properties import ServerProperties

# create query object
query_obj = ServerQuery("prod"))

# set what to output
query_obj.select("server_id", "server_name", "server_status")

# set query preset to use
query_obj.where(preset="any_in", property="status", values=["error", "shutoff"]]))

# run the query
query_obj.run()

# get output
print(query_obj.to_str())
```

anish-mudaraddi:
basically done

PR #193, opened 2023-11-30
ENH: add ability to enter string aliases instead of enums

anish-mudaraddi:
various methods for query api now accept string aliases as well as Enums - see #88 
Added Alias support and reworked how aliases can be added

### DIFF
--- a/lib/enums/query/enum_with_aliases.py
+++ b/lib/enums/query/enum_with_aliases.py
@@ -27,7 +27,9 @@ class EnumWithAliases(Enum):
             valid_aliases = cls._get_aliases().get(prop_enum, [])
 
             # if val matches prop_enum.name, then directly return without looking at mappings
-            if val.upper() == prop_enum.name or val.lower() in valid_aliases:
+            if val.casefold() == prop_enum.name.casefold() or any(
+                val.casefold() == i.casefold() for i in valid_aliases
+            ):
                 return prop_enum
         raise ParseQueryError(
             f"Could not parse string alias {val}. "
--- a/lib/enums/query/enum_with_aliases.py
+++ b/lib/enums/query/enum_with_aliases.py
@@ -1,0 +1,36 @@
+from abc import abstractmethod
+from enum import Enum
+from typing import Dict
+
+from exceptions.parse_query_error import ParseQueryError
+
+
+class EnumWithAliases(Enum):
+    """
+    Enum class which stores aliases which can be mapped onto stored enums
+    """
+
+    @staticmethod
+    @abstractmethod
+    def _get_aliases() -> Dict:
+        """
+        A method that returns all valid string aliases for a given property enum
+        """
+
+    @classmethod
+    def from_string(cls, val: str):
+        """
+        A class method that converts string alias to enum
+        :param val: string value that is to be converted to an enum
+        """
+        for prop_enum in cls:
+            valid_aliases = cls._get_aliases().get(prop_enum, [])
+
+            # if val matches prop_enum.name, then directly return without looking at mappings
+            if val.upper() == prop_enum.name or val.lower() in valid_aliases:
+                return prop_enum
+        else:
+            raise ParseQueryError(
+                f"Could not parse string alias {val}. "
+                f"The string alias not valid for any of the following Enums {','.join([prop.name for prop in cls])}"
+            )
--- a/lib/enums/query/enum_with_aliases.py
+++ b/lib/enums/query/enum_with_aliases.py
@@ -29,8 +29,7 @@ class EnumWithAliases(Enum):
             # if val matches prop_enum.name, then directly return without looking at mappings
             if val.upper() == prop_enum.name or val.lower() in valid_aliases:
                 return prop_enum
-        else:
-            raise ParseQueryError(
-                f"Could not parse string alias {val}. "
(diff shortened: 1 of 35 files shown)


anish-mudaraddi:
This should be one of the final changes to the Query Library API - now you can use string instead of presets - I'll need to edit the docs first

lib/enums/query/query_presets.py
```diff
@@ -1,9 +1,14 @@
...
 
+# pylint: disable=too-few-public-methods
 
-class QueryPresets(Enum):
-    pass
+
+class QueryPresets(EnumWithAliases):
+    @staticmethod
+    def _get_aliases():
+        pass
```
DavidFair, line 11: Do we need this? I would argue that this should just be a typing alias: https://stackoverflow.com/questions/33045222/how-do-you-alias-a-type-in-python

```
QueryPresets = EnumWithAliases
```

lib/enums/query/query_types.py
```diff
@@ -19,14 +22,25 @@ class QueryTypes(Enum):
...
+                "query_projects",
+                "to_project_query",
+            ],
+            QueryTypes.SERVER_QUERY: [
+                "server",
+                "servers",
+                "query_servers",
+                "to_server_query",
+            ],
+            QueryTypes.USER_QUERY: ["user", "users", "query_users", "to_user_query"],
```
DavidFair, line 45: I don't think we should accept aliases for query objects, as the API will bloat rapidly with every combination

Anyone using an IDE will type `pro` and get every variation of project for example

lib/openstack_query/docs/user_docs/API.md
```diff
@@ -48,6 +48,9 @@ from enums.query.props.server_properties import ServerProperties
 # create a ServerQuery and find server names and ids
 query = ServerQuery()
 query.select(ServerProperties.SERVER_NAME, ServerProperties.SERVER_ID)
+
+# or using the equivalent string aliases
+query.select("server_name", "server_id")
```
DavidFair, line 53: Too late now, but IMO this should be a list since every other API isn't implicitly converted to a list and it's not very Pythonic since I read it as two parameters

DavidFair:
Also we need to be careful we only add aliases as we need them, not proactively since every single one cannot be taken away

Plural makes sense, as does `disk` or `disk_size` but `vcpu`, `vcpus` and `vcpu_count` starts getting a bit redundant for example

lib/enums/query/query_types.py
anish-mudaraddi: this is not for query objects - its specifically for when calling `then()` or `append_from()`. Importing queries stays the same

lib/openstack_query/docs/user_docs/API.md
anish-mudaraddi: it is two parameters though? I don't see what's wrong here

lib/enums/query/query_presets.py
anish-mudaraddi: EnumWithAliases is a base class which contains the class method to_string()

lib/openstack_query/docs/user_docs/API.md
anish-mudaraddi: I'll do this in a separate PR

lib/enums/query/enum_with_aliases.py
```diff
@@ -0,0 +1,35 @@
...
+    def from_string(cls, val: str):
+        """
+        A class method that converts string alias to enum
+        :param val: string value that is to be converted to an enum
+        """
+        for prop_enum in cls:
+            valid_aliases = cls._get_aliases().get(prop_enum, [])
+
+            # if val matches prop_enum.name, then directly return without looking at mappings
+            if val.upper() == prop_enum.name or val.lower() in valid_aliases:
```
DavidFair, line 30: Nit `casefold` is the correct way to do this:
https://docs.python.org/3/library/stdtypes.html#str.casefold

```
if val.casefold() == prop_enum.name.casefold() or any(val.casefold() == i.casefold() for i in valid_aliases))

lib/enums/query/query_presets.py
DavidFair, line 11: We should drop this, and also make this an abstract class

Currently if I made a new class based on `QueryPresets` fails to implement this method it won't error until runtime, whereas dropping this means we couldn't instantiate the new class

lib/enums/query/query_types.py
DavidFair: But why do we need `then("query_servers", SOME_PROP, ...)` rather than just `then("servers")`

I agree on the plural but I don't think anybody has done `query_servers` in a then or append from for example?

DavidFair (changes requested):
Some minor nits, otherwise a nice cleanup

anish-mudaraddi:
>> But why do we need then("query_servers", SOME_PROP, ...) rather than just then("servers")

Um.. `then("servers")` is how it's called. Maybe you are thinking of `append_from` 
      - which is called like so `append_from("servers", "cloud_account", "name", "uuid") 
      (this should really be `append_from("servers", "cloud_account", ["name", "uuid"]) - which I'll do in a separate PR

I can remove `query_servers` and `to_server_query`

DavidFair (approved):

meoflynn (approved):

Issue #108, opened 2023-08-25
Query Library: protected/private attributes should be public if they are modified by tests


anish-mudaraddi:
there are many attributes all over the codebase that are used in tests  - tests count as a consumers of the class so these attributes should be public

https://softwareengineering.stackexchange.com/a/380296

anish-mudaraddi:
done

PR #195, opened 2023-12-01
TST: rewriting QueryBuilder Tests

anish-mudaraddi:
I rewrote these tests so that they aren't accessing protected methods, plus I think the tests are neater
see #108

### DIFF
--- a/tests/lib/openstack_query/query_blocks/test_query_builder.py
+++ b/tests/lib/openstack_query/query_blocks/test_query_builder.py
@@ -27,15 +27,22 @@ def mock_valid_client_handler_fixture():
     Return a client handler - mocks one handler which does contain preset
     """
 
+    def _mock_preset_known(preset):
+        """A dummy function to mock the preset_known method on the client handler"""
+        return (
+            True
+            if preset in [MockQueryPresets.ITEM_1, MockQueryPresets.ITEM_3]
+            else False
+        )
+
     def _mock_check_supported(preset, prop):
         """A dummy function to mock the check_supported method on the client handler"""
-        if preset == MockQueryPresets.ITEM_1:
-            if prop == MockProperties.PROP_1:
-                return True
+        if preset == MockQueryPresets.ITEM_1 and prop == MockProperties.PROP_1:
+            return True
         return False
 
     mock_invalid_client_handler = MagicMock()
-    mock_invalid_client_handler.preset_known.return_value = True
+    mock_invalid_client_handler.preset_known = MagicMock(wraps=_mock_preset_known)
     mock_invalid_client_handler.check_supported = MagicMock(wraps=_mock_check_supported)
     return mock_invalid_client_handler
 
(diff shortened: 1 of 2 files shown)


tests/lib/openstack_query/query_blocks/test_query_builder.py
```diff
@@ -10,72 +10,107 @@
...
+        client_side_handlers=client_handlers,
         server_side_handler=mock_server_side_handler,
     )
-    return instance
 
 
-@pytest.fixture(name="run_parse_where_test_case")
-def run_parse_where_test_case_fixture(instance, mock_server_side_handler):
+@pytest.fixture(name="run_parse_where_test")
+def run_parse_where_test_fixture(
```
DavidFair, line 71: Nit: this fixture has the name test too

IMO this should be something like `parse_where_runner` to make it clear it's a test helper or runner rather than a test that's run directly

DavidFair: Also any reason this is a fixture, rather than just a free function?

tests/lib/openstack_query/query_blocks/test_query_builder.py
```diff
@@ -257,39 +204,139 @@ def test_add_filter_no_server_side_filter(instance):
...
+@pytest.mark.parametrize(
+    "existing_filters, new_filters",
+    [
+        # one server-side filter set
+        # same filter to add again - conflicting
+        (
+            [{"filter1": "val1"}],
+            [{"filter1": "val2"}],
+        ),
+        # multiple server-side filters set
```
DavidFair, line 239: These should be split into test cases and the common assertion / setup code moved into a method

That way if it fails, it comes up with the test name i.e.
`test_add_filter_conflicting_resets_multiple_server_side_filters failed`

Whereas at the moment it will just say this failed and it's not obvious which case would be broken

DavidFair (changes requested):

tests/lib/openstack_query/query_blocks/test_query_builder.py
anish-mudaraddi: `run_parse_where_test_fixture` uses fixtures so I assumed it had to also be a fixture?
I'm leaving it like it is for now - but I'm renaming the fixture to `parse_where_runner`

tests/lib/openstack_query/query_blocks/test_query_builder.py
anish-mudaraddi: :sob:  noo! But DRY!!
but fair enough - I will do this

tests/lib/openstack_query/query_blocks/test_query_builder.py
DavidFair: Ah genuine apologies I couldn't see the decorator for the fixture on my review :man_facepalming: I'll take this either way since it's my mistake

DavidFair (dismissed):

khalford (commented):
LGTM - Just doc changes.

PR #203, opened 2024-01-05
MAINT: refactor Runner and Executer 

anish-mudaraddi:
SImplify Runner class to just be concerned with how to interact with openstacksdk. Move the other methods concerned with actually running the queries out into Executer

changes include: 
Executer has two more functions `run_with_subset` and `run_with_openstacksdk`.
`RunnerWrapper` is much simpler now
Created a `RunnerUtils` for holding static methods that are shared between different Runner classes

### DIFF
--- a/lib/openstack_query/api/query_api.py
+++ b/lib/openstack_query/api/query_api.py
@@ -103,7 +103,7 @@ class QueryAPI:
 
     def run(
         self,
-        cloud_account: Union[str, CloudDomains],
+        cloud_account: Union[str, CloudDomains] = None,
         from_subset: Optional[List[OpenstackResourceObj]] = None,
         **kwargs,
     ):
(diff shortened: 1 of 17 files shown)


lib/openstack_query/api/query_api.py
```diff
@@ -114,24 +114,27 @@ def run(
...
         """
+        if not cloud_account and not from_subset:
+            raise ParseQueryError(
+                "please provide as a parameter, one of:"
+                "\n\tcloud_account - a cloud domain to run query using openstacksdk"
+                "\n\tfrom_subset - a set of openstack objects"
+            )
 
         if from_subset:
-            logger.debug(
```
meoflynn, line 119: Why has the `logger.debug` been removed here?

meoflynn (changes requested):

lib/openstack_query/api/query_api.py
anish-mudaraddi: I moved it to the Executer Class - which is what we call. 
https://github.com/stfc/st2-cloud-pack/pull/203/files/41436e1d1d9d6058bc666f997e801f119d177821#diff-d8bf356dba2675d5eeb0704114eb88370e704d9be0109a037980d34eb1b44eb6R111

Koldstream (approved):
Looked good to me

meoflynn (approved):

PR #208, opened 2024-01-09
ENH: use email wrapper and styling

anish-mudaraddi:
Add styling to make html emails look better. 
Styles can be defined in wrapper.html with style tags

We use css-inline to automatically convert style tags to html inline styles which is required when styling html emails

### DIFF
--- a/lib/email_api/emailer.py
+++ b/lib/email_api/emailer.py
@@ -16,6 +16,8 @@ from structs.email.email_params import EmailParams
 from structs.email.smtp_account import SMTPAccount
 from structs.email.email_template_details import EmailTemplateDetails
 
+import css_inline
+
 # pylint:disable=too-few-public-methods
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


Koldstream (dismissed):

lib/email_api/emailer.py
```diff
@@ -16,6 +16,8 @@
 from structs.email.smtp_account import SMTPAccount
 from structs.email.email_template_details import EmailTemplateDetails
 
+import css_inline
```
khalford, line 19: As far as I can see, you only use one method from `css_inline`:
```suggestion
from css_inline import CSSInliner
```

lib/email_api/emailer.py
```diff
@@ -114,9 +117,20 @@ def build_email_body(self, templates: List[EmailTemplateDetails], as_html):
...
+            wrapper_template = EmailTemplateDetails(
+                template_name="wrapper", template_params={"body": msg_body}
+            )
+            html_body = self._template_handler.render_html_template(wrapper_template)
+
+            # convert style tags to inline styles for emails
+
+            # weird issue where pylint can't find the module - works fine though
+            # pylint:disable=no-member
+            inliner = css_inline.CSSInliner(keep_style_tags=True)
```
khalford, line 131: If you change the import:
```suggestion
            inliner = CSSInliner(keep_style_tags=True)
```

khalford (changes requested):

khalford (approved):

khalford (approved):

Koldstream (approved):

PR #209, opened 2024-01-10
MAINT: add extra kwarg to style output

anish-mudaraddi:
include_group_titles configures if we should print group titles when converting to string/html


khalford (approved):
LGTM

Koldstream (approved):

PR #210, opened 2024-01-10
ENH: add send decom image action

anish-mudaraddi:
Adds the send decom image script and action

### DIFF
--- a/lib/workflows/send_decom_image_email.py
+++ b/lib/workflows/send_decom_image_email.py
@@ -1,0 +1,282 @@
+from typing import List, Dict, Optional, Union
+import re
+from datetime import datetime
+from tabulate import tabulate
+from openstack_query import ImageQuery, UserQuery
+
+from enums.cloud_domains import CloudDomains
+from structs.email.smtp_account import SMTPAccount
+from structs.email.email_template_details import EmailTemplateDetails
+from structs.email.email_params import EmailParams
+
+from email_api.emailer import Emailer
+
+
+def get_affected_images_html(image_info_list: List[Dict]):
+    """
+    returns information on images that are to be decommissioned as a html table
+    :param image_info_list: A list of dictionaries containing information on images that are to be decommissioned
+    """
+
+    table_html = "<table>"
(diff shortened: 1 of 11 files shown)


khalford (dismissed):
LGTM

lib/workflows/send_decom_image_email.py
```diff
@@ -0,0 +1,282 @@
...
+    ), "please provide a list of image names to decommission"
+
+    assert len(image_name_list) == len(
+        image_eol_list
+    ), "Lists un-equal: Each image to decommission requires an EOL date"
+
+    assert len(image_name_list) == len(image_upgrade_list), (
+        "Lists un-equal: Each image to decommission requires a recommended upgraded image to replace it, "
+        "leave blank '' to specify if no upgrade image exists (and deletion is recommended)"
+    )
```
DavidFair, line 65: These should be `if` statements, asserts are for programming errors which shouldn't happen

Instead this should be either a `ValueError` or something similar

lib/workflows/send_decom_image_email.py
```diff
@@ -0,0 +1,282 @@
...
+
+    img_list = []
+    for image_name, eol, upgrade_image in zip(
+        image_name_list, image_eol_list, image_upgrade_list
+    ):
+        assert datetime.strptime(
+            eol, "%Y/%m/%d"
+        ), "End Of Life string must be in format YYYY/MM/DD"
```
DavidFair, line 73: ^

lib/workflows/send_decom_image_email.py
```diff
@@ -0,0 +1,282 @@
...
+
+        else:
+            email_params = build_email_params(
+                user_name,
+                get_affected_images_plaintext(decom_image_info)
+                if not as_html
+                else get_affected_images_html(decom_image_info),
+                server_query.to_string(groups=[user_id])
+                if not as_html
+                else server_query.to_html(groups=[user_id], include_group_titles=False),
```
DavidFair, line 276: Nit: can we do these if-else outside of this method call as it's a bit difficult to read

DavidFair (changes requested):

DavidFair (approved):

khalford (approved):

PR #211, opened 2024-01-15
BUG: error triggered when query returns no results

anish-mudaraddi:
when chaining, if the chaining results in no values found, then this error is triggered - instead we should just return nothing - as there's no values to append previous query values to. 
If this is unexpected, its up to user scripts to handle the error


khalford (dismissed):
LGTM

DavidFair (approved):

Issue #215, opened 2024-03-15
Query Library: Circular Imports from QueryTypes and QueryChainer


anish-mudaraddi:
Chaining module QueryChainer has a circular import issue with QueryTypes and QueryAPI. 

We should change the way that these fit together.

Currently Circular import is detailed here
![image](https://github.com/stfc/st2-cloud-pack/assets/81758831/d9740a81-81fe-4bc9-bb7e-b4fde18565b5)

trying to resolve imports leads to issue where `QueryChainer` needs access to `QueryTypes`

anish-mudaraddi:
I think this is fixed

PR #216, opened 2024-03-15
BUG: circular import breaking tests

anish-mudaraddi:
Fix a circular import breaking tests by dynamically importing modules. 
This needs a more concrete fix which I will create an issue for #215


khalford (approved):
LGTM - Not the most ideal way to import. However, if it works it works

khalford (approved):

jacob-ward (approved):

PR #217, opened 2024-03-19
MAINT: refactor QueryParser

anish-mudaraddi:
split up QueryParser into QueryGrouper and QuerySorter. 
simplifies QueryParser class which was growing unwieldy

Also removes need for protected-access in tests

### DIFF
--- a/lib/openstack_query/query_blocks/query_grouper.py
+++ b/lib/openstack_query/query_blocks/query_grouper.py
@@ -1,0 +1,127 @@
+from typing import List, Dict, Optional
+import logging
+from collections import OrderedDict
+from openstack_query.query_blocks.result import Result
+from enums.query.props.prop_enum import PropEnum
+from custom_types.openstack_query.aliases import GroupMappings, GroupRanges, PropValue
+from exceptions.parse_query_error import ParseQueryError
+
+logger = logging.getLogger(__name__)
+
+
+class QueryGrouper:
+    """
+    Helper class for implementing grouping on query outputs
+    """
+
+    def __init__(self, prop_enum_cls):
+        self._prop_enum_cls = prop_enum_cls
+        self._group_by = None
+        self._group_mappings = {}
+
+    def _build_unique_val_groups(
+        self,
+        obj_list: List,
+    ) -> GroupMappings:
+        """
+        helper method to find all unique values for a given property in query results, and then,
+        for each unique value, create a group mapping
+        :param obj_list: A list of openstack objects to build unique groups for
+        """
+        prop_func = self._prop_enum_cls.get_prop_mapping(self._group_by)
+        # ordered dict to mimic ordered set
+        # this is to preserve order we see unique values in - in case a sort has been done already
+        unique_vals = OrderedDict({prop_func(obj): None for obj in obj_list})
+        logger.debug(
+            "unique values found %s - each will become a group",
+            ",".join(f"{val}" for val in unique_vals),
+        )
+        group_mappings = {}
+        # build groups
+        for val in unique_vals.keys():
+            group_key = val
+            group_mappings[group_key] = (
+                lambda obj, test_val=val: prop_func(obj) == test_val
+            )
+        return group_mappings
+
+    def _add_include_missing_group(self, group_ranges: GroupRanges):
+        """
+        Helper method used to add an extra group for group_by which includes
+        all resource values not already included in any defined mappings
+        :param group_ranges: A set of predefined group ranges to check against
+        """
+        # if ungrouped group wanted - filter for all not in any range specified
+        all_prop_list = set()
+        for _, prop_list in group_ranges.items():
+            all_prop_list.update(set(prop_list))
+
+        prop_func = self._prop_enum_cls.get_prop_mapping(self._group_by)
+        logger.debug("creating filter function for ungrouped group")
+        self._group_mappings["ungrouped results"] = (
+            lambda obj: prop_func(obj) not in all_prop_list
+        )
+
+    def _parse_group_ranges(self, group_ranges: Optional[Dict[str, List[PropValue]]]):
+        """
+        helper method for parsing group ranges
+        :param group_ranges: a dictionary containing names of the group and list of prop values
+        to select for that group
+        """
+        logger.debug("creating filter functions for specified group ranges")
+        prop_func = self._prop_enum_cls.get_prop_mapping(self._group_by)
+        for name, prop_list in group_ranges.items():
+            group_vals = tuple(prop_list)
+            self._group_mappings[name] = (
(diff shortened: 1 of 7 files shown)


lib/openstack_query/query_blocks/query_grouper.py
```diff
@@ -0,0 +1,137 @@
...
+    def _add_include_missing_group(self, group_ranges: GroupRanges):
+        """
+        Helper method used to add an extra group for group_by which includes
+        all resource values not already included in any defined mappings
+        :param group_ranges: A set of predefined group ranges to check against
+        """
+        # if ungrouped group wanted - filter for all not in any range specified
+        all_prop_list = set()
+        for _, prop_list in group_ranges.items():
+            all_prop_list.update(set(prop_list))
```
khalford, line 57: Can we use .values() here instead of .items(). I see you are using `_` as an ignored parameter.
```suggestion
        for prop_list in group_ranges.values():
            all_prop_list.update(set(prop_list))
```

lib/openstack_query/query_blocks/query_grouper.py
```diff
@@ -0,0 +1,137 @@
...
+        method apply a set of group mappings onto a list of openstack objects. Returns a dictionary of grouped
+        values where the key is the group name and value is a list of result objects that belong to that group
+        :param obj_list: a list of Result objects containing query results to group by
+        """
+
+        # if group mappings not specified - make a group for each unique value found for prop
+        if not self._group_mappings:
+            logger.info(
+                "no group ranges specified - grouping by unique values of %s property",
+                self._group_by.name,
```
khalford, line 90: I'm curious to know why you seem to prefer using `%s` substitution rather than f strings.

lib/openstack_query/query_blocks/query_grouper.py
```diff
@@ -0,0 +1,137 @@
...
+            )
+            self._group_mappings = self._build_unique_val_groups(
+                [item.as_object() for item in obj_list]
+            )
+
+        return {
+            name: [item for item in obj_list if map_func(item.as_object())]
+            for name, map_func in self._group_mappings.items()
```
khalford, line 98: I think 2 comprehensions in one statement is a bit complex. Is it possible to expand this to 2 statements each with a single comprehension?

lib/openstack_query/query_blocks/query_grouper.py
```diff
@@ -0,0 +1,137 @@
...
+        if isinstance(prop, str):
+            prop = self._prop_enum_cls.from_string(prop)
+        return prop
+
+    def parse_group_by(
+        self,
+        group_by: Union[str, PropEnum],
+        group_ranges: Optional[GroupRanges] = None,
+        include_missing: Optional[bool] = False,
+    ) -> None:
```
khalford, line 115: Removing the return type `None` here would be more consistent with your other functions.
```suggestion
    ):
```

lib/openstack_query/query_blocks/query_sorter.py
```diff
@@ -0,0 +1,88 @@
...
+    def run_sort_by(self, obj_list: List[Result]) -> List[Result]:
+        """
+        method which sorts a list of query results based on a dictionary of sort_by specs
+        :param obj_list: a list of Result objects containing query results to sort
+        """
+
+        logger.debug("running multi-sort")
+        sort_num = len(self._sort_by)
+        for i, (sort_key, reverse) in enumerate(
+            reversed(tuple(self._sort_by.items())), 1
```
khalford, line 80: I think having a variable change by three functions inside an if condition is a bit much.
```suggestion
        obj = enumerate(reversed(tuple(self._sort_by.items()))
        for i, (sort_key, reverse) in obj , 1
```

tests/lib/openstack_query/query_blocks/conftest.py
```diff
@@ -0,0 +1,41 @@
+from unittest.mock import MagicMock
+import pytest
```
khalford, line 2: I think you can change this to import only fixture. Fixture is quite obvious about where it comes from since we only see it in a test file.
```suggestion
from pytest import fixture
```

tests/lib/openstack_query/query_blocks/test_query_grouper.py
```diff
@@ -0,0 +1,164 @@
...
+from enums.query.props.server_properties import ServerProperties
+
+from exceptions.parse_query_error import ParseQueryError
+
+
+from tests.lib.openstack_query.mocks.mocked_props import MockProperties
+
+
+@pytest.fixture(name="instance")
+def instance_fixture():
```
khalford, line 14: Should this fixture have a scope?

tests/lib/openstack_query/query_blocks/test_query_grouper.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+    Returns an instance with mocked prop_enum_cls inject
+    """
+    mock_prop_enum_cls = MockProperties
+    return QueryGrouper(prop_enum_cls=mock_prop_enum_cls)
+
+
+@pytest.fixture(name="run_group_by_runner")
+def run_group_by_runner_fixture(instance, mock_get_prop_mapping):
```
khalford, line 23: Scope?

khalford (changes requested):
Looks good, a few comments but nothing major.

tests/lib/openstack_query/query_blocks/test_query_grouper.py
anish-mudaraddi: The scope is function - which is the default I believe - so I don't specify it

tests/lib/openstack_query/query_blocks/test_query_grouper.py
anish-mudaraddi: see above

tests/lib/openstack_query/query_blocks/conftest.py
anish-mudaraddi: I'd rather not - I think `pytest.fixture` is more readable. Look up namespace pollution - https://stackoverflow.com/questions/22903542/what-is-namespace-pollution

where you draw the line is up to you, but "fixture" in my mind is ambiguous - and is more explicit when you do pytest.fixture in the code

lib/openstack_query/query_blocks/query_grouper.py
anish-mudaraddi: fixes this linting error
```lib/openstack_query/query_blocks/query_grouper.py:88:12: W1203: Use lazy % formatting in logging functions (logging-fstring-interpolation)```

khalford (approved):

meoflynn (approved):

PR #219, opened 2024-03-27
BUG: Routers not created in HA Mode

DavidFair:
### Description:

Routers for new projects have not been created in HA Mode as the default was False. Remove the handling since there isn't really any reason not to deploy HA routers and it's a potential thing to go wrong

### Special Notes:

This removes a param from the .yaml, so we'll need to explicitly install and restart St2 to detect it

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/router.create.yaml
+++ b/actions/router.create.yaml
@@ -36,8 +36,4 @@ parameters:
     description: The distributed state of the router, which is distributed True or not False
     type: boolean
     default: False
-  is_ha:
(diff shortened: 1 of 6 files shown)


anish-mudaraddi (approved):
Tested and it works

meoflynn (approved):

meoflynn:
I haven't tested the changes yet but LGTM

PR #220, opened 2024-04-18
MAINT: rewrite results container tests

anish-mudaraddi:
rewrite ResultsContainer class and tests so we don't need to access protected methods for tests

### DIFF
--- a/lib/openstack_query/query_blocks/results_container.py
+++ b/lib/openstack_query/query_blocks/results_container.py
@@ -1,4 +1,4 @@
-from typing import Union, List, Dict, Callable
+from typing import Union, List, Dict, Callable, Optional
 from enums.query.props.prop_enum import PropEnum
 from openstack_query.query_blocks.result import Result
 from custom_types.openstack_query.aliases import OpenstackResourceObj, PropValue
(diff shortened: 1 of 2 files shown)


lib/openstack_query/query_blocks/results_container.py
```diff
@@ -79,13 +79,11 @@ def apply_forwarded_results(
...
             item.update_forwarded_properties(forwarded_result)
 
-    @staticmethod
     def _get_forwarded_result(
-        prop_val: str, forwarded_results: Dict[str, List]
-    ) -> Dict:
+        self, prop_val: str, forwarded_results: Dict[str, List]
+    ) -> Optional[Dict]:
         """
         static helper method to return forwarded values that match a given property value. Used for chaining
```
khalford, line 88:
```suggestion
        helper method to return forwarded values that match a given property value. Used for chaining
```

tests/lib/openstack_query/query_blocks/test_result_container.py
```diff
@@ -181,129 +189,121 @@ def mock_parse_func(results):
...
 
 
-@patch(
-    "openstack_query.query_blocks.results_container.ResultsContainer._get_forwarded_result"
-)
-def test_apply_forwarded_result(mock_get_forwarded_result, setup_instance_with_results):
+def mock_result(get_prop_return):
     """
-    Test apply_forwarded_results once results set
+    Helper function to mock a Result object with a mocked value to return when get_prop() is caled
```
khalford, line 203:
```suggestion
    Helper function to mock a Result object with a mocked value to return when get_prop() is called
```

khalford (changes requested):
LGTM only doc fixes

AspectVI (approved):

khalford (approved):

PR #232, opened 2024-06-27
MAINT: update append_from function signature

anish-mudaraddi:
previously append_from would work like this:
```
query.append_from("project", cloud_account="prod", "project_id", "project_name")
```

now it works like this
```
query.append_from("project", cloud_account="prod", props=["project_id", "project_name"])
```
which is more intuitive and fits python conventions


AkhilMag13 (approved):

khalford (approved):
LGTM

PR #234, opened 2024-07-15
Emails users about shutoff VMs

AkhilMag13:
### Description:
Added an action for users to get an email when they have VMs in shutoff state

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/workflows/send_shutoff_vm_email.py
+++ b/lib/workflows/send_shutoff_vm_email.py
@@ -27,7 +27,7 @@ def find_servers_with_shutoff_vms(
     )
     server_query.run(
         cloud_account,
-        as_admin=False,
+        as_admin=True,
         from_projects=from_projects if from_projects else None,
         all_projects=not from_projects,
     )
@@ -161,14 +161,14 @@ def send_shutoff_vm_email(
 
         if not send_email:
             print_email_params(send_to[0], user_name, as_html, server_list)
+            return
 
-        else:
-            email_params = build_email_params(
-                user_name,
-                server_list,
-                email_to=send_to,
-                as_html=as_html,
-                email_cc=("cloud-support@stfc.ac.uk",) if cc_cloud_support else None,
-                **email_params_kwargs,
-            )
-            Emailer(smtp_account).send_emails([email_params])
+        email_params = build_email_params(
+            user_name,
+            server_list,
+            email_to=send_to,
+            as_html=as_html,
+            email_cc=("cloud-support@stfc.ac.uk",) if cc_cloud_support else None,
+            **email_params_kwargs,
+        )
+        Emailer(smtp_account).send_emails([email_params])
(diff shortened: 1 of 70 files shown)


lib/workflows/send_shutoff_vm_email.py
```diff
@@ -0,0 +1,174 @@
...
+        else:
+            email_params = build_email_params(
+                user_name,
+                server_list,
+                email_to=send_to,
+                as_html=as_html,
+                email_cc=("cloud-support@stfc.ac.uk",) if cc_cloud_support else None,
+                **email_params_kwargs,
+            )
+            Emailer(smtp_account).send_emails([email_params])
```
anish-mudaraddi, line 174: NIT: remove unnecessary else by returning early

```suggestion
        if not send_email:
            print_email_params(send_to[0], user_name, as_html, server_list)
            return
        
        email_params = build_email_params(
            user_name,
            server_list,
            email_to=send_to,
            as_html=as_html,
            email_cc=("cloud-support@stfc.ac.uk",) if cc_cloud_support else None,
            **email_params_kwargs,
        )
        Emailer(smtp_account).send_emails([email_params])
```

lib/workflows/send_shutoff_vm_email.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+
+    # Find VMs that have been in shutoff state for more than 60 days
+    server_query = ServerQuery()
+    server_query.where(
+        QueryPresetsGeneric.ANY_IN, ServerProperties.SERVER_STATUS, values=["SHUTOFF"]
+    )
+    server_query.run(
+        cloud_account,
+        as_admin=False,
```
anish-mudaraddi, line 30: change this to `True` otherwise it won't work for all projects!
You'll need to adjust the tests too!

anish-mudaraddi (changes requested):
small changes - otherwise LGTM

lib/workflows/send_shutoff_vm_email.py
AkhilMag13: This gives an error as if there are multiple calls being made, it returns after the first call as can be seen in the test failing in [dbf8dbb](https://github.com/stfc/st2-cloud-pack/pull/234/commits/dbf8dbbf26170e73359102d52d84c80efc08c96e)

lib/workflows/send_shutoff_vm_email.py
AkhilMag13: Thank you, fixed this in [dbf8dbb](https://github.com/stfc/st2-cloud-pack/pull/234/commits/dbf8dbbf26170e73359102d52d84c80efc08c96e)

lib/workflows/send_shutoff_vm_email.py
anish-mudaraddi: change it to `continue` - forgot we were in a loop

AkhilMag13: Thanks, fixed it with [c1a2837](https://github.com/stfc/st2-cloud-pack/pull/234/commits/c1a28373649d9e8eaf447e15fac781414a079e52)

anish-mudaraddi (approved):

RWhiting78 (approved):

PR #237, opened 2024-07-29
BUG: fix issue with server search actions

anish-mudaraddi:
These actions were not updated to using the latest API - `where()` calls were passing search params through an argument called `args`

We removed this and just pass search params as extra arguments using `**kwargs`. 

I also added an extra user-defined parameter `all_projects` that was missing - this is required to explicitly tell the action to search all projects rather than just the current `admin` project


AkhilMag13 (approved):
Looks good

RWhiting78 (approved):

PR #238, opened 2024-07-31
Hypervisor down disabled email

AkhilMag13:
### Description:
This will send an email to whoever overrides the email field in the action on the hypervisors that are in down state or have disabled status

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:


lib/workflows/get_hvs.py
anish-mudaraddi, line 1: delete this file

anish-mudaraddi (changes requested):
you've added an empty file - but other than that its good

anish-mudaraddi (approved):

meoflynn (approved):

PR #239, opened 2024-08-01
BUG: internal project creation hotfix

anish-mudaraddi:
erroneously deleted lines to create admin/stfc roles from orquesta workflow in latest PR #223 - re-added the lines. 

This has been tested on dev and works.

We will be moving away from orquesta soon so future tests should catch issues like this!


AspectVI (approved):
Looks all good to me

DavidFair (approved):

PR #242, opened 2024-08-05
MAINT:  Refactor Openstack API - openstack_roles

anish-mudaraddi:
Part of the remove_orquesta feature branch - reimplement this class as standalone functions

These tests will fail - we're working on a feature branch so these will get fixed before we merge into main


khalford (approved):
LGTM

PR #243, opened 2024-08-05
MAINT: Openstack API Refactor - security groups

anish-mudaraddi:
part of effort to remove orquesta feature branch - we're refactoring the class into standalone functions

These tests will fail - we're working on a feature branch so these will get fixed before we merge into main


khalford (approved):
LGTM

PR #244, opened 2024-08-05
MAINT: Openstack API refactor - quota

anish-mudaraddi:
part of the remove orquesta feature branch
refactor this class to be standalone functions

Tests will fail - we're working on a feature branch so these will be fixed as we go


meoflynn (approved):

meoflynn:
Looks like one of the unit tests may be failing

PR #246, opened 2024-09-02
Add initial action for querying HV details

DavidFair:
Adds the initial action, with the lib layer stubbed, to query for empty hypervisors

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:


lib/openstack_api/openstack_hypervisor.py
```diff
@@ -0,0 +1,54 @@
...
+        """
+        hv_query = self._query_cls()
+        hv_query.where(
+            QueryPresetsGeneric.EQUAL_TO,
+            HypervisorProperties.HYPERVISOR_VCPUS_USED,
+            value=0,
+        )
+        hv_query.select(HypervisorProperties.HYPERVISOR_NAME)
+        results = hv_query.run(cloud_account=cloud_account).to_props(flatten=True)
+        return sorted(results["hypervisor_name"])
```
anish-mudaraddi, line 54:
```suggestion
        return hv_query.run(cloud_account=cloud_account).sort_by(["hypervisor_name", "ascending")).to_props(flatten=True)
        return sorted(results["hypervisor_name"])
```

lib/openstack_api/openstack_hypervisor.py
```diff
@@ -0,0 +1,54 @@
...
+            search_type = _HypervisorSearchTypes[search_type]
+        except KeyError as exc:
+            raise ValueError(f"Invalid search type: {search_type}") from exc
+
+        if search_type == _HypervisorSearchTypes.EMPTY_HVS:
+            return self._find_empty_hypervisors(cloud_account)
+
+        raise NotImplementedError(f"Search type {search_type} not implemented")
+
+    def _find_empty_hypervisors(self, cloud_account: str) -> List[str]:
```
anish-mudaraddi, line 40: prefer this to be a public method - because then we can use it in other workflows

khalford (approved):

anish-mudaraddi (approved):

khalford (approved):

PR #251, opened 2024-09-05
PROTO: project creation scripts without orquesta

anish-mudaraddi:
Convert create_external_project and create_internal_project into standalone python scripts - this will enable us to deprecate the orquesta workflows that we currently use


DavidFair (approved):

anish-mudaraddi:
tests will fail - but will be fixed in a follow up PR - note I'm merging this into a feature branch

PR #252, opened 2024-09-06
PROTO: moving useful actions into workflow folder

anish-mudaraddi:
This is so we can remove all the unnecessary code under actions.src.*. We're only keeping useful actions here - other ones can be easily done using openstack cli so we shouldn't re-invent the wheel

### DIFF
--- a/lib/workflows/project_actions.py
+++ b/lib/workflows/project_actions.py
@@ -1,0 +1,36 @@
+from typing import Tuple
+
+from openstack.connection import Connection
+from openstack_api import openstack_project
+
+
+def delete_project(
+    conn: Connection, project_identifier: str, delete: bool
+) -> Tuple[bool, str]:
+    """
+    Deletes a project
+    :param conn: Openstack connection object
+    :param project_identifier: (Either) The project name to delete
+    :param delete: When true will delete the project, when false will only return it to reduce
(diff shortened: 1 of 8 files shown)


khalford (approved):

lib/workflows/project_actions.py
```diff
@@ -0,0 +1,36 @@
...
+) -> Tuple[bool, str]:
+    """
+    Deletes a project
+    :param conn: Openstack connection object
+    :param project_identifier: (Either) The project name to delete
+    :param delete: When true will delete the project, when false will only return it to reduce
+                   chances of accidental deletion
+    :return: The result of the operation
+    """
+    project = conn.identity.find_project(
```
DavidFair, line 18: We need a project blacklist here too
The following projects and IDs should be blacklisted from ever being attempted:
```
admin
Openstack Dev Cluster
0e60669c49d4497aad91d8f729cfdd77
4de86830e89b4a46b590536571b6ccd4
c9aee696c4b54f12a645af2c951327dc
```

anish-mudaraddi: done

DavidFair (approved):

khalford (approved):

PR #254, opened 2024-09-06
Refactor entrypoints for openstack actions

anish-mudaraddi:
Actions that require access to openstack now all go through openstack_actions.py
We define an lib_entry_point for each action which tells us how to forward on requests to functions that will handle the request in the lib layer.


gmatthews20 (approved):
LGTM

DavidFair (approved):

PR #255, opened 2024-09-06
Add additional flag to filter out uncontactable HVs

DavidFair:
### Description:
This reduces the noise, especially when we're powering off HVs as we have spare capacity to hand so they're unused

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/hypervisor.find.empty.yaml
+++ b/actions/hypervisor.find.empty.yaml
@@ -16,5 +16,10 @@ parameters:
     enum:
       - "dev"
       - "prod"
+  include_offline:
+    description: Include hypervisors marked offline from OpenStack in the results
+    required: true
+    type: boolean
+    default: false
 
 runner_type: python-script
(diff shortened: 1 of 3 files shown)


khalford (approved):

anish-mudaraddi (approved):

actions/hypervisor.find.empty.yaml
```diff
@@ -16,5 +16,10 @@ parameters:
     enum:
       - "dev"
       - "prod"
+  include_offline:
```
anish-mudaraddi, line 19: why `include_offline` rather than `include_down_state_hvs` - I think its more informative

DavidFair: Following our discussion in-person, do you still want me to rename this?

`down` doesn't really work here, as it's ambiguous if the HV marked down for scheduling or is uncontactable

In this case, it's referring to HVs that are not contactable, e.g. offline, nova not installed, or even decommissioned and physically removed

I'm not sure the best way to name this that way?

anish-mudaraddi: I'm going to leave this as is - we can look at it later if its a problem

PR #258, opened 2024-09-24
BUG: use new openstack actions workflow for email.ops action

anish-mudaraddi:
this action wasn't converted over to use new workflow using lib_entry_point


gmatthews20 (approved):

lizsalmon (approved):
I've tested this and it works.

PR #259, opened 2024-09-24
ENH: add hv search action

anish-mudaraddi:
add hv.search.by.property action and fix other search actions so that they work without needing to specify group-by

### DIFF
--- a/actions/hv.search.by.property.yaml
+++ b/actions/hv.search.by.property.yaml
@@ -114,29 +114,29 @@ parameters:
     description: "a comma-spaced list of values to search for"
     required: true
     type: array
-
   group_by:
-    description: "(Optional) hypervisor property to group unique results by"
+    description: "(Optional) hypervisor property to group unique results by
+    One of
+      - 'current_workload'
+      - 'local_disk_free'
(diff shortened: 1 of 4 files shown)


gmatthews20 (approved):

actions/hv.search.by.property.yaml
```diff
@@ -0,0 +1,164 @@
+---
+description: Search for hypervisors with a selected property matching, or not matching given value(s)
+enabled: true
+entry_point: src/workflow_actions.py
```
lizsalmon, line 4: I don't think that this file exists so this won't work.

lizsalmon (changes requested):

actions/hv.search.by.property.yaml
anish-mudaraddi: yes - good shout!

anish-mudaraddi: I was trying to get this in before `remove-orquesta` major change was merged

gmatthews20 (approved):

lizsalmon (approved):

PR #262, opened 2024-09-26
ENH: Email users on a HV workflow

gmatthews20:
### Description:

Add workflow for emailing users on a given hypervisor, test temporary email template. Tests for the new workflow 

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/email_templates/html/hypervisor_vm.html.j2
+++ b/email_templates/html/hypervisor_vm.html.j2
@@ -1,9 +1,0 @@
-Dear Cloud User {{username}}, <br/><br/>
-
-Test email about servers on hv <br/><br/><br/>
-
-<b> VMs: </b><br/>
-
-<br/> {{vm_table}}
-
-Please note that this is an automated notification. If you believe you received this email in error or have any concerns, please contact Cloud Operations Group for further assistance.<br/>
--- a/email_templates/html/hypervisor_vm.html.j2
+++ b/email_templates/html/hypervisor_vm.html.j2
@@ -1,0 +1,9 @@
+Dear Cloud User {{username}}, <br/><br/>
+
+Test email about servers on hv <br/><br/><br/>
+
+<b> VMs: </b><br/>
+
+<br/> {{vm_table}}
+
+Please note that this is an automated notification. If you believe you received this email in error or have any concerns, please contact Cloud Operations Group for further assistance.<br/>
--- a/lib/email_api/email_template_schemas.yaml
+++ b/lib/email_api/email_template_schemas.yaml
@@ -47,6 +47,13 @@ shutoff_vm:
   html_filepath: "html/shutoff_vm.html.j2"
   plaintext_filepath: "plaintext/shutoff_vm.txt.j2"
 
+hypervisor_vm:
+  schema:
+    username: null
+    vm_table: "No VMs"
+  html_filepath: "html/hypervisor_vm.html.j2"
+  plaintext_filepath: "plaintext/hypervisor_vm.txt.j2"
+
 decom_flavor:
   schema:
     username: null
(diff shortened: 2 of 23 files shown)


email_templates/html/hypervisor_vm.html.j2
```diff
@@ -0,0 +1,9 @@
+Dear Cloud User {{username}}, <br/><br/>
+
+Test email about servers on hv <br/><br/><br/>
```
anish-mudaraddi, line 3: @jacob-ward @DavidFair - What should the email template be here? Worth getting this nailed down before merging this into Prod

lib/email_api/email_template_schemas.yaml
```diff
@@ -47,6 +47,13 @@ shutoff_vm:
   html_filepath: "html/shutoff_vm.html.j2"
   plaintext_filepath: "plaintext/shutoff_vm.txt.j2"
 
+hypervisor_vm:
```
anish-mudaraddi, line 50: This is fine - but maybe `vms_on_hypervisor`?
The other schemas might need changing names to be more informative as we add more

lib/enums/query/props/server_properties.py
```diff
@@ -27,7 +27,7 @@ def _get_aliases():
         A method that returns all valid string alias mappings
         """
         return {
-            ServerProperties.HYPERVISOR_ID: ["host_id", "hv_id"],
+            ServerProperties.HYPERVISOR_NAME: ["hv_name"],
```
anish-mudaraddi: I know you changed this because we discovered `host_id` wasn't reliable and didn't correlate with actual hypervisor ids -  make a comment about that here for others

anish-mudaraddi: You will also need to change the docs

lib/workflows/send_hv_email.py
```diff
@@ -0,0 +1,175 @@
...
+from enums.cloud_domains import CloudDomains
+from enums.query.props import ServerProperties
+from enums.query.query_presets import QueryPresetsGeneric
+
+from structs.email.email_params import EmailParams
+from structs.email.email_template_details import EmailTemplateDetails
+from structs.email.smtp_account import SMTPAccount
+
+
+def find_servers_on_hv(
```
anish-mudaraddi, line 15: We're doing this a lot - for all our emailing notifications - maybe we should make a new folder to hold and reuse common query patterns like this in the future

anish-mudaraddi (changes requested):

lib/enums/query/props/server_properties.py
```diff
@@ -27,7 +27,10 @@ def _get_aliases():
         A method that returns all valid string alias mappings
         """
         return {
-            ServerProperties.HYPERVISOR_NAME: ["hv_name", "hypervisor_name"],
+            ServerProperties.HYPERVISOR_NAME: [
+                "hv_name",
+                "hypervisor_name",
+            ],  # Changed from Host ID as hosts aren't the same as hypervisors
```
anish-mudaraddi, line 33: Nitpick - "we're using hypervisor name as the shared-common property because the hypervisor id was inconsistent between servers and hypervisors"

anish-mudaraddi (approved):

anish-mudaraddi (approved):

anish-mudaraddi (dismissed):

email_templates/html/hypervisor_vm.html.j2
gmatthews20: Added templates for hv fault and hv maintenance emails

meoflynn (approved):

khalford (approved):

PR #266, opened 2024-10-01
ENH: Add icinga downtime actions

gmatthews20:
Schedule and remove downtime actions
Tests for icinga api and workflows
Get icinga account from stackstorm config + example

### Description:

Schedule and remove downtime actions
Tests for icinga api and workflows
Get icinga account from stackstorm config + example

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/icinga.remove.downtime.yaml
+++ b/actions/icinga.remove.downtime.yaml
@@ -1,0 +1,29 @@
+---
+description: Remove a downtime for Host or Service in Icinga
+enabled: true
+entry_point: src/openstack_actions.py
+name: icinga.remove.downtime
+parameters:
+  name:
+    type: string
+    description: Name of host or service in Icinga
+    required: true
+  object_type:
+    immutable: false
+    type: string
+    description: Icinga Object Type
+    default: Host
+    required: true
+    enum:
+      - Host
+      - Service
+  lib_entry_point:
+    type: string
+    default: workflows.icinga_downtime.remove_downtime
+    immutable: true
+  icinga_account_name:
+    type: string
+    description: "Name of Icinga Account to use. Must be configured in in the pack settings."
+    required: true
+    default: "default"
+runner_type: python-script
--- a/actions/icinga.schedule.downtime.yaml
+++ b/actions/icinga.schedule.downtime.yaml
@@ -10,7 +10,7 @@ parameters:
     required: true
   start_time:
     type: string
-    description: "Date and time (GMT) in the format: %d/%m/%y %H:%M:%S"
+    description: "Date and time (UTC) in the format: %d/%m/%y %H:%M:%S"
     required: true
   duration:
     type: integer
--- a/actions/icinga.schedule.downtime.yaml
+++ b/actions/icinga.schedule.downtime.yaml
@@ -10,7 +10,7 @@ parameters:
     required: true
   start_time:
     type: string
-    description: "Date and time (UTC) in the format: %d/%m/%y %H:%M:%S"
+    description: "Date and time (UK Time) in the format: %d/%m/%y %H:%M:%S"
     required: true
   duration:
     type: integer
--- a/actions/icinga.schedule.downtime.yaml
+++ b/actions/icinga.schedule.downtime.yaml
@@ -10,12 +10,16 @@ parameters:
     required: true
   start_time:
     type: string
-    description: "Date and time (UK Time) in the format: %d/%m/%y %H:%M:%S"
+    description: "Start Date and time (UK Time) in the format: %d/%m/%y %H:%M:%S"
+    required: true
+  end_time:
+    type: string
+    description: "End Date and time (UK Time) in the format: %d/%m/%y %H:%M:%S"
     required: true
   duration:
     type: integer
-    description: Duration of downtime in seconds
-    required: true
+    description: Duration of flexible downtime in seconds, not needed for fixed downtime
+    required: false
   comment:
     type: string
     description: Downtime reason
--- a/actions/icinga.schedule.downtime.yaml
+++ b/actions/icinga.schedule.downtime.yaml
@@ -1,0 +1,45 @@
+---
+description: Schedule a downtime for Host or Service in Icinga
+enabled: true
+entry_point: src/openstack_actions.py
+name: icinga.schedule.downtime
+parameters:
+  name:
+    type: string
+    description: Name of host or service in Icinga
+    required: true
+  start_time:
+    type: string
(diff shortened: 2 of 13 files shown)


actions/icinga.remove.downtime.yaml
```diff
@@ -0,0 +1,29 @@
...
+name: icinga.remove.downtime
+parameters:
+  name:
+    type: string
+    description: Name of host or service in Icinga
+    required: true
+  object_type:
+    immutable: false
+    type: string
+    description: Icinga Object Type
```
anish-mudaraddi, line 14: maybe give some examples - or an enum would be handy here

actions/icinga.schedule.downtime.yaml
```diff
@@ -0,0 +1,45 @@
...
+    description: Downtime reason
+    required: true
+  is_fixed:
+    type: boolean
+    description: If true, the downtime is fixed otherwise flexible
+    default: true
+  object_type:
+    immutable: false
+    type: string
+    description: Icinga Object Type
```
anish-mudaraddi, line 30: See above

config.schema.yaml
```diff
@@ -96,3 +96,32 @@ jira_accounts:
         required: true
         secret: false
         default: "https://stfc.atlassian.net"
+
+icinga_accounts:
```
anish-mudaraddi, line 100: We'll need to look at how we're deploying stackstorm using aquilon to add this config in before this gets merged

anish-mudaraddi (changes requested):
LGTM - some small changes

actions/icinga.remove.downtime.yaml
gmatthews20: object_type is an enum, either Host or Service

anish-mudaraddi: ignore - I didn't see its already an enum :)

anish-mudaraddi (dismissed):

actions/icinga.schedule.downtime.yaml
```diff
@@ -0,0 +1,45 @@
...
+    description: Name of host or service in Icinga
+    required: true
+  start_time:
+    type: string
+    description: "Date and time (UTC) in the format: %d/%m/%y %H:%M:%S"
+    required: true
+  duration:
+    type: integer
+    description: Duration of downtime in seconds
+    required: true
```
DavidFair, line 18: We can specify the start time using DMY HMS, but then if I want to have a 1 week downtime I have to calculate days -> seconds
Any reason the end time can't also use the date too?

lib/structs/icinga/downtime_details.py
```diff
@@ -0,0 +1,12 @@
+from dataclasses import dataclass
+from typing import Optional
+
+
+@dataclass
+class DowntimeDetails:
+    object_type: str
```
DavidFair, line 7: This should be a Enum class for the discreet types we support

lib/structs/icinga/downtime_details.py
```diff
@@ -0,0 +1,12 @@
...
+    object_name: str
+    start_time: int
```
DavidFair, line 9: Use a more specific type rather than a primitive type if at all possible, in this case keep the start time as a `Date` instead of an int

lib/structs/icinga/downtime_details.py
```diff
@@ -0,0 +1,12 @@
...
+    duration: int
```
DavidFair, line 10: Ditto, with the suggestion about changing this to a date too

lib/workflows/icinga_downtime.py
```diff
@@ -0,0 +1,41 @@
...
+    object_type: str,
+    name: str,
+    start_time: str,
+    duration: int,
+    comment: str,
+    is_fixed: bool,
+):
+
+    datetime_object = datetime.strptime(start_time, "%d/%m/%y %H:%M:%S").replace(
+        tzinfo=timezone.utc
```
DavidFair, line 20: Ah we need to be careful with timezones here, this assumes UTC but during the summer people will be thinking in BST. We can either change this to BST/UTC (some sort of automatic detection if it exists in the date library) or (though it puts a burden on the user) update the description to explicitly say UTC

DavidFair (changes requested):

lib/workflows/icinga_downtime.py
gmatthews20: Changed to convert a UK time input into utc which icinga uses

actions/icinga.schedule.downtime.yaml
gmatthews20: Added end_time patameter

lib/structs/icinga/downtime_details.py
gmatthews20: start_time is a timestamp which is an int?

anish-mudaraddi: @DavidFair is asking you to use a Date obect instead of unix timestamp which you're currently using

anish-mudaraddi (dismissed):

lib/structs/icinga/downtime_details.py
gmatthews20: Icinga is expecting a timestamp which is why its there, I convert to the timestamp in the workflow but I can move the conversion to the icinga_api schedule_downtime function if makes more sense there

DavidFair: IMO yes, because it's an Icinga API implementation detail 
It also makes debugging easier because the debugger gives you the date instead of `1728051804` for example

lib/enums/icinga/icinga_objects.py
```diff
@@ -0,0 +1,8 @@
+from enum import auto
+from enums.auto_name import _AutoName
+
+
+# pylint: disable=too-few-public-methods
+class IcingaObject(_AutoName):
```
DavidFair, line 6: Your using a private implementation detail from Enum incorrectly here unfortunately

It should be like [this example](https://docs.python.org/3/library/enum.html#enum.Enum._generate_next_value_)
```
from enum import auto, Enum

class IcingaObject(Enum):
    HOST = auto()
```

DavidFair (changes requested):

DavidFair (approved):

lib/structs/icinga/downtime_details.py
anish-mudaraddi: Think we should re-visit this later if its a problem

anish-mudaraddi (approved):

PR #277, opened 2024-10-18
ENH: add image search actions

anish-mudaraddi:
written by @Neonbluestoplight

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/image.search.by.datetime.yaml
+++ b/actions/image.search.by.datetime.yaml
@@ -116,12 +116,15 @@ parameters:
     required: false
     default: null
     enum:
+      - null
       - "image_creation_date"
       - "image_creation_progress"
+      - "image_id"
       - "image_last_updated_date"
       - "image_minimum_ram"
       - "image_minimum_disk"
       - "image_name"
+      - "image_size"
       - "image_status"
   sort_by:
     description: "(Optional) comma-spaced list of image properties to sort by (by ascending only)
--- a/actions/image.search.by.datetime.yaml
+++ b/actions/image.search.by.datetime.yaml
@@ -7,7 +7,7 @@ parameters:
   timeout:
     default: 5400
   lib_entry_point:
-    default: workflows.search_by_datetime_search_by_datetime
+    default: workflows.search_by_datetime.search_by_datetime
     immutable: true
     type: string
   query_type:
--- a/actions/image.search.by.datetime.yaml
+++ b/actions/image.search.by.datetime.yaml
@@ -14,6 +14,10 @@ parameters:
     default: ImageQuery
     immutable: true
     type: string
+  as_admin:
+    default: True
+    immutable: true
+    type: boolean
   cloud_account:
     description: "The clouds.yaml account to use whilst performing this action"
     required: true
@@ -100,8 +104,8 @@ parameters:
     type: integer
     default: 0
     required: false
-  limit_by_projects:
-    description: "comma-spaced project to limit action to - incompatible with all_projects"
+  from_projects:
+    description: "(Optional) comma-spaced list of projects id/names to limit query to, if not provided, runs against all projects"
     type: array
     default: null
     required: false
@@ -143,4 +147,9 @@ parameters:
     type: array
     required: false
     default: null
+  webhook:
+    description: "Optional string representing a stackstorm webhook url path, can't be used alongside group_by"
+    default: null
+    type: string
+    required: false
 runner_type: python-script
--- a/actions/image.search.by.datetime.yaml
+++ b/actions/image.search.by.datetime.yaml
@@ -127,15 +127,15 @@ parameters:
     description: "(Optional) comma-spaced list of image properties to sort by (by ascending only)
     - multiple of the same property will be ignored.
     Any of:
-      - "image_creation_date"
-      - "image_creation_progress"
-      - "image_id"
-      - "image_last_updated_date"
-      - "image_minimum_ram"
-      - "image_minimum_disk"
-      - "image_name"
-      - "image_size"
-      - "image_status"
+      - 'image_creation_date'
+      - 'image_creation_progress'
+      - 'image_id'
+      - 'image_last_updated_date'
+      - 'image_minimum_ram'
+      - 'image_minimum_disk'
+      - 'image_name'
+      - 'image_size'
+      - 'image_status'
     Anything else will raise an error"
     type: array
     required: false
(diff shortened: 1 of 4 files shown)


actions/image.search.by.regex.yaml
```diff
@@ -0,0 +1,102 @@
...
+    description: The clouds.yaml account to use whilst performing this action
+    required: true
+    type: string
+    default: "dev"
+    enum:
+      - "dev"
+      - "prod"
+  properties_to_select:
+    default:
+      - "image_name"
```
gmatthews20, line 27: Should this default to selecting all properties like the other actions

actions/image.search.by.regex.yaml
```diff
@@ -0,0 +1,102 @@
...
+      - 'to_props' - properties dicts as a python list
+      - 'to_objects - as a list of openstack resources
+      - 'to_string' - a tabulate table"
+    required: true
+  property_to_search_by:
+    default: "image_name"
+    description: "choose property to search by"
+    enum:
+    - "image_name"
+    - "image_id"
```
gmatthews20, line 62: Don't think we need to search ID by regex

actions/image.search.by.expression.yaml
gmatthews20, line 1: can you add search all_projects and limit_by_projects parameters

actions/image.search.by.regex.yaml
gmatthews20, line 1: can you add search all_projects and limit_by_projects parameters

actions/image.search.by.property.yaml
gmatthews20, line 1: can you add search limit_by_projects parameters

gmatthews20 (changes requested):

actions/image.search.by.regex.yaml
Neonbluestoplight: Yes have changed this thank you

actions/image.search.by.expression.yaml
Neonbluestoplight: Used `for_projects` instead of `limit_by_projects` as suggested by the runner actions

actions/image.search.by.property.yaml
```diff
@@ -0,0 +1,144 @@
...
+  group_by:
+    description: "(Optional) image property to group unique results by"
+    type: string
+    required: false
+    default: null
+    enum:
+      - null
+      - "image_creation_date"
+      - "image_creation_progress"
+      - "image_id"
```
gmatthews20, line 115: I don't think grouping by image_id is needed

actions/image.search.by.datetime.yaml
```diff
@@ -0,0 +1,155 @@
...
+  from_projects:
+    description: "(Optional) comma-spaced list of projects id/names to limit query to, if not provided, runs against all projects"
+    type: array
+    default: null
+    required: false
+  all_projects:
+    type: boolean
+    description: "tick to search in all projects - default True"
+    required: true
+    default: true
```
gmatthews20, line 116: @anish-mudaraddi Is all_projects needed if when from_projects is not provided it runs against all projects?

actions/image.search.by.datetime.yaml
```diff
@@ -0,0 +1,155 @@
...
+  group_by:
+    description: "(Optional) image property to group unique results by"
+    type: string
+    required: false
+    default: null
+    enum:
+      - null
+      - "image_creation_date"
+      - "image_creation_progress"
+      - "image_id"
```
gmatthews20, line 126: I don't think grouping by image_id is needed

actions/image.search.by.expression.yaml
```diff
@@ -0,0 +1,140 @@
...
+      - "to_string"
+    description: "
+    A string representing how to return the results of the query
+      - 'to_html' - a tabulate table (in html)
+      - 'to_props' - properties dicts as a python list
+      - 'to_objects - as a list of openstack resources
+      - 'to_string' - a tabulate table"
+    required: true
+  property_to_search_by:
+    default: "image_id"
```
gmatthews20, line 45: This default won't work as it isn't in the enum, should be changed to something else, e.g. image_creation_progress
```suggestion
    default: "image_creation_progress"
```

actions/image.search.by.expression.yaml
```diff
@@ -0,0 +1,140 @@
...
+      - "image_id"
+      - "image_last_updated_date"
+      - "image_minimum_ram"
+      - "image_minimum_disk"
+      - "image_name"
+      - "image_size"
+      - "image_status"
+    type: array
+    description: "
+    A comma-spaced list of server properties to display for the resulting images - leave empty for all properties. One of:
```
gmatthews20, line 81:
```suggestion
    A comma-spaced list of image properties to display for the resulting images - leave empty for all properties. One of:
```

actions/image.search.by.regex.yaml
```diff
@@ -0,0 +1,128 @@
...
+  group_by:
+    description: "(Optional) image property to group unique results by"
+    type: string
+    required: false
+    default: null
+    enum:
+      - null
+      - "image_creation_date"
+      - "image_creation_progress"
+      - "image_id"
```
gmatthews20, line 99: I don't think grouping by image_id is needed

gmatthews20 (changes requested):

actions/image.search.by.regex.yaml
anish-mudaraddi: image id is not needed

actions/image.search.by.property.yaml
Neonbluestoplight: have removed and also remove image_name for the same reason from all of them

actions/image.search.by.expression.yaml
Neonbluestoplight: Have changed

actions/image.search.by.expression.yaml
Neonbluestoplight: Thank you for pointing that out! I tried to make sure all references were removed but I missed some

actions/image.search.by.datetime.yaml
gmatthews20: On this if from_projects and all_projects are false the it will default to searching the admin project

Might be worth changing the from_projects description to something along the lines of: `(Optional) comma-spaced list of projects id/names to limit query to, if not provided, runs against all projects when all_projects is true`

Neonbluestoplight: Have changed the description in all the image search actions now thank you

PR #280, opened 2024-11-07
MAINT: Make icinga downtime raise http errors

gmatthews20:
### Description:
Schedule and Remove downtime should raise http errors, e.g. 404 when host/service is not found

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/tests/lib/icinga_api/test_downtime.py
+++ b/tests/lib/icinga_api/test_downtime.py
@@ -2,6 +2,7 @@ import json
 from unittest.mock import MagicMock, patch
 
 import pytest
+from requests import HTTPError
 
 from exceptions.missing_mandatory_param_error import MissingMandatoryParamError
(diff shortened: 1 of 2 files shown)


lizsalmon (approved):

tests/lib/icinga_api/test_downtime.py
khalford, line 1: Are there tests to cover `test_schedule_flexible_downtime_failed`? We should test that `res.raise_for_status()` does raise an error when the mock_post returns codes other than 300. e.g. 403

gmatthews20: I have added tests to check that an HTTPError is raised when one is returned by the request

khalford (approved):

lizsalmon (approved):

PR #282, opened 2024-11-13
Hypervisor Patch and Reboot workflow

lizsalmon:
### Description:

A downtime is scheduled for the selected hypervisor and then during the downtime two ssh commands are run on the hv, patch and reboot.  These will work with the stackstorm user (currently in an aquilon sandbox) to run a patch and a reboot script.  

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/hv.patch.reboot.yaml
+++ b/actions/hv.patch.reboot.yaml
@@ -25,4 +25,12 @@ parameters:
     description: "Name of Icinga Account to use. Must be configured in in the pack settings."
     required: true
     default: "default"
+  host:
+    description: Host to run action on
+    required: true
+    type: string
+  private_key_path:
+    description: Private key to authenticate with
+    required: true
+    type: string
 runner_type: python-script
--- a/actions/hv.patch.reboot.yaml
+++ b/actions/hv.patch.reboot.yaml
@@ -1,0 +1,28 @@
+---
+description: Patch and Reboot a hypervisor
+enabled: true
+entry_point: src/openstack_actions.py
+name: hv.patch.reboot
+parameters:
+  name:
+    type: string
+    description: Name of host in Icinga
+    required: true
+  start_time:
+    type: string
+    description: "Start Date and time (UK Time) in the format: %d/%m/%y %H:%M:%S"
+    required: true
+  end_time:
+    type: string
+    description: "End Date and time (UK Time) in the format: %d/%m/%y %H:%M:%S"
+    required: true
+  lib_entry_point:
+    type: string
+    default: workflows.hv_patch_and_reboot.patch_and_reboot
+    immutable: true
+  icinga_account_name:
+    type: string
+    description: "Name of Icinga Account to use. Must be configured in in the pack settings."
+    required: true
+    default: "default"
+runner_type: python-script
--- a/actions/hv.patch.reboot.yaml
+++ b/actions/hv.patch.reboot.yaml
@@ -8,14 +8,6 @@ parameters:
     type: string
     description: Name of host in Icinga
     required: true
-  start_time:
-    type: string
-    description: "Start Date and time (UK Time) in the format: %d/%m/%y %H:%M:%S"
-    required: true
-  end_time:
-    type: string
-    description: "End Date and time (UK Time) in the format: %d/%m/%y %H:%M:%S"
-    required: true
   lib_entry_point:
     type: string
     default: workflows.hv_patch_and_reboot.patch_and_reboot
--- a/actions/hv.patch.reboot.yaml
+++ b/actions/hv.patch.reboot.yaml
@@ -4,21 +4,17 @@ enabled: true
 entry_point: src/openstack_actions.py
 name: hv.patch.reboot
 parameters:
-  name:
-    type: string
-    description: Name of host in Icinga
-    required: true
   lib_entry_point:
     type: string
     default: workflows.hv_patch_and_reboot.patch_and_reboot
     immutable: true
   icinga_account_name:
     type: string
-    description: "Name of Icinga Account to use. Must be configured in in the pack settings."
+    description: Name of Icinga Account to use. Must be configured in the pack settings.
     required: true
     default: "default"
-  host:
-    description: Host to run action on
+  hypervisor_name:
+    description: Hypervisor to run action on, this should also be the host name on icinga.
     required: true
     type: string
   private_key_path:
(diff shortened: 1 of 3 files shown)


actions/hv.patch.reboot.yaml
```diff
@@ -0,0 +1,28 @@
+---
+description: Patch and Reboot a hypervisor
+enabled: true
+entry_point: src/openstack_actions.py
+name: hv.patch.reboot
+parameters:
+  name:
+    type: string
+    description: Name of host in Icinga
+    required: true
```
gmatthews20, line 10: We could remove this parameter as the hypervisor hostname is used as the name in icinga

actions/hv.patch.reboot.yaml
```diff
@@ -0,0 +1,28 @@
...
+  lib_entry_point:
+    type: string
+    default: workflows.hv_patch_and_reboot.patch_and_reboot
+    immutable: true
+  icinga_account_name:
+    type: string
+    description: "Name of Icinga Account to use. Must be configured in in the pack settings."
+    required: true
+    default: "default"
+  host:
```
gmatthews20, line 20: I think `hypervisor_name` may be better here, we can use this for both icinga and ssh connection, and also matches other workflows

lib/workflows/hv_patch_and_reboot.py
```diff
@@ -0,0 +1,44 @@
...
+from structs.icinga.downtime_details import DowntimeDetails
+from structs.icinga.icinga_account import IcingaAccount
+from structs.ssh.ssh_connection_details import SSHDetails
+from ssh_api.exec_command import SSHConnection
+
+
+def patch_and_reboot(
+    icinga_account: IcingaAccount,
+    name: str,
+    host: str,
```
gmatthews20, line 13: See previous comment on parameters

lib/workflows/hv_patch_and_reboot.py
```diff
@@ -0,0 +1,44 @@
...
+    private_key_path: str,
+):
+    connection_details = SSHDetails(
+        host=host, username="stackstorm", private_key_path=private_key_path
+    )
+    ssh_client = SSHConnection(connection_details)
+    start_time = datetime.datetime.now()
```
gmatthews20, line 20: We should use `datetime.utcnow()` as icinga expects the time to be in UTC and makes sure its not affected by DST

lib/workflows/hv_patch_and_reboot.py
```diff
@@ -0,0 +1,44 @@
...
+        start_time=start_timestamp,
+        end_time=end_timestamp,
+        comment=f"starting downtime to patch and reboot host: {name}",
+        is_fixed=True,
+        duration=end_timestamp - start_timestamp,
+    )
+    schedule_downtime(icinga_account=icinga_account, details=downtime_details)
+    try:
+        patch_out = ssh_client.run_command_on_host("patch").decode()
+        reboot_out = ssh_client.run_command_on_host("reboot").decode()
```
gmatthews20, line 36: I think it might make sense to move the `decode` to the return statement

lib/workflows/hv_patch_and_reboot.py
```diff
@@ -0,0 +1,44 @@
+import datetime
+
+from icinga_api.downtime import schedule_downtime, remove_downtime
+from structs.icinga.downtime_details import DowntimeDetails
+from structs.icinga.icinga_account import IcingaAccount
+from structs.ssh.ssh_connection_details import SSHDetails
+from ssh_api.exec_command import SSHConnection
+
+
+def patch_and_reboot(
```
gmatthews20, line 10: Can you add a docstring to this function

tests/lib/workflows/test_hv_patch_and_reboot.py
```diff
@@ -0,0 +1,125 @@
...
+    )
+    mock_ssh_conn.assert_called_once_with(
+        SSHDetails(
+            host=mock_host,
+            username="stackstorm",
+            private_key_path=mock_private_key_path,
+        )
+    )
+    mock_ssh_conn.return_value.run_command_on_host.assert_any_call("patch")
+    mock_ssh_conn.return_value.run_command_on_host.assert_any_call("reboot")
```
gmatthews20, line 57: I think this should be `assert_called_once_with`

tests/lib/workflows/test_hv_patch_and_reboot.py
```diff
@@ -0,0 +1,125 @@
...
+
+    mock_remove_downtime.assert_called_once_with(
+        icinga_account=icinga_account, object_type="Host", object_name=name
+    )
+
+
+@patch("workflows.hv_patch_and_reboot.schedule_downtime")
+@patch("workflows.hv_patch_and_reboot.SSHConnection")
+@patch("workflows.hv_patch_and_reboot.remove_downtime")
+def test_failed_schedule(mock_remove_downtime, mock_ssh_conn, mock_schedule_downtime):
```
gmatthews20, line 69: I think this should be called `test_failed_schedule_downtime`

tests/lib/workflows/test_hv_patch_and_reboot.py
```diff
@@ -0,0 +1,125 @@
...
+        patch_and_reboot(
+            icinga_account,
+            "example_host",
+            host=mock_host,
+            private_key_path=mock_private_key_path,
+        )
+
+    mock_ssh_conn.assert_called_once()
+    mock_schedule_downtime.assert_called_once()
+    mock_remove_downtime.assert_called_once()
```
gmatthews20, line 125: Could these be `assert_called_once_with`

gmatthews20 (changes requested):
Looks good just a few small tweaks

lib/workflows/hv_patch_and_reboot.py
```diff
@@ -0,0 +1,44 @@
...
+    try:
+        patch_out = ssh_client.run_command_on_host("patch").decode()
+        reboot_out = ssh_client.run_command_on_host("reboot").decode()
+
+    finally:
+        remove_downtime(
+            icinga_account=icinga_account,
+            object_type="Host",
+            object_name=name,
+        )
```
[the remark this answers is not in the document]
gmatthews20: Each of the icinga and ssh command will raise their own exceptions so we don't need to catch them and re-raise them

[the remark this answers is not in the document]
khalford: 👍

khalford (approved):

gmatthews20 (approved):

PR #289, opened 2024-12-11
Enable and Disable Hypervisor in Openstack StacksStorm Action

Dmitry-Popovichev:
### Description:

Created the Stackstorm action runner, workflow.py files and api call files to enable and disable hypervisors in openstack using stackstorm. Additionally added the unit tests to test the .py files.

### Submitter:

Have you (where applicable):

* [x] Added unit tests? Yes
* [x] Checked the latest commit runs on Dev? Yes

### Reviewer

Does this PR:

### DIFF
--- a/lib/openstack_api/openstack_service.py
+++ b/lib/openstack_api/openstack_service.py
@@ -17,6 +17,7 @@ def disable_service(
     :param hypervisor_name: The name or ID of the hypervisor.
     :param service_binary: The name of the service.
     :param disabled_reason: The reason for disabling the service.
+    :return: Returns the Service object.
     """
 
     if service.status == "enabled":
@@ -26,7 +27,9 @@ def disable_service(
             binary=service_binary,
             disabled_reason=disabled_reason,
         )
-    return None
+    raise RuntimeError(
+        f"Failed to disable {service_binary} on {hypervisor_name}. Already disabled."
+    )
 
 
 def enable_service(
@@ -41,6 +44,7 @@ def enable_service(
     :param service: The instance of the service class.
     :param hypervisor_name: The name or ID of the hypervisor.
     :param service_binary: The name of the service.
+    :return: Returns the Service object.
     """
 
     if service.status == "disabled":
--- a/lib/openstack_api/openstack_service.py
+++ b/lib/openstack_api/openstack_service.py
@@ -14,7 +14,7 @@ def disable_service(
     Disables an Openstack service
     :param conn: Openstack connection
     :param service: The instance of the service class.
-    :param hypervisor_name: (str): The name or ID of the hypervisor.
+    :param hypervisor_name: The name or ID of the hypervisor.
     :param service_binary: The name of the service.
     :param disabled_reason: The reason for disabling the service.
     """
@@ -39,7 +39,7 @@ def enable_service(
     Enables an Openstack service
     :param conn: Openstack connection
     :param service: The instance of the service class.
-    :param hypervisor_name: (str): The name or ID of the hypervisor.
+    :param hypervisor_name: The name or ID of the hypervisor.
     :param service_binary: The name of the service.
     """
 
@@ -47,4 +47,6 @@ def enable_service(
         return conn.compute.enable_service(
             service, host=hypervisor_name, binary=service_binary
         )
-    return None
+    raise RuntimeError(
+        f"Failed to enable {service_binary} on {hypervisor_name}. Already enabled."
+    )
--- a/lib/openstack_api/openstack_service.py
+++ b/lib/openstack_api/openstack_service.py
@@ -1,0 +1,67 @@
+from typing import Dict
+from openstack.connection import Connection
+from openstack.compute.v2.service.Service import *
+from openstack.compute.v2._proxy.Proxy import *
+
+'''
+Find hypervisor, using user input // Hypervisor not found
+Check to see if Nova-Compute service exists using the service ID
+If it does, check if it's already enbabled/disabled
+Enable/disable the service
+Let user know that's the action is complete (print output)
+
+
+Side effects:
+Authenticate the connection (already done)
+Api call to endpoint to find service (hypervisor not found, service not found)
+Authentication failed to hypervisor and/or service (handled by )
+Networking issues, talking to openstack api, too slow, no response (ahndled by openstack api)
+'''    
+
+
+def disable_service(
+    conn: Connection,
+    service_identifier: str,
+    hypervisor_name: str,
+    service_binary:str,
+    disable_reason: str,
+) -> None:
+    """
+    Disables an Openstack service
+    :param conn: Openstack connection
+    :param service_identifier: The service ID or instance.
+    :param hypervisor_name (str): The name or ID of the hypervisor.
+    :param service_binary: The name of the service.
+    :param disable_reason: The reason for disabling the service.
+    """
+
+    hypervisor_status = hypervisor_name.get("hypervisor_status")
+    if hypervisor_status == "enabled":
+        conn.compute.disable_service(service=service_identifier, host=hypervisor_name, binary=service_binary, disable_reason=disable_reason)
+        return f"The {service_binary} service has been disabled."
+    else:
+        return False, "Hypervisor is currently disabled - aborting."
+
+
+
+
+def enable_service(
+    conn: Connection, 
+    service_identifier: str,
+    hypervisor_name: str,
+    service_binary:str,
+) -> None:
+    """
+    Enables an Openstack service
+    :param conn: Openstack connection
+    :param service_identifier: The service ID or instance.
+    :param hypervisor_name (str): The name or ID of the hypervisor.
+    :param service_binary: The name of the service.
+    """
+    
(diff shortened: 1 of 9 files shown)


lib/openstack_api/openstack_service.py
```diff
@@ -0,0 +1,50 @@
...
+    service: Service,
+    hypervisor_name: str,
+    service_binary: str,
+    disabled_reason: str,
+) -> Optional[Service]:
+    """
+    Disables an Openstack service
+    :param conn: Openstack connection
+    :param service: The instance of the service class.
+    :param hypervisor_name: (str): The name or ID of the hypervisor.
```
khalford, line 17: We don't need this as the type is in the parameters above
```suggestion
    :param hypervisor_name: The name or ID of the hypervisor.
```

lib/openstack_api/openstack_service.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+
+    if service.status == "enabled":
+        return conn.compute.disable_service(
+            service,
+            host=hypervisor_name,
+            binary=service_binary,
+            disabled_reason=disabled_reason,
+        )
+    return None
```
khalford, line 29: Food for thought:

I wonder if an error should be raised rather than returning `None`.
E.g.
```suggestion
    raise RunetimeError(f"Failed to disable OpenStack service {binary} on host {host}. Reason: Service is already disabled.")
```

lib/openstack_api/openstack_service.py
```diff
@@ -0,0 +1,50 @@
...
+    :param service: The instance of the service class.
+    :param hypervisor_name: (str): The name or ID of the hypervisor.
+    :param service_binary: The name of the service.
+    """
+
+    if service.status == "disabled":
+        return conn.compute.enable_service(
+            service, host=hypervisor_name, binary=service_binary
+        )
+    return None
```
khalford, line 50: Ditto:
See above comment

lib/openstack_api/openstack_service.py
```diff
@@ -0,0 +1,50 @@
...
+    conn: Connection,
+    service: Service,
+    hypervisor_name: str,
+    service_binary: str,
+) -> Optional[Service]:
+    """
+    Enables an Openstack service
+    :param conn: Openstack connection
+    :param service: The instance of the service class.
+    :param hypervisor_name: (str): The name or ID of the hypervisor.
```
khalford, line 42: Type is already given in fuction definition
```suggestion
    :param hypervisor_name: The name or ID of the hypervisor.
```

lib/workflows/hv_service_actions.py
```diff
@@ -0,0 +1,42 @@
...
+def hv_service_disable(
+    conn: Connection,
+    hypervisor_name: str,
+    service_binary: str,
+    disabled_reason: str,
+) -> None:
+    """
+    Disables an Openstack service
+    :param conn: Openstack connection
+    :param hypervisor_name: (str): The name or ID of the hypervisor.
```
khalford, line 14: See other comments about typing
```suggestion
    :param hypervisor_name: The name or ID of the hypervisor.
```

lib/workflows/hv_service_actions.py
```diff
@@ -0,0 +1,42 @@
...
+
+def hv_service_enable(
+    conn: Connection,
+    hypervisor_name: str,
+    service_binary: str,
+) -> None:
+    """
+    Enables an Openstack service
+    :param conn: Openstack connection
+    :param hypervisor_name: (str): The name or ID of the hypervisor.
```
khalford, line 34: See other comments about typing
```suggestion
    :param hypervisor_name: The name or ID of the hypervisor.
```

tests/lib/openstack_api/test_openstack_service.py
```diff
@@ -0,0 +1,106 @@
+from unittest.mock import MagicMock, NonCallableMagicMock
+from openstack_api.openstack_service import disable_service, enable_service
+
+
+def test_disable_service_function_disables_service_when_enabled():
```
khalford, line 5: I would keep your fuction names short and leave the description to be in the doc string
```suggestion
def test_disable_service():
```

tests/lib/openstack_api/test_openstack_service.py
```diff
@@ -0,0 +1,106 @@
...
+        mock_service,
+        host=mock_hypervisor_name,
+        binary=mock_service_binary,
+        disabled_reason=mock_disable_reason,
+    )
+    # test that function returns == conn.compute.disable_service.return_value
+    assert res == mock_conn.compute.disable_service.return_value
+
+
+def test_disable_service_function_returns_none_when_disabled():
```
khalford, line 38: See above comment about naming. In this case I would append `fails` or something similar to the end to give a rough idea. But leave the details to be in the description.
```suggestion
def test_disable_service_fails():
```

tests/lib/openstack_api/test_openstack_service.py
```diff
@@ -0,0 +1,106 @@
...
+        mock_hypervisor_name,
+        mock_service_binary,
+        mock_disable_reason,
+    )
+
+    # assert expectations
+    assert res is None
+
+
+def test_enable_service_function_enables_service_when_disabled():
```
khalford, line 63: See above comments about naming

tests/lib/openstack_api/test_openstack_service.py
```diff
@@ -0,0 +1,106 @@
...
+    mock_conn.compute.enable_service.assert_called_once_with(
+        mock_service,
+        host=mock_hypervisor_name,
+        binary=mock_service_binary,
+    )
+
+    assert res == mock_conn.compute.enable_service.return_value
+
+
+def test_enable_service_function_returns_none_when_enabled():
```
khalford, line 89: Ditto about naming

tests/lib/workflows/test_hv_service_actions.py
khalford, line 1: Potentially missing tests for when the disable service function fails and when the enable function fails

khalford (changes requested):
Looks great. Just some small things.

About comments:
Personally I wouldn't put comments in the code unless you need to explain what a very complicated statement does I.E lambda functions. In your case it's just grouping together blocks of code and naming them. Your code should be understandable by it's structure and layout without comments. Feel free to ignore me if you think they are beneficial to yourself and others.

About the comment I made on raising an error rather than returning none. @anish-mudaraddi Do you have any thoughts on this? I am not completely aware of how the rest of the openstack api is written.

lib/openstack_api/openstack_service.py
Dmitry-Popovichev: I had something very similar in before, but it was running on a `print` statement rather than a `RuntimeError`. After a discussion with Anish and for the simplicity of the unit tests, we went ahead with returning `None` instead.

lib/openstack_api/openstack_service.py
Dmitry-Popovichev: Oh yep nice, good spot, I think that must have been a copy and paste mistake.

tests/lib/openstack_api/test_openstack_service.py
Dmitry-Popovichev: Another conversation I had about unit tests with Anish where the reason we put a long description in the function title, is to easier see locate what function in an error log.

In this case it might be difficult to differentiate the two function names as one tests when the service is already enabled and the other tests when it's not.

Let me know your thoughts.

tests/lib/workflows/test_hv_service_actions.py
Dmitry-Popovichev: Would that not be the tests that test when for example: if the user is trying to disable the service when it's already disabled, it returns none?

Or is there another fail scenario that perhaps I didn't think of?

khalford: I looked again and it's my mistake. Since your function does not return anything and an error is not raised if it "fails" then there is no other case to test

tests/lib/openstack_api/test_openstack_service.py
khalford: If it makes working on code easier for you and others than I would say keep it. No harm either ways

lib/openstack_api/openstack_service.py
Dmitry-Popovichev: That is a very valid scenario and I'm sort of on the fence about this now because as I mentioned I did have an error printing earlier but Anish made a compelling argument to not have to have one. You both have a compelling argument really. 

EDIT: After playing around with StackStorm and looking at the logs after disabling and enabling an HV, I think I would agree, I would like a `RuntimeError` or something similar to print out.

We can search by status using a seperate stackstorm action to check if the HV is enabled/disabled, but it does take extra steps.

Happy to show you how it all works in stackstorm and perhaps we can go over the code together.

khalford: Sounds good

gmatthews20 (changes requested):

khalford (approved):

gmatthews20 (approved):

PR #293, opened 2024-12-20
Migrate VMs not in ACTIVE state

lizsalmon:
### Description:
- Added in cold migration if the state is STUTOFF
- Returns an error if the state is not ACTIVE or SHUTOFF
- Added the snapshot and migrate function into the `openstack_api` so that we don't need the extra `server_migrate.py` file and then adjusted the action to run from the api rather than a workflow.  

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:


lib/openstack_api/openstack_server.py
```diff
@@ -4,19 +4,30 @@
...
+    if snapshot:
+        snapshot_server(conn=conn, server_id=server_id)
+    if server_status == "ACTIVE":
+        live = True
+    elif server_status == "SHUTOFF":
+        live = False
+    else:
+        raise ValueError(
+            f"Server status: {server_status}. The server must be ACTIVE or SHUTOFF to be migrated"
+        )
```
anish-mudaraddi, line 30: here's a more concise way to write this

```python
if server_status not in ["ACTIVE", "SHUTOFF"]:
    raise ValueError(...)

if server_status == "SHUTOFF":
    return conn.compute.migrate_server(...)
return conn.compute.live_migrate_server(...)
```

saves you a few lines :) - makes it easier to read too!

anish-mudaraddi (changes requested):

lib/workflows/search_by_property.py
```diff
@@ -43,6 +43,8 @@ def search_by_property(
     query = getattr(openstackquery, query_type)()
     if not properties_to_select:
         query.select_all()
+    elif webhook == "migrate-server":
+        query.select(*properties_to_select, "server_id", "server_status")
```
anish-mudaraddi, line 47: why do you need this?

lib/openstack_api/openstack_server.py
```diff
@@ -4,25 +4,31 @@
...
-        conn.compute.migrate_server(server=server_id, host=dest_host)
-    else:
-        conn.compute.live_migrate_server(
-            server=server_id, host=dest_host, block_migration=True
+    if snapshot:
+        snapshot_server(conn=conn, server_id=server_id)
+    if server_status not in ["ACTIVE", "SHUTOFF"]:
+        raise ValueError(
+            f"Server status: {server_status}. The server must be ACTIVE or SHUTOFF to be migrated"
         )
```
khalford, line 26: I wonder if it's better to check for active and shutoff before migrate? The way it currently is. We will build up snapshots of every VM even if we don't migrate them.
```suggestion
   if server_status not in ["ACTIVE", "SHUTOFF"]:
        raise ValueError(
            f"Server status: {server_status}. The server must be ACTIVE or SHUTOFF to be migrated"
        )
    if snapshot:
        snapshot_server(conn=conn, server_id=server_id)
    
```

khalford (changes requested):

lib/openstack_api/openstack_server.py
lizsalmon: Yeah that seems like a more sensible option to me, thanks- I will change it now

khalford (approved):

lib/workflows/search_by_property.py
gmatthews20: the `to_webhook` functions just sends whichever properties you've selected with the query.

Something like this to send everything might work better saves writing checks for each webhook
`to_webhook(webhook=webhook, payload=query.select_all().to_props())`

lizsalmon: Ah sorry yeah that makes sense! Just changed.

anish-mudaraddi (approved):
Thanks @lizsalmon @gmatthews20 for sorting this out - looks much better

khalford (approved):

PR #294, opened 2024-12-20
Changed tests to use pytest instead of unittest.

lizsalmon:
### Description:
- After a discussion with Anish I've changed some of the `unittest` classes to be `pytest` tests and added `pytest.fixtures` if needed. 
- There are a few changes that I am unsure of with the fixtures (specifically in `test_hypervisor_state_sensor`. 
---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/tests/sensors/test_hypervisor_state_sensor.py
+++ b/tests/sensors/test_hypervisor_state_sensor.py
@@ -6,12 +6,11 @@ from sensors.src.hypervisor_state_sensor import HypervisorStateSensor
 
 @pytest.fixture(name="sensor")
 def state_sensor_fixture():
-    sensor_service = MagicMock()
-    config = {"hypervisor_sensor": {"cloud_account": "dev", "uptime_limit": 180}}
-    poll_interval = 10
-    sensor = HypervisorStateSensor(sensor_service, config, poll_interval)
-    sensor.sensor_service = sensor_service  # So that I dont need two fixtures?
-    return sensor
+    return HypervisorStateSensor(
+        sensor_service=MagicMock(),
+        config={"hypervisor_sensor": {"cloud_account": "dev", "uptime_limit": 180}},
+        poll_interval=10,
+    )
 
 
 @patch("sensors.src.hypervisor_state_sensor.get_hypervisor_state")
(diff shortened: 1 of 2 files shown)


khalford (dismissed):

tests/sensors/test_hypervisor_state_sensor.py
```diff
@@ -1,62 +1,57 @@
...
-    Test cases for testing hypervisor state sensor
-    """
+@pytest.fixture(name="sensor")
+def state_sensor_fixture():
+    sensor_service = MagicMock()
+    config = {"hypervisor_sensor": {"cloud_account": "dev", "uptime_limit": 180}}
+    poll_interval = 10
+    sensor = HypervisorStateSensor(sensor_service, config, poll_interval)
+    sensor.sensor_service = sensor_service  # So that I dont need two fixtures?
+    return sensor
```
anish-mudaraddi, line 14: you could do this: 

```python

@pytest.fixture(name="sensor")
def state_sensor_fixture():
    return HypervisorStateSensor(
        sensor_service=MagicMock(),
        config={"hypervisor_sensor": {"cloud_account": "dev", "uptime_limit": 180}},
        poll_interval=10
    )
```

I don't get why you're reasserting 
`sensor.sensor_service = sensor_service`
since you're passing sensor_serivce as a magic_mock when creating the object it should be set already in the constructor

tests/sensors/test_hypervisor_state_sensor.py
```diff
@@ -1,62 +1,57 @@
...
-        )
-
-        expected_payload = {
+@patch("sensors.src.hypervisor_state_sensor.get_hypervisor_state")
+@patch("sensors.src.hypervisor_state_sensor.query_hypervisor_state")
+def test_poll(mock_query_hypervisor_state, mock_get_hypervisor_state, sensor):
+    """
+    Test main function of sensor, polling state of hypervisor state
+    """
+    sensor.sensor_service = MagicMock()
```
anish-mudaraddi, line 23: you don't need this line since you set sensor_service to MagicMock() in the fixture

anish-mudaraddi (changes requested):

tests/sensors/test_hypervisor_state_sensor.py
lizsalmon: Yeah you're right that's not needed - thanks.

anish-mudaraddi (approved):

khalford (approved):

PR #295, opened 2025-01-10
BUG: group by does not accept default of null

khalford:
### Description:

Action fails to run with default values as null is not an accepted value, adding "null" to the list of Enums fixes this.

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/project.search.by.property.yaml
+++ b/actions/project.search.by.property.yaml
@@ -105,6 +105,7 @@ parameters:
       - "is_enabled"
       - "name"
       - "parent_id"
+      - "null"
   sort_by:
     description: "(Optional) comma-spaced list of project properties to sort by (by ascending only)
     - multiple of the same property will be ignored.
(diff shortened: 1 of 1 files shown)


actions/project.search.by.property.yaml
```diff
@@ -105,6 +105,7 @@ parameters:
       - "is_enabled"
       - "name"
       - "parent_id"
+      - "null"
```
lizsalmon, line 108: I thing it needs to not be a string. Like in the `image.search.by.expression.yaml`
```suggestion
      - null
```

lizsalmon (changes requested):

lizsalmon (approved):

DaveW-STFC (changes requested):
If this was an issue on the Group By then should we not also look at the Sort By clause?

khalford:
> If this was an issue on the Group By then should we not also look at the Sort By clause?

Good point, I've looked and the Sort By doesn't use Enums and so isn't affected by this issue

DaveW-STFC (approved):

PR #296, opened 2025-01-13
ENH: Add restrictions to automatic migrations

gmatthews20:
### Description:

- Prevents any migration of GPU (g-*) flavors, this applies to all migration actions
- Blacklist and Whitelist flavor able to be migrated through the migration webhook, these flavors can still be migrated by running the action manually
- Add a concurrency policy to limit the number of concurrent migrations running at a time
- Fix icinga downtime action

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/openstack_api/openstack_server.py
+++ b/lib/openstack_api/openstack_server.py
@@ -8,6 +8,7 @@ def snapshot_and_migrate_server(
     conn: Connection,
     server_id: str,
     server_status: str,
+    flavor_id: str,
     snapshot: bool,
     dest_host: Optional[str] = None,
 ) -> None:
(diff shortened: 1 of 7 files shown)


lizsalmon (approved):

rules/webhook.server.migrate.yaml
```diff
@@ -9,8 +9,24 @@ trigger:
...
+    type: regex
+    pattern: (^l3.*)|(^l2.*)
+
+  # Blacklist specific flavors from migration
+  # e.g. pattern ["l2.tiny", "l3.large"] will prevent migrations of those flavors
+  # specifically but not all l2 and l3 flavors. Flavors here can still be migrated
+  # manually using the stackstorm_openstack.server.migrate action
+  trigger.body.flavor_id#2:
+    type: ninside
+    pattern: ["l3.micro"]
```
DavidFair, line 25: Why have we blocked this / allowed these or is this an example?
Just surprised we've not got l5/l6 and micro is blocked

lib/openstack_api/openstack_server.py
```diff
@@ -16,8 +17,15 @@ def snapshot_and_migrate_server(
...
     :param server_id: Server ID to migrate
     :param server_status: Status of machine to migrate - must be ACTIVE or SHUTOFF
+    :param flavor_name: Server flavor name
     :param dest_host: Optional host to migrate to, otherwise chosen by scheduler
     """
+
+    if flavor_id.startswith("g-"):
+        raise ValueError(
+            f"Attempted to move GPU flavor, {flavor_id}, which is not allowed!"
+        )
```
DavidFair, line 27: We should use an `if any(flavour_id.startswith(i) for i in BLOCKED_PREFIXES)" then we can easily add values since I think some other flavours (`c-`) need to be added here too

DavidFair (changes requested):

rules/webhook.server.migrate.yaml
gmatthews20: This was just an example that I was using for testing, are there any that should be actually blocked?

DavidFair (approved):

PR #297, opened 2025-01-15
ENH: Add rule to remove downtime

lizsalmon:
### Description:
This PR:
- Adds values to the enum in `hypervisor_states.py` to add states for all possible machines.  REBOOTED, EMPTY, DOWN
- Adds in a rule to remove the downtime on a hypervisor being "patched and rebooted" only once it is back up (in rebooted state)
---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/workflows/hv_patch_and_reboot.py
+++ b/lib/workflows/hv_patch_and_reboot.py
@@ -1,6 +1,5 @@
 import datetime
-
-from icinga_api.downtime import schedule_downtime, remove_downtime
+from icinga_api.downtime import schedule_downtime
 from structs.icinga.downtime_details import DowntimeDetails
 from structs.icinga.icinga_account import IcingaAccount
 from structs.ssh.ssh_connection_details import SSHDetails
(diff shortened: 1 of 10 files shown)


gmatthews20 (changes requested):

lib/workflows/hv_patch_and_reboot.py
```diff
@@ -37,14 +37,6 @@ def patch_and_reboot(
...
-    try:
-        patch_out = ssh_client.run_command_on_host("patch")
-        reboot_out = ssh_client.run_command_on_host("reboot")
-
-    finally:
-        remove_downtime(
-            icinga_account=icinga_account,
-            object_type="Host",
-            object_name=hypervisor_name,
-        )
```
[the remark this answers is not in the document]
lizsalmon: Should I only catch `SSHException`?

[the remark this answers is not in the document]
gmatthews20: Yes I think so, you may also have to re-raise the exception after removing the downtime

tests/lib/openstack_api/test_openstack_hypervisor.py
```diff
@@ -58,68 +59,115 @@ def test_get_state_drained():
...
 
-def test_get_state_unkown_state():
+def test_get_state_empty():
     """
-    Test hypervisor state is unknown when missing state
+    Test hypervisor state is empty for given variables
     """
     hypervisor = {
-        "hypervisor_uptime_days": 100,
+        "hypervisor_uptime_days": 40,
```
meoflynn, line 81: Is this the HV uptime limit by default?

lizsalmon: I am not sure - the value from the templates seems to be 180.

gmatthews20: Yes the default value is 180, it is set here: https://github.com/stfc/st2-cloud-pack/blob/d761543432c2870bd5142034c1b59c1e0e26b056/sensors/src/hypervisor_state_sensor.py#L27

gmatthews20 (approved):

PR #299, opened 2025-01-21
ENH: Merge project creation actions

gmatthews20:
### Description:

Create new action which can create both internal and external projects and can also take an domain in which to create the project 
Placeholder action which new project creation action calls

### Special Notes:

- Merge internal and external project creation
- Network, Subnet, and Router set to <project_name>-{Network, Subnet, Router} respectively instead of user set parameters
- Project domain parameter, not yet fully implemented
- jasmin networking_type option, currently calls `create_external_project`

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:


actions/project.create.yaml
```diff
@@ -38,35 +46,30 @@ parameters:
     type: boolean
   parent_id:
     default: ""
-    description: Set the ID of the parent Domain/project. Leave as blank and it will use the "Default Domain"
+    description: Set the ID of the parent project.
     type: string
   admin_user_list:
-    required: false
+    required: true
```
DavidFair, line 52: These should remain `false` especially as we're hopefully going to drop them as another incentive to use groups for membership instead of direct role assignment

Plus some projects get created ahead of knowing which users we'll have if we do stick with it here

actions/project.create.yaml
```diff
@@ -38,35 +46,30 @@ parameters:
...
     default: []
     type: array
     description: List of Users (IDs or Names) to give administrator access
-  stfc_user_list:
-    required: false
+  user_list:
+    required: true
```
DavidFair, line 57: ^

actions/project.create.yaml
```diff
@@ -38,35 +46,30 @@ parameters:
...
     default: []
     type: array
-    description: List of Users (stfc domain) (IDs or Names) to give local user access
-  network_name:
-    required: true
+    description: List of Users (IDs or Names) to give local user access
+  networking_type:
+    description: Type of networking, internal or external
```
DavidFair, line 62: Should be `internal`, `external` or `jasmin`
Where `jasmin` == external under the hood at the moment

actions/project.create.yaml
```diff
@@ -38,35 +46,30 @@ parameters:
...
-    description: Router Name to create for Project
+    default: internal
+    enum:
+      - internal
+      - external
   number_of_floating_ips:
     required: true
     type: integer
     default: 1
-    description: Number of Floating Ips to allocate to the project - and is also the quota
```
DavidFair, line 69: If this is also the quota I think it's wrong?

@jacob-ward will confirm, but I think it's something like 10 by default? So we should fix that whilst we're here

lib/workflows/create_project.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def create_project(
+    conn: Connection,
+    project_name: str,
+    project_email: str,
+    project_description: str,
+    project_domain: str,
+    networking_type: str,
+    number_of_floating_ips: int = 1,
```
DavidFair, line 15: ^

lib/workflows/create_project.py
```diff
@@ -0,0 +1,53 @@
...
+    project_immutable: Optional[bool] = False,
+    parent_id: Optional[str] = None,
+    admin_user_list: Optional[List[str]] = None,
+    user_list: Optional[List[str]] = None,
+):
+    if project_domain != "default":
+        raise RuntimeError(
+            "domain support is work in progress, please use default domain",
+        )
+    return (
```
DavidFair, line 26: Whilst this is brief I feel like it's a bit clunky readability-wise. Could we do something like:


```
if networking_type == "internal":
           create_internal_project(...)
else if networking_type == "external" or networking_type == "jasmin"
           create_external_project(...)
else:
           raise NotImplementedError(f"Unknown networking type {networking_type}")
```

DavidFair (changes requested):

actions/project.create.yaml
jacob-ward: We've dropped the default quota to 1 globally to discourage overuse of floating IPs. For JASMIN bumping this to 5 is a sensible mid ground. Projects are expected be small in terms of VM count but all will need to be externally accessible in some form.

actions/project.create.yaml
meoflynn: I think I have set that variable to be optional in the initial project workflow PR https://github.com/stfc/st2-cloud-pack/pull/302

gmatthews20: Should both user and admins list be optional, shouldn't we have at least one admin added to the project until we move to assigning groups?

actions/project.create.yaml
gmatthews20: Should the quota parameters be removed and the set the quota based on networking_type/domain/...?

jacob-ward: There are still cases where users specifically request higher counts (or we know they'll need them based on their use case) so leaving it as is would be preferred please

actions/project.create.yaml
DavidFair: Not in this case, because it would they would become an OpenStack admin (for the whole of OpenStack) not just a project admin
This gets used typically for cloud team members e.g. ops

actions/project.create.yaml
DavidFair: Yeah, this should be done in the backend code @meoflynn

meoflynn (approved):

actions/project.create.yaml
```diff
@@ -0,0 +1,77 @@
...
+    required: true
+    type: string
+    default: "dev"
+    enum:
+      - "dev"
+      - "prod"
+  project_domain:
+    description: Domain to create project in
+    type: string
+    default: stfc
```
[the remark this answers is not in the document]
anish-mudaraddi: yea projects have "default" domain id as default - I'd change this to match

actions/project.create.yaml
meoflynn: Would this be adding `jasmin`to the `RbacNetworkActions` enum to use?

lib/workflows/create_project.py
```diff
@@ -0,0 +1,56 @@
...
+            admin_user_list=admin_user_list,
+            stfc_user_list=user_list,
+        )
+    elif networking_type in ("external", "jasmin"):
+        create_external_project(
+            conn=conn,
+            project_name=f"{project_name}",
+            project_email=project_email,
+            project_description=project_description,
+            network_name=f"{project_name}-network",
```
meoflynn, line 45: Isn't the network for external projects just `External`?

lib/workflows/create_project.py
```diff
@@ -0,0 +1,56 @@
...
+    networking_type: str,
+    number_of_floating_ips: int = 1,
+    number_of_security_group_rules: int = 200,
+    project_immutable: Optional[bool] = False,
+    parent_id: Optional[str] = None,
+    admin_user_list: Optional[List[str]] = None,
+    user_list: Optional[List[str]] = None,
+):
+
+    if project_domain != "default":
```
meoflynn, line 23: Could we have a docstring for this method?

meoflynn (approved):

meoflynn:
I have tested the PR on a dev instance and the project creation works

DavidFair (approved):

PR #302, opened 2025-01-23
ENH: Add initial new project workflow

meoflynn:
Initial workflow for creating an external project in a new domain. Updated the create_project method to include option for domain id and added new domain to user_domains

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/openstack_api/openstack_project.py
+++ b/lib/openstack_api/openstack_project.py
@@ -36,7 +36,7 @@ def create_project(
     }
 
     # get domain id for the domain to use
-    domain_details = conn.identity.get_domain(project_details.project_domain)
+    domain_details = conn.identity.find_domain(project_details.project_domain)
     create_project_kwargs["domain_id"] = domain_details.id
 
     if project_details.parent_id:
--- a/lib/openstack_api/openstack_project.py
+++ b/lib/openstack_api/openstack_project.py
@@ -26,7 +26,7 @@ def create_project(
     email_regex = r"^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$"
     if not re.match(email_regex, project_details.email):
         raise ValueError("The project contact email is invalid")
-    if not project_details.domain:
+    if not project_details.project_domain:
         raise MissingMandatoryParamError("The domain for the project is missing")
 
     create_project_kwargs = {
@@ -36,7 +36,7 @@ def create_project(
     }
 
     # get domain id for the domain to use
-    domain_details = conn.identity.get_domain(project_details.domain)
+    domain_details = conn.identity.get_domain(project_details.project_domain)
 
     create_project_kwargs["domain_id"] = domain_details.id
 
--- a/lib/openstack_api/openstack_project.py
+++ b/lib/openstack_api/openstack_project.py
@@ -18,7 +18,7 @@ def create_project(
     :param project_details: A datastructure containing all the project details
     :return: A clouds project, or None if it was unsuccessful
     """
-    # domain_id can be blank, we will create a project in the default domain
+
     if not project_details.name:
         raise MissingMandatoryParamError("The project name is missing")
     if not project_details.email:
@@ -32,6 +32,11 @@ def create_project(
         "description": project_details.description,
         "is_enabled": project_details.is_enabled,
     }
+
+    # create project in a specific domain if ID given, otherwise default domain
+    if project_details.domain_id:
+        create_project_kwargs["domain_id"] = project_details.domain_id
+
     if project_details.parent_id:
         create_project_kwargs["parent_id"] = project_details.parent_id
 
--- a/lib/openstack_api/openstack_project.py
+++ b/lib/openstack_api/openstack_project.py
@@ -26,8 +26,8 @@ def create_project(
     email_regex = r"^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$"
     if not re.match(email_regex, project_details.email):
         raise ValueError("The project contact email is invalid")
-    # if not project_details.project_domain:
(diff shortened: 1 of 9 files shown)


meoflynn:
The `icinga` file is only in the PR when I fixed the formatting of the files in my branch

lib/structs/project_details.py
```diff
@@ -2,11 +2,13 @@
 from typing import Optional
 
 
+# pylint: disable=too-many-instance-attributes
 @dataclass
 class ProjectDetails:
     name: str
     email: str
     description: str = ""
+    domain: str = None
```
anish-mudaraddi, line 11: `domain: Optional[str] = None`

lib/workflows/create_jasmin_external_project.py
```diff
@@ -0,0 +1,180 @@
...
+    conn: Connection,
+    project_name: str,
+    project_email: str,
+    domain: str,
+    project_description: str,
+    network_name: str,
+    subnet_name: str,
+    router_name: str,
+    number_of_floating_ips: int = 1,
+    number_of_security_group_rules: int = 200,
```
anish-mudaraddi, line 46: both of these need `Optional[int]` type hinting

lib/workflows/create_jasmin_external_project.py
```diff
@@ -0,0 +1,180 @@
...
+    :param number_of_floating_ips: Number of floating ips to assign to project (default=1)
+    :param number_of_security_group_rules: Quota for the no. of security group rules to assign (default=200)
+    :param project_immutable: (Optional) A boolean representing if project is immutable (True) or not (False)
+    :param parent_id: (Optional) A string for parent project
+    :param admin_user_list: (Optional) A list of strings to add as admins to the project
+    :param jasmin_user_list: A list of strings to add as regular users (jasmin domain)
+    """
+
+    if not admin_user_list:
+        admin_user_list = []
```
anish-mudaraddi, line 71: you can write this more concise 
```admin_user_list = admin_user_list or []```

File: lib/workflows/create_jasmin_external_project.py
Comment (anish-mudaraddi) [inline lib/workflows/create_jasmin_external_project.py:1]: isn't this a direct copy of `create_external_project`? 
Maybe we could just modify that to accept both generic and jasmin external projects?

The only thing I can see changing is using Jasmin users instead of STFC ones and maybe some more bespoke networking rules

I guess it's fine for now since we're under a deadline - but something to think about

**Review by anish-mudaraddi — CHANGES_REQUESTED** [review]

File: lib/openstack_api/openstack_project.py [hunk 4]
```diff
@@ -18,20 +18,30 @@ def create_project(
...
         "description": project_details.description,
         "is_enabled": project_details.is_enabled,
     }
+
+    # get domain id for the domain to use
+    domain_name = project_details.domain
+    domain_details = conn.identity.get_domain(domain_name)
+    domain_id = domain_details.id
+
+    create_project_kwargs["domain_id"] = domain_id
```
khalford, line 43: I think this might be simpler as we are not reusing these variables elsewhere.
```suggestion
    domain_details = conn.identity.get_domain(project_details.domain)

    create_project_kwargs["domain_id"] = domain_details.id
```

lib/structs/project_details.py
meoflynn: There's an edit I do in the openstack_project where we need to take a domain id as a required parameter - this is because the openstacksdk doesn't pass the parent id as a domain id so we would need to differentiate between the domains for the projects

lib/workflows/create_jasmin_external_project.py
meoflynn: This section isn't part of getting the initial workflow up - to be done in a separate PR (or built on this PR)

lib/openstack_api/openstack_project.py
meoflynn: Good spot, I was meaning to simplify the line but forgot when writing the unit test for it 😅

lib/structs/project_details.py
meoflynn: I have a feature branch for the internal project and external project workflow that I'm working on as well that take domain as a parameter (as the domain is a new field in the project creation action)

lib/workflows/create_jasmin_external_project.py
DavidFair: TBH I think we should drop these args across the board and hardcode them to whatever ops say rather than passing them as args everywhere
Instead we should make quota changes using the dedicated action later on and start with sensible defaults

It's out of scope for this PR but I'd like to do a follow-on PR that nukes this everywhere

lib/openstack_api/openstack_project.py
```diff
@@ -18,20 +18,27 @@ def create_project(
...
+        raise MissingMandatoryParamError("The domain for the project is missing")
 
     create_project_kwargs = {
         "name": project_details.name,
         "description": project_details.description,
         "is_enabled": project_details.is_enabled,
     }
+
+    # get domain id for the domain to use
+    domain_details = conn.identity.get_domain(project_details.project_domain)
```
gmatthews20, line 39:
```suggestion
    domain_details = conn.identity.find_domain(project_details.project_domain)
```
This should be `find_domain` since `get_domain` requires us to already know the domain id to get the domain, `find_domain` can use either name or id

gmatthews20 (changes requested):

lib/openstack_api/openstack_project.py
DavidFair: I've addressed this in a follow-up commit

gmatthews20 (approved):

LunaBarrett (commented):
Looks good, need to check with James Cornall if the workflow has been tested

anish-mudaraddi (approved):

PR #304, opened 2025-01-24
Create alertmanager api

jose-caballero:
Address tickets:

* https://stfc.atlassian.net/browse/CLDSCRM-2238
* https://stfc.atlassian.net/browse/CLDSCRM-2290

### DIFF
--- a/lib/alertmanager_api/silence.py
+++ b/lib/alertmanager_api/silence.py
@@ -1,0 +1,198 @@
+import logging
+import requests
+from exceptions.missing_mandatory_param_error import MissingMandatoryParamError
+from requests.auth import HTTPBasicAuth
+from requests.exceptions import RequestException
+from alertmanager_api.silence_details import SilenceDetailsHandler
+
+
+class AlertManagerAPI:
+    """
+    class implementing an interface to the Kayobe's alertmanager
+    functionalities:
+        * create a silence
+        * remove an existing silence
+    """
+
+    def __init__(self, alertmanager_account):
+        """
+        :param alertmanager_account: object with credentials
+        :type alertmanager_account: AlertManagerAccount
+        """
+        self.log = logging.getLogger("AlertManagerAPI")
+        self.alertmanager_account = alertmanager_account
+        self.auth = HTTPBasicAuth(
+            alertmanager_account.username, alertmanager_account.password
+        )
+        self.api_endpoint = alertmanager_account.alertmanager_endpoint
+
+    def schedule_silence(self, silence):
+        """
+        Schedules a silence in alertmanager
+
+        :param silence: object with the specs to create a silence
+        :type silence: SilenceDetails
+        :return: a new silence object, including the silence_id
+        :rtype: SilenceDetails
+        :raises MissingMandatoryParamError:
+            if any of the required attributes are missing
+            in the silence object:
+            instance_name
+            start time
+            end time
+            author
+            comment
+        :raises requests.exceptions.RequestException:
+            when the request to the AlertManager failed
+        """
+        if not silence.instance_name:
+            msg = "Missing silence instance name"
+            self.log.critical(msg)
+            raise MissingMandatoryParamError(msg)
+        if not silence.start_time_dt:
+            msg = "Missing silence start time"
+            self.log.critical(msg)
+            raise MissingMandatoryParamError(msg)
+        if not silence.end_time_dt:
+            msg = "Missing silence end time"
+            self.log.critical(msg)
+            raise MissingMandatoryParamError(msg)
+        if not silence.author:
+            msg = "Missing silence author"
+            self.log.critical(msg)
+            raise MissingMandatoryParamError(msg)
+        if not silence.comment:
+            msg = "Missing silence comment"
+            self.log.critical(msg)
+            raise MissingMandatoryParamError(msg)
+
+        payload = {
+            "matchers": [
+                {
+                    "name": "instance",
+                    "value": silence.instance_name,
+                    "isRegex": False,
+                }
+            ],
+            "startsAt": silence.start_time_str,
+            "endsAt": silence.end_time_str,
+            "createdBy": silence.author,
+            "comment": silence.comment,
+        }
+        try:
+            api_url = f"{self.api_endpoint}/api/v2/silences"
+            response = requests.post(
+                api_url,
+                auth=self.auth,
+                headers={"Accept": "application/json"},
+                json=payload,
+                timeout=10,
+            )
+            response.raise_for_status()
+            silence_id = response.json()["silenceID"]
+            silence.silence_id = silence_id
+            return silence
+        except requests.HTTPError as req_ex:
+            msg = f"Failed to create silence in Alertmanager: {req_ex} "
+            if req_ex.response is not None:
+                msg += (
+                    "Failed to create silence. "
+                    f"Response status code: {req_ex.response.status_code} "
+                    f"Response text: {req_ex.response.text}"
+                )
+            self.log.critical(msg)
+            raise req_ex
+        except Exception as req_ex:
+            # to capture any other Exception that is not HTTPError
+            msg = f"Failed to create silence in Alertmanager: {req_ex} "
+            self.log.critical(msg)
+            raise req_ex
+
+    def remove_silence(self, silence):
+        """
+        Removes a previously scheduled silence in alertmanager
+
+        :param silence: object including the silence_id to be removed
+        :type silence: SilenceDetails
+        :raises MissingMandatoryParamError: when the silence_id is missing
+        :raises requests.exceptions.RequestException:
+            when the request to the AlertManager failed
+        """
+        if not silence.silence_id:
+            msg = "Missing silence silence_id"
+            self.log.critical(msg)
+            raise MissingMandatoryParamError(msg)
+
+        try:
+            api_url = f"{self.api_endpoint}/api/v2/silence/{silence.silence_id}"
+            response = requests.delete(
+                api_url,
+                auth=self.auth,
+                timeout=10,
+            )
+            response.raise_for_status()
+        except requests.HTTPError as req_ex:
+            msg = f"Failed to delete silence in Alertmanager: {req_ex} "
+            if req_ex.response is not None:
+                msg += (
+                    f"Failed to delete silence with ID {silence.silence_id}. "
+                    f"Response status code: {response.status_code} "
+                    f"Response text: {response.text}"
+                )
+            self.log.critical(msg)
+            raise req_ex
+        except Exception as req_ex:
+            # to capture any other Exception that is not HTTPError
+            msg = f"Failed to create silence in Alertmanager: {req_ex} "
+            self.log.critical(msg)
+            raise req_ex
+
+    def remove_silences(self, silence_instances):
+        """
+        Removes a list of previously scheduled silences in alertmanager
+
+        :param silence_instances: a list of silence instances
+        :type silence: SilenceDetailsHandler
+        """
+        for silence in silence_instances:
+            self.remove_silence(silence)
+
+    def get_silences(self):
+        """
+        download all silence events recorded in AlertManager
+
+        :return: the list of Silence events currently recorded in AlertManager
+        :rtype: SilenceDetailsHandler
+        """
+        try:
+            api_url = f"{self.api_endpoint}/api/v2/silences"
+            response = requests.get(api_url, auth=self.auth, timeout=10)
+            response.raise_for_status()
+            silences = response.json()
+            handler = SilenceDetailsHandler()
+            handler.add_from_json(silences)
+            return handler
+        except requests.HTTPError as req_ex:
+            msg = f"Failed to get silences from Alertmanager: {req_ex} "
+            if req_ex.response is not None:
+                msg += (
+                    f"Failed to get silences. "
+                    f"Response status code: {response.status_code} "
+                    f"Response text: {response.text}"
+                )
+            self.log.critical(msg)
+            raise req_ex
+        except RequestException as req_ex:
+            msg = f"Failed to get silences from Alertmanager: {req_ex} "
+            if req_ex.response is not None:
+                msg += (
+                    f"Response status code: {req_ex.response.status_code} "
+                    f"Response text: {req_ex.response.text}"
+                )
+            self.log.critical(msg)
+            raise req_ex
+        except Exception as req_ex:
+            # to capture any other Exception that is not HTTPError
+            msg = f"Failed to get silences from Alertmanager: {req_ex} "
+            self.log.critical(msg)
+            raise req_ex
--- a/lib/alertmanager_api/silence_details.py
+++ b/lib/alertmanager_api/silence_details.py
@@ -1,0 +1,265 @@
+import json
+from datetime import datetime
+
+
+class SilenceDetails:  # pylint: disable=too-few-public-methods
+    def __init__(
+        self,
+        instance_name=None,
+        start_time_dt=None,
+        end_time_dt=None,
+        author=None,
+        comment=None,
+        silence_id=None,
+        status=None,
+    ):
+        """
+        class to handle together all attributes describing a Silence event.
+        The values for the start and end times are datetime objects.
+        To convert a string like this (we assume it is UTC):
+
+            2025-01-22T12:50:00
+
+        into a datetime object, code like this can be used:
+
+            naive_datetime = datetime.strptime(datetime_string, datetime_format)
+            utc_datetime = naive_datetime.replace(tzinfo=timezone.utc)
+
+        :param instance_name: name of the server
+        :type instance_name: str
+        :param start_time_dt: start time of the silence event
+        :type start_time_dt: datetime
+        :param end_time_dt: end time of the silence event
+        :type end_time_dt: datetime
+        :param author: entity creating the silence event
+        :type author: str
+        :param comment: explanation for the silence event
+        :type comment: str
+        :param silence_id: the ID for a valid silence event
+        :type comment: str
+        :param status: the status of a valid silence event (pending, expired, active)
+        :type status: str
+        """
+        # we want to ensure start_time_dt and end_time_dt
+        # are really instances of type datetime
+        # and not just strings
+        if start_time_dt is not None and not isinstance(start_time_dt, datetime):
+            raise TypeError("start_time_dt must be of type datetime.")
+        if end_time_dt is not None and not isinstance(end_time_dt, datetime):
+            raise TypeError("end_time_dt must be of type datetime.")
+        self.instance_name = instance_name
+        self.start_time_dt = start_time_dt
+        self.end_time_dt = end_time_dt
+        self.author = author
+        self.comment = comment
+        self.silence_id = silence_id
+        self.status = status
+
+    @property
+    def is_active(self):
+        """
+        check if the current date and time is between the start time
+        and the end time of the silence event
+
+        :return: True now is between the start time and the end time
+        :rtype: bool
+        """
+        return self.status == "active"
+
+    @property
+    def is_pending(self):
+        """
+        :return: True if silence status is pending, False otherwise.
+        :rtype: bool
+        """
+        return self.status == "pending"
+
+    @property
+    def has_expired(self):
+        """
+        :return: True if silence status is expired, False otherwise.
+        :rtype: bool
+        """
+        return self.status == "expired"
+
+    @property
+    def is_valid(self):
+        """
+        :return: True if the silence has not finished yet
+        :rtype: bool
+        """
+        return self.status in ["active", "pending"]
+
+    @property
+    def start_time_str(self):
+        """
+        converts the content of self.start_time_dt to a string with format
+        "2025-01-22T11:50:00Z"
+
+        :return: the converted start date
+        :rtype: str
+        """
+        return self.start_time_dt.strftime("%Y-%m-%dT%H:%M:%SZ")
+
+    @property
+    def end_time_str(self):
+        """
+        converts the content of self.end_time_dt to a string with format
+        "2025-01-22T11:50:00Z"
+
+        :return: the converted start date
+        :rtype: str
+        """
+        return self.end_time_dt.strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+class SilenceDetailsHandler(list):
+    """
+    class to host and handle multiple SilenceDetails objects
+    """
+
+    def add_from_json(self, json_doc_l):
+        """
+        create one or more new SilenceDetails objects with info
+        from a JSON doc.
+        This JSON doc can be, for example, the output of querying the
+        alertmanager for all existing silence events.
+
+        The JSON looks like this
+
+            [
+                {
+                    "id": "949fec12-0e14-492d-b2de-c08cf10e65a5",
+                    "status": {
+                        "state": "expired"
+                    },
+                    "updatedAt": "2025-01-16T07:31:31.257Z",
+                    "comment": "Testing.",
+                    "createdBy": "admin",
+                    "endsAt": "2025-01-16T10:51:00.000Z",
+                    "matchers": [
+                        {
+                            "isEqual": true,
+                            "isRegex": false,
+                            "name": "instance",
+                            "value": "hv123.matrix.net"
+                        }
+                    ],
+                    "startsAt": "2025-01-16T10:50:00.000Z"
+                },
+                {
+                    "id": "29da0e56-3aee-4b92-b9fb-abf551224f34",
+                    "status": {
+                        "state": "expired"
+                    },
+                    "updatedAt": "2025-01-16T07:32:25.792Z",
+                    "comment": "Testing.",
+                    "createdBy": "admin",
+                    "endsAt": "2025-01-16T10:51:00.000Z",
+                    "matchers": [
+                        {
+                            "isEqual": true,
+                            "isRegex": false,
+                            "name": "instance",
+                            "value": "hv123.matrix.net"
+                        }
+                    ],
+                    "startsAt": "2025-01-16T10:50:00.000Z"
+                }
+            ]
+
+        :param json_doc_l: JSON document with specifications for SilenceDetails instances
+        :type json_doc_l: str
+        """
+        # Convert JSON string to a list of dictionaries if needed
+        if isinstance(json_doc_l, str):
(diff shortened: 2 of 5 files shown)


lib/alertmanager_api/silence.py
```diff
@@ -0,0 +1,185 @@
...
+                    f"Response text: {req_ex.response.text}"
+                )
+            self.log.critical(msg)
+            raise req_ex
+
+    def remove_silences(self, silence_l):
+        """
+        Removes a list of previously scheduled silences in alertmanager
+
+        :param silence_l: a list of silence instances
```
meoflynn, line 157: Could we call this param something like `silence_instances`?

tests/lib/alertmanager_api/test_remove_silence.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    alertmanager_account = MagicMock()
+    class_instance = AlertManagerAPI(alertmanager_account)
+    date_start = datetime(2025, 1, 1, 10, 0, 0)
+    date_end = datetime(2025, 1, 1, 10, 0, 0)
+    silence = SilenceDetails("hostname", date_start, date_end, "author", "comment")
+    try:
+        class_instance.remove_silence(silence)
+    except MissingMandatoryParamError as miss_ex:
+        assert "Missing silence silence_id" in str(miss_ex)
```
meoflynn, line 39: This test is looking more at the specific message instead of checking whether the `MissingMandatoryParamError` is being raised first in the `remove_silence` method. It might be better for testing whether errors are return using `pytest.raises([exceptiontype])` to test the error is raised

tests/lib/alertmanager_api/test_remove_silence.py
```diff
@@ -0,0 +1,82 @@
...
+    mock_response = MagicMock()
+    mock_response.status_code = 201
+    mock_response.text = "rc is 201"
+    mock_post.return_value = mock_response
+    try:
+        class_instance.remove_silence(silence)
+    except RequestException as req_ex:
+        assert f"Failed to delete silence with ID {silence.silence_id}" in str(req_ex)
+        assert "Response status code: 201" in str(req_ex)
+        assert "Response text: rc is 201" in str(req_ex)
```
meoflynn, line 63: This test is looking more at the specific message instead of checking whether the exception is being raised in the first place in the method. It might be better for testing whether errors are return using `pytest.raises([exceptiontype])` to test the error is raised

tests/lib/alertmanager_api/test_remove_silence.py
```diff
@@ -0,0 +1,82 @@
...
+    class_instance = AlertManagerAPI(alertmanager_account)
+    date_start = datetime(2025, 1, 1, 10, 0, 0)
+    date_end = datetime(2025, 1, 1, 10, 0, 0)
+    silence = SilenceDetails("hostname", date_start, date_end, "author", "comment")
+    silence.silence_id = "silence_id"
+    mock_post.side_effect = RequestException("An error occurred")
+    try:
+        class_instance.remove_silence(silence)
+    except RequestException as req_ex:
+        assert "An error occurred" in str(req_ex)
```
meoflynn, line 82: This test is looking more at the specific message instead of checking whether the exception is being raised in the first place in the method. It might be better for testing whether errors are return using `pytest.raises([exceptiontype])` to test the error is raised

tests/lib/alertmanager_api/test_schedule_silence.py
```diff
@@ -0,0 +1,198 @@
...
+    mock_response = MagicMock()
+    mock_response.status_code = 201
+    mock_response.text = "rc is 201"
+    mock_post.return_value = mock_response
+    try:
+        class_instance.schedule_silence(silence)
+    except RequestException as req_ex:
+        assert "Failed to create silence. " in str(req_ex)
+        assert "Response status code: 201" in str(req_ex)
+        assert "Response text: rc is 201" in str(req_ex)
```
meoflynn, line 178: This test is looking more at the specific message instead of checking whether the exception is being raised in the first place in the method. It might be better for testing whether errors are return using `pytest.raises([exceptiontype])` to test the error is raised

tests/lib/alertmanager_api/test_schedule_silence.py
```diff
@@ -0,0 +1,198 @@
...
+        alertmanager_account
+    )
+    date_start = datetime(2025, 1, 1, 10, 0, 0)
+    date_end = datetime(2025, 1, 1, 10, 0, 0)
+    silence = SilenceDetails("hostname", date_start, date_end, "author", "comment")
+    mock_post.side_effect = RequestException("An error occurred")
+    try:
+        class_instance.schedule_silence(silence)
+    except RequestException as req_ex:
+        assert "An error occurred" in str(req_ex)
```
meoflynn, line 198: This test is looking more at the specific message instead of checking whether the exception is being raised in the first place in the method. It might be better for testing whether errors are return using `pytest.raises([exceptiontype])` to test the error is raised

meoflynn (changes requested):

lib/alertmanager_api/silence.py
```diff
@@ -0,0 +1,185 @@
...
+        try:
+            api_url = f"{self.api_endpoint}/api/v2/silences"
+            response = requests.post(
+                api_url,
+                auth=self.auth,
+                headers={"Accept": "application/json"},
+                json=payload,
+                timeout=10,
+            )
+            if response.status_code != 200:
```
gmatthews20, line 91: Is 200 the expected status code instead of 201? Could use `response.raise_for_status()` instead?

lib/alertmanager_api/silence.py
```diff
@@ -0,0 +1,185 @@
...
+                raise RequestException(msg, response=response)
+        except RequestException as req_ex:
+            msg = f"Failed to delete silence in Alertmanager: {req_ex} "
+            if req_ex.response is not None:
+                msg += (
+                    f"Response status code: {req_ex.response.status_code} "
+                    f"Response text: {req_ex.response.text}"
+                )
+            self.log.critical(msg)
+            raise req_ex
```
gmatthews20, line 151: Would something like this work here and also for schedule_downtime above?
```suggestion
            response.raise_for_status()
        except HTTPError as req_ex:
            msg = f"Failed to delete silence in Alertmanager: {req_ex} "
            if req_ex.response is not None:
                msg += (
                    f"Failed to delete silence with ID {silence.silence_id}. "
                    f"Response status code: {response.status_code} "
                    f"Response text: {response.text}"
                )
            self.log.critical(msg)
            raise req_ex
```

lib/alertmanager_api/silence.py
```diff
@@ -0,0 +1,185 @@
...
+        :rtype: SilenceDetailsHandler
+        """
+        try:
+            api_url = f"{self.api_endpoint}/api/v2/silences"
+            response = requests.get(api_url, auth=self.auth, timeout=10)
+            silences = response.json()
+            handler = SilenceDetailsHandler()
+            handler.add_from_json(silences)
+            return handler
+        except RequestException as req_ex:
```
gmatthews20, line 177: Does requests.get raise a RequestExceptions here?

tests/lib/alertmanager_api/test_remove_silence.py
```diff
@@ -0,0 +1,82 @@
...
+from requests.exceptions import RequestException
+from exceptions.missing_mandatory_param_error import MissingMandatoryParamError
+from alertmanager_api.silence_details import SilenceDetails
+from alertmanager_api.silence import AlertManagerAPI
+
+
+@patch("alertmanager_api.silence.requests.delete")
+def test_remove_silence_sucess(mock_post):
+    """
+    use case: the call to schedule_silence() works fine
```
gmatthews20, line 12:
```suggestion
    use case: the call to remove_silence() works fine
```

gmatthews20 (changes requested):

lib/alertmanager_api/silence.py
jose-caballero: sometimes

tests/lib/alertmanager_api/test_schedule_silence.py
```diff
@@ -0,0 +1,195 @@
...
+        mock_critical.assert_called_once_with("Missing silence author")
+
+
+def test_missing_attribute_comment(
+    class_instance, missing_comment
+):  #  pylint: disable=redefined-outer-name
+    with patch.object(class_instance.log, "critical") as mock_critical:
+        with pytest.raises(MissingMandatoryParamError, match="Missing silence comment"):
+            class_instance.schedule_silence(missing_comment)
+        mock_critical.assert_called_once_with("Missing silence comment")
```
lizsalmon, line 135:
```suggestion
@pytest.fixture
def silence_details(start_time_dt, end_time_dt):
    """Base fixture for SilenceDetails."""
    return SilenceDetails(
        instance_name="name",
        start_time_dt=start_time_dt,
        end_time_dt=end_time_dt,
        comment="comment",
        author="author",
    )

@pytest.mark.parametrize(
    "missing_attr, expected_message",
    [
        ("instance_name", "Missing silence instance name"),
        ("start_time_dt", "Missing silence start time"),
        ("end_time_dt", "Missing silence end time"),
        ("author", "Missing silence author"),
        ("comment", "Missing silence comment"),
    ]
)
def test_missing_mandatory_fields(class_instance, silence_details, missing_attr, expected_message):
    """Test missing required attributes in SilenceDetails."""
    setattr(silence_details, missing_attr, None)  # Set missing attribute to None

    with patch.object(class_instance.log, "critical") as mock_critical:
        with pytest.raises(MissingMandatoryParamError, match=expected_message):
            class_instance.schedule_silence(silence_details)

        mock_critical.assert_called_once_with(expected_message)

```
You could remove all of these similar tests by using `pytest.mark.parametrize` and having one base fixture for the silence_details.

tests/lib/alertmanager_api/test_schedule_silence.py
```diff
@@ -0,0 +1,195 @@
...
+
```
lizsalmon, line 136: There are also a few lines that are not covered from the codecov report that you could cover in a similar way. 
``` python
@pytest.mark.parametrize(
    "attr, value, expected_message",
    [
        ("start_time_dt", "not_dt", "start_time_dt must be of type datetime."),
        ("end_time_dt", "not_dt", "end_time_dt must be of type datetime."),
    ]
)
def test_invalid_datetime_fields(attr, value, expected_message):
    """Test that SilenceDetails raises a TypeError for invalid datetime fields."""
    kwargs = {
        "instance_name": "name",
        "start_time_dt": datetime(2025, 1, 1, 10, 0, 0),
        "end_time_dt": datetime(2025, 1, 1, 10, 0, 0),
        "author": "author",
        "comment": "comment",
    }
    kwargs[attr] = value  # Set the invalid attribute

    with pytest.raises(TypeError, match=expected_message):
        SilenceDetails(**kwargs)
```

lib/alertmanager_api/silence_details.py
```diff
@@ -0,0 +1,265 @@
...
+        :param status: the status of a valid silence event (pending, expired, active)
+        :type status: str
+        """
+        # we want to ensure start_time_dt and end_time_dt
+        # are really instances of type datetime
+        # and not just strings
+        if start_time_dt is not None and not isinstance(start_time_dt, datetime):
+            raise TypeError("start_time_dt must be of type datetime.")
+        if end_time_dt is not None and not isinstance(end_time_dt, datetime):
+            raise TypeError("end_time_dt must be of type datetime.")
```
meoflynn, line 49: I think these are missing in the unit test, could we add a unit test for these two cases to check the error is raised?

meoflynn (changes requested):
I spotted one part of the silence_details class that might be missing a unit test, would it be possible to re-run the codecov action to check if there are any other unit test gaps?

lib/alertmanager_api/silence_details.py
gmatthews20, line 1: Most of this file is missing unit tests can you add them please

lib/alertmanager_api/silence.py
gmatthews20, line 1: Most of the except blocks aren't covered by unit tests can you add them please

gmatthews20 (changes requested):

lib/alertmanager_api/silence_details.py
gmatthews20: You can see the missing coverage here: https://app.codecov.io/gh/stfc/st2-cloud-pack/pull/304?src=pr&el=tree&filepath=lib%2Falertmanager_api%2Fsilence_details.py&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=stfc#diff-bGliL2FsZXJ0bWFuYWdlcl9hcGkvc2lsZW5jZV9kZXRhaWxzLnB5

lib/alertmanager_api/silence.py
gmatthews20: You can see the missing coverage here: https://app.codecov.io/gh/stfc/st2-cloud-pack/pull/304?src=pr&el=tree&filepath=lib%2Falertmanager_api%2Fsilence.py&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=stfc#diff-bGliL2FsZXJ0bWFuYWdlcl9hcGkvc2lsZW5jZS5weQ==

gmatthews20:
Could you also update the example config here: [stackstorm_openstack.yaml.example](https://github.com/stfc/st2-cloud-pack/blob/main/stackstorm_openstack.yaml.example), with the alert manager account details

lizsalmon (changes requested):

tests/lib/alertmanager_api/test_silence_details.py
```diff
@@ -0,0 +1,213 @@
...
+def test_add_from_json_with_string():
+    """Test that add_from_json correctly parses a JSON string into a list of dictionaries."""
+
+    json_input = '[{"id": "123",\
+"startsAt": "2025-01-16T10:50:00.000Z",\
+"endsAt": "2025-01-16T10:51:00.000Z",\
+"createdBy": "admin",\
+"comment": "Testing.",\
+"status": {"state": "active"},\
+"matchers": [{"name": "instance", "value": "hv123.matrix.net"}]}]'
```
gmatthews20, line 63: We can improve coverage by adding this:
```suggestion
"matchers": [{"name": "env", "value": "prod"}, {"name": "instance", "value": "hv123.matrix.net"}]}]'
```

tests/lib/alertmanager_api/test_silence_details.py
```diff
@@ -0,0 +1,213 @@
...
+
+    json_list = [
+        {
+            "id": "456",
+            "startsAt": "2025-01-16T10:50:00.000Z",
+            "endsAt": "2025-01-16T10:51:00.000Z",
+            "createdBy": "admin",
+            "comment": "Testing.",
+            "status": {"state": "pending"},
+            "matchers": [{"name": "instance", "value": "hv123.matrix.net"}],
```
gmatthews20, line 100: We can improve coverage by adding this:
```suggestion
            "matchers": [{"name": "env", "value": "prod"}, {"name": "instance", "value": "hv123.matrix.net"}],
```

lib/alertmanager_api/silence.py
```diff
@@ -0,0 +1,182 @@
...
+            raise MissingMandatoryParamError(msg)
+        if not silence.comment:
+            msg = "Missing silence comment"
+            self.log.critical(msg)
+            raise MissingMandatoryParamError(msg)
+
+        payload = {
+            "matchers": [
+                {
+                    "name": "instance_name",
```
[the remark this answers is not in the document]
jose-caballero: fixed

tests/lib/alertmanager_api/test_silence_details.py
jose-caballero: Hi George. 
Is that needed? The actual code to create silences has this:

 69         payload = {
 70             "matchers": [
 71                 {
 72                     "name": "instance",
 73                     "value": silence.instance_name,
 74                     "isRegex": False,
 75                 }
 76             ],
 77             "startsAt": silence.start_time_str,
 78             "endsAt": silence.end_time_str,
 79             "createdBy": silence.author,
 80             "comment": silence.comment,
 81         }

There is no "env". 
Should I add "env" to the code (and then to the unit tests)?

lizsalmon (dismissed):

khalford (approved):

gmatthews20 (approved):

lib/alertmanager_api/silence.py
```diff
@@ -0,0 +1,198 @@
...
+            self.log.critical(msg)
+            raise MissingMandatoryParamError(msg)
+        if not silence.author:
+            msg = "Missing silence author"
+            self.log.critical(msg)
+            raise MissingMandatoryParamError(msg)
+        if not silence.comment:
+            msg = "Missing silence comment"
+            self.log.critical(msg)
+            raise MissingMandatoryParamError(msg)
```
anish-mudaraddi, line 67:
```suggestion
    required_fields = {
        'instance_name': 'silence instance name',
        'start_time_dt': 'silence start time',
        'end_time_dt': 'silence end time', 
        'author': 'silence author',
        'comment': 'silence comment'
    }
    
    for field, description in required_fields.items():
        if not getattr(silence, field):
            msg = f"Missing {description}"
            self.log.critical(msg)
            raise MissingMandatoryParamError(msg)
```

khalford (approved):

lib/alertmanager_api/silence.py
anish-mudaraddi, line 1: The name of this file might need changing - silence_api? Fine for now

lib/alertmanager_api/silence.py
anish-mudaraddi, line 67: unnecessary - this function is doing too many things 

validating the values of `silence` dataclass should happen when the object is created - separation of concerns. See my comment on making a dataclass in `silence_details.py`

lib/alertmanager_api/silence.py
```diff
@@ -0,0 +1,198 @@
...
+        try:
+            api_url = f"{self.api_endpoint}/api/v2/silences"
+            response = requests.post(
+                api_url,
+                auth=self.auth,
+                headers={"Accept": "application/json"},
+                json=payload,
+                timeout=10,
+            )
+            response.raise_for_status()
```
anish-mudaraddi, line 91: only this part should be in the try-catch block - limit the try block to only those calls that we expect to error to occur in - its just good practice

lib/alertmanager_api/silence.py
```diff
@@ -0,0 +1,198 @@
...
+                    f"Response status code: {req_ex.response.status_code} "
+                    f"Response text: {req_ex.response.text}"
+                )
+            self.log.critical(msg)
+            raise req_ex
+        except Exception as req_ex:
+            # to capture any other Exception that is not HTTPError
+            msg = f"Failed to create silence in Alertmanager: {req_ex} "
+            self.log.critical(msg)
+            raise req_ex
```
anish-mudaraddi, line 109: catching generic exceptions is a bad idea - this will catch `ctrl-c` errors for instance. 
Also logging that the silence creation fails doesn't add any extra value that we don't already get from looking at the python error traceback

lib/alertmanager_api/silence.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def remove_silence(self, silence):
```
anish-mudaraddi, line 111: why do we need to provide silence details in order to remove the silence - surely we only need to provide ID?

lib/alertmanager_api/silence.py
```diff
@@ -0,0 +1,198 @@
...
+        :param silence: object including the silence_id to be removed
+        :type silence: SilenceDetails
+        :raises MissingMandatoryParamError: when the silence_id is missing
+        :raises requests.exceptions.RequestException:
+            when the request to the AlertManager failed
+        """
+        if not silence.silence_id:
+            msg = "Missing silence silence_id"
+            self.log.critical(msg)
+            raise MissingMandatoryParamError(msg)
```
anish-mudaraddi, line 124: this confuses me - I think your SilenceDetails Dataclass is doing too many things 
1. its being used to create silences 
2. its being used to store existing silences

These two jobs require different data-structures - trying to use one for everything is unnecessary and complex

lib/alertmanager_api/silence.py
```diff
@@ -0,0 +1,198 @@
...
+            response.raise_for_status()
+        except requests.HTTPError as req_ex:
+            msg = f"Failed to delete silence in Alertmanager: {req_ex} "
+            if req_ex.response is not None:
+                msg += (
+                    f"Failed to delete silence with ID {silence.silence_id}. "
+                    f"Response status code: {response.status_code} "
+                    f"Response text: {response.text}"
+                )
+            self.log.critical(msg)
```
anish-mudaraddi, line 142: pylint complains about using f-strings in logs 
Also this needs to be `req_ex.response.text` etc - or else you'll get a reference before assignment error

lib/alertmanager_api/silence.py
```diff
@@ -0,0 +1,198 @@
...
+        """
+        download all silence events recorded in AlertManager
+
+        :return: the list of Silence events currently recorded in AlertManager
+        :rtype: SilenceDetailsHandler
+        """
+        try:
+            api_url = f"{self.api_endpoint}/api/v2/silences"
+            response = requests.get(api_url, auth=self.auth, timeout=10)
+            response.raise_for_status()
```
anish-mudaraddi, line 170: only this part needs to be in the try-catch block

lib/alertmanager_api/silence.py
```diff
@@ -0,0 +1,198 @@
...
+            silences = response.json()
+            handler = SilenceDetailsHandler()
+            handler.add_from_json(silences)
+            return handler
+        except requests.HTTPError as req_ex:
```
anish-mudaraddi, line 175: what's the purpose of having 2 different except statements for requests.HTTPError and RequestException if they both log the same message? We can combine into one

lib/alertmanager_api/silence.py
```diff
@@ -0,0 +1,198 @@
...
+        except RequestException as req_ex:
+            msg = f"Failed to get silences from Alertmanager: {req_ex} "
+            if req_ex.response is not None:
+                msg += (
+                    f"Response status code: {req_ex.response.status_code} "
+                    f"Response text: {req_ex.response.text}"
+                )
+            self.log.critical(msg)
+            raise req_ex
+        except Exception as req_ex:
```
anish-mudaraddi, line 194: generic exception catching is bad idea

lib/alertmanager_api/silence_details.py
```diff
@@ -0,0 +1,265 @@
+import json
+from datetime import datetime
+
+
+class SilenceDetails:  # pylint: disable=too-few-public-methods
```
anish-mudaraddi, line 5: this should be a dataclass

lib/alertmanager_api/silence_details.py
```diff
@@ -0,0 +1,265 @@
...
+    def __init__(
+        self,
+        instance_name=None,
```
anish-mudaraddi, line 8: what's this for - hypervisors? 
Makes this SilenceDetails struct only useful for hypervisor related silences - lets remove this to make the struct more generic - so we can use it for silencing other types of alerts in the future

lib/alertmanager_api/silence_details.py
```diff
@@ -0,0 +1,265 @@
...
+        converts the content of self.end_time_dt to a string with format
+        "2025-01-22T11:50:00Z"
+
+        :return: the converted start date
+        :rtype: str
+        """
+        return self.end_time_dt.strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+class SilenceDetailsHandler(list):
```
anish-mudaraddi, line 116: This is way too overengineered imo - :sweat_smile:  It's going to be too complex to maintain for such a simple purpose

I'd make a separate standalone function to parse json string from `get_silences` 
1. this will likely be the only place we'll need to ever do that - so make it a private method
2. make separate get_vaild_silences,  get_active_silences and get_hv_silences functions that will call get_silences and filter out the correct results

lib/alertmanager_api/silence.py
```diff
@@ -0,0 +1,198 @@
+import logging
+import requests
+from exceptions.missing_mandatory_param_error import MissingMandatoryParamError
+from requests.auth import HTTPBasicAuth
+from requests.exceptions import RequestException
+from alertmanager_api.silence_details import SilenceDetailsHandler
+
+
+class AlertManagerAPI:
```
anish-mudaraddi, line 9: Don't think you get anything out of making this a class - just convert your CRUD methods into standalone functions which take `AlertmanagerAccount` as an extra input - allows us to call the function directly from the actions/sensor layer without needing to create an extra middle `workflows` layer

tests/lib/alertmanager_api/test_silence_details.py
anish-mudaraddi, line 1: this needs to go in `tests/lib/structs/alertmanager/test_silence_details`

tests/lib/alertmanager_api/test_silence.py
```diff
@@ -0,0 +1,366 @@
...
+    start_time_dt, end_time_dt
+):  #  pylint: disable=redefined-outer-name
+    """Base fixture for SilenceDetails."""
+    return SilenceDetails(
+        instance_name="name",
+        start_time_dt=start_time_dt,
+        end_time_dt=end_time_dt,
+        comment="comment",
+        author="author",
+    )
```
anish-mudaraddi, line 39: you aren't using these fixtures?

tests/lib/alertmanager_api/test_silence.py
```diff
@@ -0,0 +1,366 @@
...
+
+    with patch.object(class_instance.log, "critical") as mock_critical:
+        with pytest.raises(HTTPError):
+            class_instance.schedule_silence(silence)
+
+        # Verify that the log message was correctly formatted
+        expected_message = (
+            f"Failed to create silence in Alertmanager: {mock_response.raise_for_status.side_effect} "
+            f"Failed to create silence. Response status code: 500 Response text: Internal Server Error"
+        )
```
anish-mudaraddi, line 112: I wouldn't bother testing that the contents of the logging message 
1. way too much work!
2. log message doesn't affect the functionality of the code
3. makes the test brittle - what if we want to fix a typo in the logs? Doubles the work for no gain really

anish-mudaraddi (commented):
since this is merged I have made a PR that fixes all these identified problems

PR #305, opened 2025-01-24
Build test server

gmatthews20:
### Description:

- Added create a server action
- Added hypervisor test build server action, builds either a single flavor or all available flavor on a given hypervisor

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/hv.create.test.server.yaml
+++ b/actions/hv.create.test.server.yaml
@@ -1,0 +1,34 @@
+---
+description: Create a new server
+enabled: true
+entry_point: src/openstack_actions.py
+name: hv.create.test.server
+parameters:
+  timeout:
+    default: 5400
+  lib_entry_point:
+    default: workflows.hv_create_test_server.create_test_server
+    immutable: true
+    type: string
+  requires_openstack:
+    default: true
+    immutable: true
+    type: boolean
+  cloud_account:
+    description: "The clouds.yaml account to use whilst performing this action"
+    required: true
+    type: string
+    default: "dev"
+    enum:
+      - "dev"
+      - "prod"
+  hypervisor_name:
+    type: string
+    required: true
+    description: Name of server
+  test_all_flavors:
+    type: boolean
+    description: Test all avaliable flavors for the hypervisor, otherwise test random flavor
+    default: false
+    required: true
+runner_type: python-script
--- a/actions/hv.create.test.server.yaml
+++ b/actions/hv.create.test.server.yaml
@@ -1,5 +1,5 @@
 ---
-description: Create a new server
+description: Build a test server on a given hypervisor, can optionally test all possible flavors avaliable to the hypervisor
 enabled: true
 entry_point: src/openstack_actions.py
 name: hv.create.test.server
--- a/lib/openstack_api/openstack_hypervisor.py
+++ b/lib/openstack_api/openstack_hypervisor.py
@@ -1,4 +1,4 @@
-from typing import Dict
+from typing import Dict, List
 
 from enums.hypervisor_states import HypervisorState
 from openstack.connection import Connection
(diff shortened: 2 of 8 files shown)


lib/openstack_api/openstack_hypervisor.py
```diff
@@ -46,3 +48,21 @@ def valid_state(state) -> bool:
     if not isinstance(hypervisor_server_count, int) or hypervisor_server_count < 0:
         return False
     return True
+
+
+def get_available_flavors(conn: Connection, hypervisor_name: str):
```
meoflynn, line 53: Could we add a docstring for this method?

lib/openstack_api/openstack_server.py
```diff
@@ -57,3 +58,40 @@ def snapshot_server(conn: Connection, server_id: str) -> Image:
     conn.image.update_image(image, owner=project_id)
 
     return image
+
+
+def build_server(
```
meoflynn, line 63: Could we add a docstring for this method and the method below?

lib/workflows/hv_create_test_server.py
```diff
@@ -0,0 +1,22 @@
+import random
+
+from openstack.connection import Connection
+from openstack_api.openstack_hypervisor import get_available_flavors
+from openstack_api.openstack_server import build_server, delete_server
+
+
+def create_test_server(conn: Connection, hypervisor_name: str, test_all_flavors: bool):
```
meoflynn, line 8: Could we add a docstring to this method?

tests/lib/openstack_api/test_openstack_server.py
```diff
@@ -146,3 +151,74 @@ def test_block_gpu_migration():
             dest_host=None,
             snapshot=True,
         )
+
+
+def test_build_server():
```
meoflynn, line 156: Could we add a docstring to this unit test and the two unit tests below?

tests/lib/workflows/test_hv_create_test_server.py
meoflynn, line 1: Could we add docstrings to these unit tests?

meoflynn (changes requested):

actions/hv.create.test.server.yaml
```diff
@@ -0,0 +1,34 @@
+---
+description: Create a new server
+enabled: true
```
Dmitry-Popovichev, line 3: Should this be something like `Creates a new test server` so that it differentiates a little from the other action you wrote?

lib/openstack_api/openstack_server.py
```diff
@@ -57,3 +58,70 @@ def snapshot_server(conn: Connection, server_id: str) -> Image:
...
+    :type conn: Connection
+    :param server_id: ID of server to delete
+    :type server_id: str
+    :param force: Option to force delete server
+    :type force: bool
+    :return: None
+    :rtype: None
+    """
+    server = conn.compute.find_server(server_id)
+    conn.compute.delete_server(server, force)
```
Dmitry-Popovichev, line 125: Do we need to `force` delete the server? If necessary, that's fair enough but wondering if there's a softer way of deleting i.e. if we set that bool to false.

gmatthews20: The default value for the force delete parameter is false, so if its not specifically set to true it won't force delete

actions/hv.create.test.server.yaml
gmatthews20: Yeah will update this

Dmitry-Popovichev (approved):

meoflynn (approved):

PR #307, opened 2025-02-05
Move JiraConnection to a separate file

jose-caballero:
For clarity, move the class JiraConnection to its own module. This way, it is more clear the purpose of the module jira_issue.py, just to create new tickets.


gmatthews20 (changes requested):
Could you also move the jira connection unit test to their own file as well to match this

gmatthews20 (approved):

PR #309, opened 2025-02-05
Fix matcher

jose-caballero:
### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/alertmanager_api/silence_details.py
+++ b/lib/alertmanager_api/silence_details.py
@@ -5,7 +5,7 @@ from datetime import datetime
 class SilenceDetails:  # pylint: disable=too-few-public-methods
     def __init__(
         self,
-        instance_name=None,
+        hostname=None,
         start_time_dt=None,
         end_time_dt=None,
         author=None,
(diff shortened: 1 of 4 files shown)


anish-mudaraddi:
PR - https://github.com/stfc/st2-cloud-pack/pull/311 - will make this unecessary

anish-mudaraddi (approved):

lib/alertmanager_api/silence_details.py
```diff
@@ -196,16 +196,16 @@ def add_from_json(self, json_doc_l):
...
             comment = json_doc["comment"]
 
-            # Extract the instance name from the matchers list (look for a 'name' == 'instance')
+            # Extract the hostname name from the matchers list (look for a 'name' == 'hostname')
             for matcher in json_doc.get("matchers", []):
-                if matcher["name"] == "instance":
-                    instance_name = matcher["value"]
+                if matcher["name"] == "hostname":
+                    hostname = matcher["value"]
                     break
```
anish-mudaraddi, line 203: actually I'm not sure this works. 
As @lizsalmon  - We have 2 alerts - one which stores hv name under `matchers.name = "instance"`` and one that has `matchers.name = "hostname"` - we need to grab both silences - so you need to keep the previous one and add elif

anish-mudaraddi (changes requested):

PR #311, opened 2025-02-05
Simplify Alertmanager API

anish-mudaraddi:
various changes to make the alertmanager API easier to use and more extendible
fixes various comments made here: https://github.com/stfc/st2-cloud-pack/pull/304#pullrequestreview-2596190435


lizsalmon (dismissed):

jose-caballero:
I have a few comments/questions:

1. I thought that **typing.List** and **typing.Dict** were deprecated, or in the process of being deprecated. Is that not true? If they are, or will be, is still worth it using them? 
2. Even though the functions are now using type annotation, could be make the docstrings fully Sphinx compliant? Lines for param type and return type are missing.
3. Question: how are matchers being created and provided by the client code? I only see in the Details class matchers as a list attribute. That forces the client to understand the matchers and create the list itself. 
4. Now the functions **get_active_silences( )** / **get_valid_silences( )** are disconnected from **get_silences( )**. The function that is more relevant to the client is now **get_hv_silences( )**, which only calls **get_silences( )**. The other two are never being called. We can leave them for now, they are harmless, but there is currently no filter to only remove silences that are "valid". 
5. Question: we don't unit test **silence_details.py** anymore? 

Extra:

There are a couple of pieces of code that, even though they work, I find them a little bit obscure. If that is OK with you, I would like to refactor them if a future PR, and make them more clear. The result will be less elegant, but I believe easier to follow. Examples:

- the creation of dictionaries in **get_active_silences( )** and **get_valid_silences( )**
- the entire code of **check_matchers( )**

anish-mudaraddi:
1. I didn't realize this - looks like its been deprecated since python3.9. I'll change this here. This needs to be changed throughout the codebase so I'll do that in a future PR
2. Yes, agreed A lot of functions in this entire codebase are not Sphinx compliant - another PR is necessary
3. Good Question! Right now the client/user must create matcher dictionary containing the appropriate keys - I'll have a think about a better way for them to enter in that info without needing to understand Alertmanager syntax
4. True, I just copied them in from the previous PR - does `get_hv_silences( )` only need to get "valid" silences or all silences? At the moment it gets all silences - I don't see the harm in removing "expired" silences though
5. We do - I just moved it into `tests/structs/test_silence_details.py`

Extra: sure if you think that's best! I tend to go a bit overboard with list/dict comprehensions to limit the number of indentations!

jose-caballero (dismissed):
There is further work to be done, but this is OK for the time being.

jose-caballero (approved):

lizsalmon (approved):

PR #319, opened 2025-02-18
ENH: Migration retries

gmatthews20:
### Description:

- Retry failed migrations 3 times, with a 2 min delay (max stackstorm allows) between each attempt
- Set an TTL for hypervisor states to retry (after 2 weeks by default) any actions for a given state in an attempt to unstuck the hypervisor

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/rules/hv.maintenance.notification.yaml
+++ b/rules/hv.maintenance.notification.yaml
@@ -6,8 +6,8 @@ enabled: true
 
 criteria:
   trigger.previous_state:
-    type: equals
-    pattern: PENDING_MAINTENANCE
+    type: inside
+    pattern: [null, PENDING_MAINTENANCE]
   trigger.current_state:
     type: equals
     pattern: DRAINING
--- a/config.schema.yaml
+++ b/config.schema.yaml
@@ -138,3 +138,6 @@ hypervisor_sensor:
     uptime_limit:
       type: "number"
       description: "Number of uptime days before marking hypervisor for maintenance"
+    state_expire_after:
+      type: "integer"
+      description: "Number of seconds before forcefully deleting previous state to update hypervisors stuck in a state"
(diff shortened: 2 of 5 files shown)


rules/hv.maintenance.notification.yaml
```diff
@@ -6,8 +6,8 @@ enabled: true
 
 criteria:
   trigger.previous_state:
-    type: equals
-    pattern: PENDING_MAINTENANCE
+    type: inside
+    pattern: [null, PENDING_MAINTENANCE]
```
lizsalmon, line 10: What is this null for?

gmatthews20: When the state expires after set TTL, the previous state becomes null. I have updated this rule so that it will re attempt a hypervisor drain after the state expires

lizsalmon: So if a migration fails it will retry 3 times (2 min delay) and if still stuck in draining after two weeks it's previous state will be put as null which will trigger the notification/migrate action again?

gmatthews20: Yeah so it will try each server migration up to 3 times if they fail, then after 2 weeks it will attempt to drain the hypervisor again if it hasn't drained naturally in those 2 weeks

lizsalmon: Okay yeah that makes sense! Thank you

lizsalmon (dismissed):

lizsalmon:
There are conflicts that need resolving I think

lizsalmon (approved):

jose-caballero:
Hi George, 
One of the commit messages says something about "hypervisor state KVs". I don't know what a KV is. Would it be possible to make the commit message more clear for beginners?

jose-caballero (approved):

PR #320, opened 2025-02-18
Bug fixes from testing

lizsalmon:
### Description:

This includes all bug fixes that were found during testing.  I have not included changing the polling sensor or the specific hypervisor limitations that we added in. 

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/alertmanager_api/silence.py
+++ b/lib/alertmanager_api/silence.py
@@ -130,7 +130,7 @@ def get_silences(alertmanager_account: AlertManagerAccount) -> dict:
         raise req_ex
 
     silences = {}
-    for silence in json.loads(response.json()):
+    for silence in response.json():
         # Create SilenceDetails instance
         silence_details = SilenceDetails(
             # Convert ISO format strings to datetime objects
@@ -223,12 +223,11 @@ def get_hv_silences(alertmanager_account: AlertManagerAccount, hostname: str):
     # [("name1", "value1"), ("name2", "value2")] - must match these 2 matchers (logical AND)
     # [("name3", "value3"), ("name4", "value4")] - OR this other set of matchers
     hv_matcher_conditions = [
-        [("alertname", "InstanceDown"), ("instance", hostname)],
-        [("alertname", "PrometheusTargetMissing"), ("instance", hostname)],
-        [("alertname", "OpenStackServiceDown"), ("hostname", hostname)],
+        [("instance", hostname)],
+        [("hostname", hostname)],
     ]
 
-    def check_matchers(matchers: AlertMatcherDetails) -> bool:
+    def check_matchers(matchers: List[AlertMatcherDetails]) -> bool:
         """
         check alert matchers against a set of conditions to see if they are ones pertaining to HVs
         :param matchers: list of AlertMatcherDetails objects
--- a/lib/alertmanager_api/silence.py
+++ b/lib/alertmanager_api/silence.py
@@ -1,6 +1,7 @@
 import logging
 import json
 from datetime import datetime
+from typing import List
 
 import requests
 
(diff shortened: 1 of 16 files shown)


lib/alertmanager_api/silence.py
```diff
@@ -222,12 +222,11 @@ def get_hv_silences(alertmanager_account: AlertManagerAccount, hostname: str):
     # [("name1", "value1"), ("name2", "value2")] - must match these 2 matchers (logical AND)
     # [("name3", "value3"), ("name4", "value4")] - OR this other set of matchers
     hv_matcher_conditions = [
-        [("alertname", "InstanceDown"), ("instance", hostname)],
-        [("alertname", "PrometheusTargetMissing"), ("instance", hostname)],
-        [("alertname", "OpenStackServiceDown"), ("hostname", hostname)],
+        [("instance", hostname)],
```
lizsalmon, line 225: @gmatthews20 Why did we change this bit? It breaks the unittest? Or can I put it back to this: 
``` 
 [("alertname", "InstanceDown"), ("instance", hostname)],
 [("alertname", "PrometheusTargetMissing"), ("instance", hostname)],
 [("alertname", "OpenStackServiceDown"), ("hostname", hostname)],
```

lib/workflows/hv_post_reboot.py
```diff
@@ -29,12 +29,12 @@ def post_reboot(
         object_type=IcingaObject.HOST,
         object_name=hypervisor_hostname,
     )
-    create_test_server(
-        conn=conn, hypervisor_name=hypervisor_hostname, test_all_flavors=False
-    )
     hv_service_enable(
         conn=conn, hypervisor_name=hypervisor_hostname, service_binary="nova-compute"
     )
+    create_test_server(
```
Neonbluestoplight, line 35: For curiosity, why change the position of the function?

Neonbluestoplight (approved):
Asked question for curiosity but looks good to me!

lib/workflows/hv_post_reboot.py
lizsalmon: We discovered that in order to create a test server the hv needs to be enabled (which makes a lot of sense)

lib/openstack_api/openstack_server.py
```diff
@@ -34,10 +34,16 @@ def snapshot_and_migrate_server(
...
+        conn.compute.migrate_server(server=server_id, host=dest_host)
+        server = conn.compute.get_server(server_id)
+        conn.compute.wait_for_server(server, status="VERIFY_RESIZE")
+        conn.compute.confirm_server_resize(server_id)
+        return conn.compute.wait_for_server(server, status="SHUTOFF")
+    conn.compute.live_migrate_server(
         server=server_id, host=dest_host, block_migration=True
     )
+    server = conn.compute.get_server(server_id)
+    return conn.compute.wait_for_server(server, status="ACTIVE")
```
gmatthews20, line 46: We can move the `get_server` before the if statement to make it more tidy, something like this:
```suggestion
server = conn.compute.get_server(server_id)
if server_status == "SHUTOFF":
        conn.compute.migrate_server(server=server, host=dest_host)
        conn.compute.wait_for_server(server, status="VERIFY_RESIZE")
        conn.compute.confirm_server_resize(server)
        return conn.compute.wait_for_server(server, status="SHUTOFF")
    conn.compute.live_migrate_server(
        server=server, host=dest_host, block_migration=True
    )
    return conn.compute.wait_for_server(server, status="ACTIVE")
```

gmatthews20 (changes requested):

lib/workflows/hv_post_reboot.py
Neonbluestoplight: Makes sense yeah!

gmatthews20 (approved):

Neonbluestoplight (approved):

PR #322, opened 2025-02-21
Image share action

Neonbluestoplight:
### Description:

ENH: Added an action to share an image to a project when the image and project 

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/image.share.to.project.yaml
+++ b/actions/image.share.to.project.yaml
@@ -1,5 +1,5 @@
 ---
-description: Create a new server
+description: Share an image to a project
 enabled: true
 entry_point: src/openstack_actions.py
 name: image.share.to.project
--- a/actions/image.share.to.project.yaml
+++ b/actions/image.share.to.project.yaml
@@ -1,0 +1,31 @@
+---
+description: Create a new server
+enabled: true
+entry_point: src/openstack_actions.py
+name: image.share.to.project
+parameters:
+  timeout:
+    default: 5400
+  lib_entry_point:
+    default: openstack_api.openstack_image.share_image_to_project
+    immutable: true
+    type: string
+  requires_openstack:
+    default: true
+    immutable: true
+    type: boolean
+  cloud_account:
+    description: "The clouds.yaml account to use whilst performing this action"
+    required: true
+    type: string
+    default: "dev"
+    enum:
+      - "dev"
+      - "prod"
+  project_id:
+    required: true
+    type: string
+  image_id:
+    required: true
+    type: string
+runner_type: python-script
--- a/actions/image.share.to.project.yaml
+++ b/actions/image.share.to.project.yaml
@@ -22,10 +22,12 @@ parameters:
     enum:
       - "dev"
       - "prod"
-  project_id:
+  project_identifier:
+    description: "The name or ID of the project to share the image to"
     required: true
     type: string
-  image_id:
+  image_identifier:
+    description: "Name or ID of image to share"
     required: true
     type: string
 runner_type: python-script
--- a/lib/openstack_api/openstack_image.py
+++ b/lib/openstack_api/openstack_image.py
@@ -3,18 +3,20 @@ from openstack.connection import Connection
 
 def share_image_to_project(
     conn: Connection,
-    image_id: str,
-    project_id: str,
+    image_identifier: str,
+    project_identifier: str,
 ) -> None:
     """
     Shares given image with given project
 
     :param conn: openstack connection object
-    :param image_id: The image ID that is shared with project
-    :param project_id: Openstack ID for the project
+    :param image_identifier: The image name or ID that is shared with project
+    :param project_identifier: Project name or ID to share image to
     """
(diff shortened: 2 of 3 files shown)


lizsalmon (approved):

tests/lib/openstack_api/test_openstack_image.py
```diff
@@ -0,0 +1,26 @@
...
+from openstack_api.openstack_image import share_image_to_project
+
+
+def test_share_image_to_project():
+    """
+    Test sharing image to project
+    """
+    mock_connection = MagicMock()
+    mock_image_id = "test-image-id"
+    mock_project_id = "test-project-id"
```
Dmitry-Popovichev, line 11: These could be `NonCallableMagicMock()` by importing the `NonCallableMagicMock` module from unittest but I don't think it makes much of a difference. Either a string or a NonCallableMagicMock() works, just a suggestion.

Dmitry-Popovichev (commented):
Looks good to me, just a couple of curiosity points and the action description otherwise looks good

actions/image.share.to.project.yaml
```diff
@@ -0,0 +1,31 @@
+---
+description: Create a new server
```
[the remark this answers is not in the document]
Neonbluestoplight: Good catch thank you, will change

lib/openstack_api/openstack_image.py
```diff
@@ -0,0 +1,21 @@
...
+    :param conn: openstack connection object
+    :param image_id: The image ID that is shared with project
+    :param project_id: Openstack ID for the project
+    """
+    image = conn.image.find_image(image_id, ignore_missing=False)
+    destination_project = conn.identity.find_project(project_id, ignore_missing=False)
+
+    conn.image.add_member(image["id"], member_id=destination_project["id"])
+
+    conn.image.update_member(destination_project["id"], image["id"], status="accepted")
```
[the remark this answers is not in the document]
Neonbluestoplight: We're creating image and project objects by using `conn.image.find_image` and `conn.identity.find_project` but only technically require the id part of it which is why we specify it later. 

It could in theory be replaced by just the strings but we chose this to also maintain consistency with other files as well

As for `update_member`, as far as I understand it the image after being added isn't immediately accepted and update_member ensures that the image is accepted. Not having this in makes it so that the image is theoretically added but not actually visible in the Openstack project so this makes sure that it actually sends

tests/lib/openstack_api/test_openstack_image.py
Neonbluestoplight: I think @gmatthews20 mentioned that in this particular case, this makes the code a bit more readable but yes `NonCallableMagicMock()` could work thank you!

lib/openstack_api/openstack_image.py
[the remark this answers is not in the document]
lizsalmon: This is probably the best description of why `update_member` was needed: https://docs.openstack.org/api-ref/image/v2/index.html#sharing

Dmitry-Popovichev (approved):

actions/image.share.to.project.yaml
```diff
@@ -0,0 +1,31 @@
...
+    default: "dev"
+    enum:
+      - "dev"
+      - "prod"
+  project_id:
+    required: true
+    type: string
+  image_id:
+    required: true
+    type: string
```
gmatthews20, line 30:
```suggestion
  project_identifier:
    description: "The name or ID of the project to share the image to"
    required: true
    type: string
  image_identifier:
    description: "Name or ID of image to share"
    required: true
    type: string
```

lib/openstack_api/openstack_image.py
```diff
@@ -0,0 +1,21 @@
+from openstack.connection import Connection
+
+
+def share_image_to_project(
+    conn: Connection,
+    image_id: str,
+    project_id: str,
```
gmatthews20, line 7:
```suggestion
    image_identifier: str,
    project_identifier: str,
```

lib/openstack_api/openstack_image.py
```diff
@@ -0,0 +1,21 @@
...
+) -> None:
+    """
+    Shares given image with given project
+
+    :param conn: openstack connection object
+    :param image_id: The image ID that is shared with project
+    :param project_id: Openstack ID for the project
```
gmatthews20, line 14:
```suggestion
    :param image_identifier: The image name or ID that is shared with project
    :param project_identifier: Project name or ID to share image to
```

gmatthews20 (changes requested):

actions/image.share.to.project.yaml
Neonbluestoplight: Thank you for this, have changed it here and also changed `image_id` and `project_id` to image_identifier` and `project_identifier` to the testing file as well for consistency

gmatthews20 (approved):

PR #325, opened 2025-02-27
Add hypervisor downtime action

gmatthews20:
New action which is specific to hypervisors, sets icinga downtimes and silences in alertmanager for all alerts related to the hypervisor

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/hypervisor.downtime.yaml
+++ b/actions/hypervisor.downtime.yaml
@@ -10,8 +10,8 @@ parameters:
     required: true
   duration_hours:
     type: integer
-    description: Duration of flexible downtime in hours, CANNOT be a decimal
-    required: false
+    description: Duration of downtime in hours, CANNOT be a decimal
+    required: true
   comment:
     type: string
     description: Downtime reason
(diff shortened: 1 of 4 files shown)


jacob-ward:
Slightly unclear on the fixed vs flexible downtimes. I would have expected to have to set a period for the fixed downtime but this reads like the opposite, i.e. we'd need to set a period for flexible downtimes? Is my thinking incorrect or am I just reading something wrong?

gmatthews20:
> Slightly unclear on the fixed vs flexible downtimes. I would have expected to have to set a period for the fixed downtime but this reads like the opposite, i.e. we'd need to set a period for flexible downtimes? Is my thinking incorrect or am I just reading something wrong?

@jacob-ward The fixed/flexible down time is a icinga thing. https://icinga.com/docs/icinga-2/latest/doc/08-advanced-topics/#fixed-and-flexible-downtimes. I think alertmanager silences only has the equivalent of the fixed downtime so we can drop the flexible option.

We could simplify it so the user would only have to set a start and end time in the action and a duration, or start time and a duration, or start the downtime immediately and the use just sets a duration. I'm happy to use which ever is easiest for you

DaveW-STFC (approved):

jose-caballero:
Hi @gmatthews20 I see the action requires start time and end time, as mandatory parameters, and duration as optional one. What happens if the users passes the 3 of them, but they are not consistent?

jose-caballero (approved):

PR #328, opened 2025-02-27
Fixed the migration functions to wait for status

lizsalmon:
Tested in dev-deployment

### Description:

This PR adds checks in for the status of image and migration so that the actions in stackstorm will fail.

I am not sure why the code coverage is failing - I think the line is covered. 

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/openstack_api/openstack_server.py
+++ b/lib/openstack_api/openstack_server.py
@@ -1,10 +1,10 @@
 from datetime import datetime
+import time
 from typing import Optional
 from openstack.connection import Connection
 from openstack.compute.v2.image import Image
 from openstack.compute.v2.server import Server
-from openstack.exceptions import ResourceTimeout
-from openstack.exceptions import ResourceFailure
+from openstack.exceptions import ResourceFailure, ResourceTimeout
 
 
 def snapshot_and_migrate_server(
@@ -23,12 +23,10 @@ def snapshot_and_migrate_server(
     :param flavor_name: Server flavor name
     :param dest_host: Optional host to migrate to, otherwise chosen by scheduler
     """
-
     if flavor_id.startswith("g-"):
         raise ValueError(
             f"Attempted to move GPU flavor, {flavor_id}, which is not allowed!"
         )
-
     if server_status not in ["ACTIVE", "SHUTOFF"]:
         raise ValueError(
             f"Server status: {server_status}. The server must be ACTIVE or SHUTOFF to be migrated"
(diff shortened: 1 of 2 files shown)


lib/openstack_api/openstack_server.py
```diff
@@ -60,13 +59,57 @@ def snapshot_server(conn: Connection, server_id: str) -> Image:
...
+    :param image: The Image object
+    :param status: The status of the image that is required
+    :param interval:How long to wait between checks
+    :param timeout: Timeout of the function
+    """
+    if image.status == status:
+        return image
+    start_time = time.time()
+    while time.time() - start_time < timeout:
+        image = conn.image.get_image(image.id)
```
Neonbluestoplight, line 82: curiosity: why is this here and not at the start of the function before the if statement?

lizsalmon: Which bit sorry? the while?

Neonbluestoplight: Sorry I don't know why it showed it like this. I meant `image = conn.image.get_image(image.id)`

Neonbluestoplight: That part does make sense but there's an if statement (ln 78-79):
```
    if image.status == status:
        return image
```
wouldn't that require this as well or is that handled by the function parameter?

lizsalmon: No, so the actual function takes an image as a parameter which is like the initial one? Although the test coverage is failing so there is a good chance this is not working how I think it should work

Neonbluestoplight: Ok that's what I had guessed too thanks for the clarification!

tests/lib/openstack_api/test_openstack_server.py
```diff
@@ -157,6 +223,119 @@ def test_block_gpu_migration():
...
+    image = MagicMock(id="123", name="test-image", status="pending")
+    mock_conn.image.get_image.side_effect = itertools.cycle([image])
+    with pytest.raises(
+        ResourceTimeout,
+        match=f"Timeout waiting for image {image.name} to become active.",
+    ):
+        wait_for_image_status(mock_conn, image, "active", interval=0, timeout=0)
+
+
+def test_wait_for_image_status_immediate_success():
```
Neonbluestoplight, line 257: Since this covers the if statement at the beginning maybe in `test_wait_for_image_status_success()` you need to check if `mock_conn.image.get_image.assert_called_once_with()` to cover the if staement in the loop?

Neonbluestoplight (commented):
Everything looks good so am fine to approve it but have added some comments for my own learning and clarity

tests/lib/openstack_api/test_openstack_server.py
lizsalmon: I have added this thanks Nalin, this still does not fix the coverage report though :(

Neonbluestoplight (approved):

gmatthews20 (approved):

Neonbluestoplight (approved):

PR #329, opened 2025-02-28
Added new parameters to quota.set workflow

DaveW-STFC:
### Description:

Added new compute parameters to the quota.set workflow.
Changed method calls to pass the quotas as a dictionary instead of individual lines.
Updated the unit tests to reflect the changes.
Added start of quota.show methods for review.

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:


anish-mudaraddi (approved):

AlexCK-STFC (changes requested):
@DaveW-STFC
1. Is this PR still active? If so:

I resolved the formatting issues for you, but [unit tests are failing](https://github.com/stfc/st2-cloud-pack/actions/runs/17464155556/job/49595622761?pr=329).

```
FAILED tests/lib/workflows/test_create_project.py::test_setup_external_networking
FAILED tests/lib/workflows/test_create_project.py::test_setup_jasmin_networking
```

This is because you've updated the variables for `set_quota`, but the old ones are still being used in the [create_project workflow](https://github.com/stfc/st2-cloud-pack/blob/set_compute_quota/lib/workflows/create_project.py#L234).

Is there a reason why we can't set the defaults for the arguments of `quota_set` to be 0 making them optional arguments, rather than requiring them all to be set to 0 for the function to work? Right now, they're all required positional arguments

The `create_project` workflow needs updating either way.

Edit: Reading more, it seems it's set so 0 means no change. I find that quite confusing.

khalford (approved):

anish-mudaraddi (approved):

PR #330, opened 2025-02-28
Change issue status

jose-caballero:
New function in jira_api directory in stackstorm pack, to change the state of a given JIRA Issue.

### DIFF
--- a/lib/exceptions/jira.py
+++ b/lib/exceptions/jira.py
@@ -1,0 +1,30 @@
+class MismatchedState(Exception):
+    """
+    Exception to be raised when the JIRA Issue is explictly expected to be
+    in a different state that it actually is.
+    For example, when a transition is requested from state A to state B,
+    and the value A is being explicitly passed,
+    but the Issue is not in that state A.
+    """
+
+    def __init__(self, issue_key, current_state):
+        self.value = (
+            f'The JIRA Issue "{issue_key}" is currently in state "{current_state}"'
+        )
+
+    def __str__(self):
+        return repr(self.value)
+
+
+class ForbiddenTransition(Exception):
+    """
+    Exception to be raised when there is a desire to move a JIRA Issue
+    to a different state it current is in,
+    but that state is not reachable according that the Issue workflow.
+    """
+
+    def __init__(self, issue_key, to_state):
+        self.value = f'The JIRA Issue "{issue_key}" is not in any state allowing a transition to "{to_state}"'
+
+    def __str__(self):
+        return repr(self.value)
--- a/lib/jira_api/jira_issue.py
+++ b/lib/jira_api/jira_issue.py
@@ -1,5 +1,6 @@
 from typing import Optional
 from exceptions.missing_mandatory_param_error import MissingMandatoryParamError
+from exceptions.jira import MismatchedState, ForbiddenTransition
 from structs.jira.jira_account import JiraAccount
 from structs.jira.jira_issue_details import IssueDetails
 from jira_api.connection import JiraConnection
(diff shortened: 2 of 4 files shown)


meoflynn (changes requested):

lib/jira_api/jira_issue.py
```diff
@@ -59,3 +60,40 @@ def add_comment(
...
+        issue = conn.issue(issue_key)
+        if from_state and issue.fields.status.name != from_state:
+            raise InvalidState(issue_key, from_state)
+        allowed_transitions = conn.transitions(issue_key)
+        for transition in allowed_transitions:
+            if transition["to"]["name"] == to_state:
+                conn.transition_issue(issue_key, transition["id"])
+                break
+        else:
+            raise ForbiddenTransition(issue_key, to_state)
```
gmatthews20, line 99: Do we need to check before making the transition request that it is a valid transition? Can we let conn.transition_issue attempt to make the transition and raise its own exception if it fails?

lib/exceptions/jira.py
```diff
@@ -0,0 +1,16 @@
+class InvalidState(Exception):
```
gmatthews20, line 1: Just a thought on the naming, if this is just raised when the state doesn't match what is expected maybe something like `MismatchedState` becuase to me `InvalidState` suggests that the state doesn't exist or issue can never have that state

gmatthews20 (changes requested):

lib/jira_api/jira_issue.py
jose-caballero: I have been checking that possibility. Indeed, JIRA raises its own exception. It looks like this:

> JiraError HTTP 400 url: https://stfc.atlassian.net/rest/api/2/issue/DCTE-4/transitions
> 	text: Transition id '124' is not valid for this issue.
>
> 	response headers = {'Content-Type': 'application/json;charset=UTF-8', 'Transfer-Encoding': 'chunked', 'Connection': 'keep-alive', 'Date': 'Mon, 03 Mar 2025 09:55:52 GMT', 'Server': 'AtlassianEdge', 'Timing-Allow-Origin': '*', 'X-Arequestid': '3ef86f516d92e943cc0acb1168649004', 'X-Aaccountid': '610bbda1ec92ab0070e43a1b', 'Cache-Control': 'no-store, no-cache', 'X-Content-Type-Options': 'nosniff', 'X-Xss-Protection': '1; mode=block', 'Atl-Traceid': '8beb76e3667c4eb1bdc7d23d52c9324e', 'Atl-Request-Id': '8beb76e3-667c-4eb1-bdc7-d23d52c9324e', 'Strict-Transport-Security': 'max-age=63072000; includeSubDomains; preload', 'Report-To': '{"endpoints": [{"url": "https://dz8aopenkvv6s.cloudfront.net"}], "group": "endpoint-1", "include_subdomains": true, "max_age": 600}', 'Nel': '{"failure_fraction": 0.001, "include_subdomains": true, "max_age": 600, "report_to": "endpoint-1"}', 'Vary': 'Accept-Encoding', 'Content-Encoding': 'gzip', 'Server-Timing': 'atl-edge;dur=172,atl-edge-internal;dur=16,atl-edge-upstream;dur=156,atl-edge-pop;desc="aws-eu-central-1"', 'X-Cache': 'Error from cloudfront', 'Via': '1.1 71cbe01df9e5102d886edc4f5a32c1ea.cloudfront.net (CloudFront)', 'X-Amz-Cf-Pop': 'LHR50-P2', 'X-Amz-Cf-Id': 'yL_7TFoI_Vn4F0Fq7jY__t4NxDH0ElQ8YmRMSasD423kCaw97ak6UA=='}
> 	response text = {"errorMessages":["Transition id '124' is not valid for this issue."],"errors":{}}

So I guess it is safe to remove that ForbiddenTransition class.

jose-caballero: Apparently we cannot do that, as we cannot pass the name of the destination state as input option to method transition_issue( ) and let jira try it.

lib/exceptions/jira.py
```diff
@@ -0,0 +1,16 @@
...
+    def __init__(self, issue_key, current_state):
+        self.value = (
+            f'The JIRA Issue "{issue_key}" is currently in state "{current_state}"'
+        )
+
+    def __str__(self):
+        return repr(self.value)
+
+
+class ForbiddenTransition(Exception):
```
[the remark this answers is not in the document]
jose-caballero: actually, we still need the class, so I will do it

lib/exceptions/jira.py
jose-caballero: agree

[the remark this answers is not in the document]
jose-caballero: done

lib/exceptions/jira.py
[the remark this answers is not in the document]
jose-caballero: done

meoflynn (approved):

gmatthews20 (approved):

PR #336, opened 2025-03-07
Simplify unit test code

jose-caballero:
This unit test code grew organically as new functionalities have been added to the library lib/jira_api/
As a consequence, there was a lot of duplication.

This commit attempts to fix that situation:

* introduces fixtures to avoid code duplication: in particular, removes the need to use patch() and MagicMock() multiple times

* uses parametrization for handle similar tests:
  - search tests: condensed into a single test with different requirements
  - task creation tests: handle missing fields and Epic ID variations
  - commment tests: covers internal vs external comments
  - state change tests: covers valid and invalid transitions in one function

Hopefully this will make easier to add more tests for future functionalities.

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/tests/lib/jira_api/test_jira_issue.py
+++ b/tests/lib/jira_api/test_jira_issue.py
@@ -11,250 +11,141 @@ from jira_api.jira_issue import (
 )
 
 
-def test_search_issues_no_requirements():
-    """
-    Tests search_issues with no additional requirements
-    """
+@pytest.fixture(name="mock_jira")
+def mock_jira_client():
     with patch("jira.client.JIRA") as mock_conn:
-        mock_account = MagicMock()
-        mock_conn = mock_conn.return_value
-        mock_conn.search_issues.return_value = ["ISSUE-1", "ISSUE-2"]
-
-        # Call the function
-        result = search_issues(mock_account, "TEST_PROJECT")
-
-        # Assertions
-        mock_conn.search_issues.assert_called_once_with("project = TEST_PROJECT")
-        assert result == ["ISSUE-1", "ISSUE-2"]
+        yield mock_conn.return_value
 
 
-def test_search_issues_with_requirements():
-    """
-    Tests search_issues with multiple requirements
-    """
-    with patch("jira.client.JIRA") as mock_conn:
-        mock_account = MagicMock()
-        mock_conn = mock_conn.return_value
-        mock_conn.search_issues.return_value = ["ISSUE-3", "ISSUE-4"]
-
-        # Call the function
-        result = search_issues(
-            mock_account, "TEST_PROJECT", ["status = Open", "assignee = currentUser()"]
-        )
-
-        # Assertions
-        jql = "project = TEST_PROJECT AND status = Open AND assignee = currentUser()"
-        mock_conn.search_issues.assert_called_once_with(jql)
-        assert result == ["ISSUE-3", "ISSUE-4"]
+@pytest.fixture(name="mock_account")
+def mock_jira_account():
+    return MagicMock()
 
 
-def test_search_issues_empty_results():
-    """
-    Tests search_issues when no issues are found
-    """
-    with patch("jira.client.JIRA") as mock_conn:
-        mock_account = MagicMock()
-        mock_conn = mock_conn.return_value
-        mock_conn.search_issues.return_value = []
-
-        # Call the function
-        result = search_issues(mock_account, "TEST_PROJECT", ["status = Done"])
-
-        # Assertions
-        mock_conn.search_issues.assert_called_once_with(
-            "project = TEST_PROJECT AND status = Done"
-        )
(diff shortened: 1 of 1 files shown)


Barfy007JB1 (dismissed):

tests/lib/jira_api/test_jira_issue.py
```diff
@@ -11,250 +11,141 @@
...
-    """
-    Tests exception is raised when missing description
-    """
-    mock_issue_details = MagicMock()
-    mock_issue_details.project_id = "foo"
-    mock_issue_details.summary = "bar"
-    mock_issue_details.description = ""
-
-    mock_account = MagicMock()
+        yield mock_conn.return_value
```
khalford, line 17: Why yield here if you are only returning one object? Is return not better suited?

tests/lib/jira_api/test_jira_issue.py
```diff
@@ -11,250 +11,141 @@
...
+    "requirements, expected_issues, expected_query",
+    [
+        (None, ["ISSUE-1", "ISSUE-2"], "project = TEST_PROJECT"),
+        (
+            ["status = Open", "assignee = currentUser()"],
+            ["ISSUE-3", "ISSUE-4"],
+            "project = TEST_PROJECT AND status = Open AND assignee = currentUser()",
+        ),
+        (["status = Done"], [], "project = TEST_PROJECT AND status = Done"),
+    ],
```
khalford, line 45: This is only mock data so it is not critical. But the first tuple shows a query for every issue in the project so should return 1,2 and 3,4

tests/lib/jira_api/test_jira_issue.py
```diff
@@ -11,250 +11,141 @@
...
+@pytest.mark.parametrize(
+    "comment_text, internal_flag",
+    [("This is a test comment", True), ("Public comment", False)],
+)
+def test_add_comment(mock_jira, mock_account, comment_text, internal_flag):
+    """Tests add_comment with different internal flag values"""
+    add_comment(mock_account, "ISSUE-123", comment_text, internal=internal_flag)
+    mock_jira.add_comment.assert_called_once_with(
+        "ISSUE-123", comment_text, internal_flag
+    )
```
khalford, line 97: I think renaming the variables this way is clearer.
`comment_text -> comment` and `internal_flag -> is_internal`
That way:
comment = "This is a comment"
is_internal = True

```suggestion
@pytest.mark.parametrize(
    "comment, is_internal",
    [("This is a test comment", True), ("Public comment", False)],
)
def test_add_comment(mock_jira, mock_account, comment, is_internal):
    """Tests add_comment with normal and internal comments."""
    add_comment(mock_account, "ISSUE-123", comment, internal=is_internal)
    mock_jira.add_comment.assert_called_once_with(
        "ISSUE-123", comment, is_internal
    )
```

tests/lib/jira_api/test_jira_issue.py
```diff
@@ -11,250 +11,141 @@
...
+    "current_state, target_state, transitions, expected_exception",
+    [
+        (
+            "Open",
+            "In Progress",
+            [
+                {"id": "21", "to": {"name": "In Progress"}},
+                {"id": "31", "to": {"name": "Closed"}},
+            ],
+            None,
```
khalford, line 116: I don't understand this one. If the current state is `open` and we want to go to `todo` why would we need to go through the transition `closed`?

khalford (commented):
Looks much better. Just a few questions about naming

tests/lib/jira_api/test_jira_issue.py
jose-caballero: inline comments added to the code explaining why

tests/lib/jira_api/test_jira_issue.py
jose-caballero: I changed ISSUE-3, ISSUE-4 to ISSUE-1, ISSUE-2 to avoid a possible misinterpretation that the numbers are relevant and there is semantic content in the fact they are different in each tuple.

tests/lib/jira_api/test_jira_issue.py
jose-caballero: fair enough. Changed in the code.

tests/lib/jira_api/test_jira_issue.py
jose-caballero: nope. The values in the 3rd position are the entire list of allowed transitions from the current state. That doesn't mean we want to go thru them. Let me try to figure out how to explain that in a inline comment.

jose-caballero: add some extra explanation to the docstring of the actual function

khalford (approved):

PR #339, opened 2025-03-12
Jira new project issue class

khalford:
### Description:
Adds a JIRA issue base class and new project issue class. These classes serialise data from the proforma forms used in the JIRA portal. That way we can access the data much more easily.

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/jira_api/issue_types/new_project_request.py
+++ b/lib/jira_api/issue_types/new_project_request.py
@@ -1,0 +1,37 @@
+from typing import Dict
+from jira_api.issue_types.issue_base import IssueBase
+
+
+class NewProjectRequestIssue(IssueBase):  # pylint: disable=too-few-public-methods
+    """Issue class to handle new project requests"""
+
+    def _create_properties(self) -> Dict:
+        properties = {}
+        proforma_form = [
+            prop
+            for prop in self.conn.issue_properties(self.issue_key)
+            if prop.key == "proforma.forms.i1"
+        ][0]
+        answers = proforma_form.value.state.answers
+
+        properties["project_name"] = getattr(answers, "3").text
+        properties["users"] = getattr(answers, "29").adf.content[0].content[0].text
+        properties["cpus"] = int(getattr(answers, "32").text)
+        properties["vms"] = int(getattr(answers, "45").text)
+        properties["memory"] = int(getattr(answers, "33").text)
+        properties["shared_storage"] = int(getattr(answers, "34").text)
+        properties["object_storage"] = int(getattr(answers, "35").text)
+        properties["block_storage"] = int(getattr(answers, "44").text)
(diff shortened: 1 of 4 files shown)


lib/jira_api/issue_types/new_project_request.py
```diff
@@ -0,0 +1,37 @@
...
+        properties["users"] = getattr(answers, "29").adf.content[0].content[0].text
+        properties["cpus"] = int(getattr(answers, "32").text)
+        properties["vms"] = int(getattr(answers, "45").text)
+        properties["memory"] = int(getattr(answers, "33").text)
+        properties["shared_storage"] = int(getattr(answers, "34").text)
+        properties["object_storage"] = int(getattr(answers, "35").text)
+        properties["block_storage"] = int(getattr(answers, "44").text)
+        properties["gpus"] = getattr(answers, "17").adf.content[0].content[0].text
+        properties["contact_email"] = getattr(answers, "25").text
+        properties["reporting_fed_id"] = getattr(answers, "42").text
```
Dmitry-Popovichev, line 27: Do we need to get all of these properties? Since the StackStorm action only requires `project_email`, `project_name`, `project_description` from the form and this script is specifically for `new_project_request`.

At the same time, it can't hurt to store them either. Interested in your thoughts?

tests/lib/jira_api/test_new_project_request.py
```diff
@@ -0,0 +1,218 @@
...
+    assert res["memory"] == 50
+    assert res["shared_storage"] == 0
+    assert res["object_storage"] == 0
+    assert res["block_storage"] == 0
+    assert res["gpus"] == "Big powerful GPUs because they’re fun"
+    assert res["contact_email"] == "mock_user@mock_email"
+    assert res["reporting_fed_id"] == "mock_fedid"
+    # Assert specifically for "True" here as
+    assert res["externally_accessible"] is False
+    assert res["tos_agreement"] is True
```
Dmitry-Popovichev, line 218: Is that comment meant to say "False"? And should the tos_agreement be "False" too? If not should we test for a false tos or are they just not able to submit the form at all if that's the case (I think the latter).

khalford: Sorry this comment is not clear. I have removed the comment. What it originally meant was to assert the value "is" the boolean value of True and not that it evaulates to True.

lib/jira_api/issue_types/new_project_request.py
khalford: We need all of these values to set the quota of the project when we create it. I don't know how StackStorm creates projects but we do need this information

Dmitry-Popovichev (approved):

PR #340, opened 2025-03-12
Jira automations skeleton

jose-caballero:
Create an skeleton for the new set of Automations to perform changes in OpenStack triggered by JIRA Issues.

In this PR, we are only adding the skeleton, but with no actual implementation for real Actions. Not yet. This is some for of placeholder for it. From now on, the implementation of the actual Automations, including the first one, can be done in incremental steps instead of all of it in a single big bang.

### DIFF
--- a/actions/jira.request_new_project.yaml
+++ b/actions/jira.request_new_project.yaml
@@ -1,0 +1,32 @@
+---
+description: Creates a new Project in OpenStack. This requires an associated JIRA ticket, for manual Project creation see project.create
+enabled: true
+entry_point: src/openstack_actions.py
+name: jira.request_new_project
+parameters:
+  lib_entry_point:
+    type: string
+    default: workflows.jira_openstack_request_new_project.request_new_project
+    immutable: true
+  requires_openstack:
+    default: true
+    immutable: true
+    type: boolean
+  cloud_account:
+    description: The clouds.yaml account to use whilst performing this action
+    required: true
+    type: string
+    default: "dev"
+    enum:
+      - "dev"
+      - "prod"
+  jira_account_name:
+    type: string
+    description: Name of JIRA Account to use. Must be configured in the pack settings.
+    required: true
+    default: "default"
+  issue_key:
+    type: string
+    description: The unique ID to identify each JIRA Issue
+    required: true
+runner_type: python-script
--- a/lib/workflows/jira_openstack_request_new_project.py
+++ b/lib/workflows/jira_openstack_request_new_project.py
@@ -1,9 +1,11 @@
 from structs.jira.jira_account import JiraAccount
 
+# pylint: disable=unused-argument
+
 
 def request_new_project(
-    jira_account: JiraAccount,  # pylint: disable=unused-argument
-    issue_key,  # pylint: disable=unused-argument
+    jira_account: JiraAccount,
+    issue_key,
 ):
     """
     to be implemented
--- a/lib/workflows/jira_openstack_request_new_project.py
+++ b/lib/workflows/jira_openstack_request_new_project.py
@@ -1,0 +1,11 @@
+from structs.jira.jira_account import JiraAccount
+
+
+def request_new_project(
+    jira_account: JiraAccount,  # pylint: disable=unused-argument
+    issue_key,  # pylint: disable=unused-argument
+):
+    """
+    to be implemented
+    """
+    raise NotImplementedError
--- a/sensors/jira.request_new_project.yaml
+++ b/sensors/jira.request_new_project.yaml
@@ -1,0 +1,15 @@
+---
+class_name: JiraIssueSensor
+entry_point: src/jira_issue_sensor.py
+description: Search for new JIRA issues
+poll_interval: 300
+enabled: true
+trigger_types:
+  - name: "jira.request_new_project"
+    description: "Triggers when there is a new JIRA issue requesting the creation of a new OpenStack project"
+    payload_schema:
+      type: "object"
+      properties:
+        issue_key:
+          type: "string"
+          description: "unique ID to identify the JIRA Issue with the needed data for the creation of a new OpenStack project"
--- a/sensors/src/jira_issue_sensor.py
+++ b/sensors/src/jira_issue_sensor.py
@@ -1,0 +1,86 @@
+from st2reactor.sensor.base import PollingSensor
+from jira_api.jira_issue import search_issues
+from structs.jira.jira_account import JiraAccount
+import jira
+
+# pylint: disable=attribute-defined-outside-init
+
+
+class JiraIssueSensor(PollingSensor):
+    """
+    class to implement an active Sensor to query a JIRA project
+    for specific types of JIRA tickets (issues) and perform
+    the corresponding actions via a StackStorm Trigger.
+
+    Active sensors poll a remote service periodically
+    (like a cron) instead of waiting for an event input.
+    """
+
+    def setup(self):
+        """
+        set the connection object for our JIRA instance
+        """
+
+        # At sensor startup, self.config holds the pack’s config
+        if not self.config:
+            raise ValueError("No config found for sensor")
+        jira_account_name = self.config["jira_account_name"]
+        self.jira_account = JiraAccount.from_pack_config(self.config, jira_account_name)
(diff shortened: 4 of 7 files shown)


sensors/src/jira_issue_sensor.py
```diff
@@ -0,0 +1,69 @@
...
+        check JIRA for new tickets to process
+
+        We only search for JIRA tickets in the "TO DO" status.
+        This way, once this automation starts working on them and set them
+        as "In Progress", they won't be picked up again.
+        """
+        # different types of JIRA issues:
+        request_type_list = [
+            "Request New Project",
+            "Add User",
```
DavidFair, line 45: I would use a dict, it removes having to do string manipulations and allows us to de-couple the request type from Jira to the action name:

```python
# Maps from Jira ticket types to ST2 workflow names 
ticket_map: Dict[str, str] = {
    "Request New Project": "jira.request_new_project",
    "Add User": "jira.add_user"
}

target_project = "STFCCLOUD"
issues_list = search_issues(JiraAccount, target_project, ticket_map.keys())
```

sensors/src/jira_issue_sensor.py
```diff
@@ -0,0 +1,69 @@
...
+                JiraAccount,
+                "STFCCLOUD",
+                requirements_list,
+            )
+            for issue in issues_list:
+                # e.g. "Request New Project" -> "request_new_project"
+                trigger_name = f"jira.{request_type_label}"
+                self.sensor_service.dispatch(
+                    trigger=trigger_name, payload={"issue_key": issue.key}
+                )
```
DavidFair, line 63: Then this can simply do:

```python
self.sensor_service.dispatch(trigger=ticket_map[issue.label], payload=...)
```

sensors/src/jira_issue_sensor.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def cleanup(self):
+        """
+        clean up
+        """
+        self.jira_client.close()
```
DavidFair, line 69: Should check for null / none in-case we failed to setup or we cleanup before setup (me just applying very defensive programming):
```python

if self.jira_client:
   self.jira_client.close()
# Unconditionally drop the resource in-case something tries to use it between now and __exit__
self.jira_client = None  
```

As we're not in `__del__` or `__exit__` so we can't be sure we're the final finalizer

tests/sensors/test_jira_issue_sensor.py
```diff
@@ -0,0 +1,105 @@
...
+    # Setup mock JIRA issues that would be returned by search_issues
+    mock_issue_1 = MagicMock(key="ISSUE-101")
+    mock_issue_2 = MagicMock(key="ISSUE-202")
+    # The sensor calls search_issues for multiple request types:
+    # ["Request New Project", "Add User", ...].
+    # We'll just pretend both calls return the same set of issues.
+    mock_search_issues.return_value = [mock_issue_1, mock_issue_2]
+
+    # Run sensor setup, which initializes self.jira_client
+    sensor.setup()
```
DavidFair, line 47: This test should be split into multiple cases since anything changing just results in `FAILED test_poll`

Instead if we had:

```python
def test_sensor_setup(sensor):
    sensor.setup()
    some_mock.assert_called_once_with(....)


def test_sensor_polling_with_no_returned_issues(sensor)
    sensor.setup()
    # We don't need asserts here since it's covered in the other case

    should_not_appear_one = JiraIssue(...)
    mock_search_issues.return_value = [should_not_appear_one]
    sensor.poll(...)
    poll(...)

def test_sensor_polling_with_expected_issues(sensor)
    sensor.setup()
    # We don't need asserts here since it's covered in the other case

    mock_search_issues.return_value = ....
    sensor.poll(...)
    poll(...)
```


It makes it really obvious if you see:

```
FAILED test_sensor_setup
PASSED test_sensor_polling_with_no_returned_issues
FAILED test_sensor_polling_with_expected_issues
```

Which parts of the code you need to fix

DavidFair: In addition your missing asserts / checks for your setup code

tests/sensors/test_jira_issue_sensor.py
```diff
@@ -0,0 +1,105 @@
...
+    sensor.sensor_service.dispatch.assert_any_call(
+        trigger="jira.add_user", payload={"issue_key": "ISSUE-202"}
+    )
+
+    # Optionally, ensure total dispatch calls match the expected count:
+    # 2 issues * 2 request types = 4 dispatches
+    assert sensor.sensor_service.dispatch.call_count == 4
+
+    # If you want to confirm the JIRA client was correctly created in setup():
+    mock_jira_client.assert_called_once()
```
DavidFair, line 105: Ah actually it's hidden at the bottom, even more reason for a dedicated test case :wink:

DavidFair (changes requested):

sensors/src/jira_issue_sensor.py
```diff
@@ -0,0 +1,69 @@
+from st2reactor.sensor.base import PollingSensor
+from jira_api.jira_issue import search_issues
+from structs.jira.jira_account import JiraAccount
+import jira
+
+
+class JiraIssueSensor(PollingSensor):
+    """
+    class to implement an active Sensor to query a JIRA project
+    for very specific type of JIRA tickets (issues) and perform
```
khalford, line 10: More specific than specific?
```suggestion
    for specific type of JIRA tickets (issues) and perform
```

khalford, line 10: More specific than specific?
```suggestion
    for specific type of JIRA tickets (issues) and perform
```

sensors/src/jira_issue_sensor.py
```diff
@@ -0,0 +1,69 @@
...
+            )
+        )
+
+    def poll(self):
+        """
+        check JIRA for new tickets to process
+
+        We only search for JIRA tickets in the "TO DO" status.
+        This way, once this automation starts working on them and set them
+        as "In Progress", they won't be picked up again.
```
khalford, line 40: As discussed
```suggestion
        We only search for JIRA tickets in the "Ready for Automation" status.
        This way, once this automation starts working on them and set them
        as "ST2 Working", they won't be picked up again.
```

actions/jira.request_new_project.yaml
```diff
@@ -0,0 +1,32 @@
+---
+description: Create new Project in OpenStack
```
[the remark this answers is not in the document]
jose-caballero: done

lib/workflows/jira_openstack_request_new_project.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def request_new_project(
+    jira_account: JiraAccount,  # pylint: disable=unused-argument
+    issue_key,  # pylint: disable=unused-argument
+):
+    """
+    to be implemented
+    """
+    return True
```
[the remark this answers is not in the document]
jose-caballero: done

sensors/src/jira_issue_sensor.py
jose-caballero: done

sensors/src/jira_issue_sensor.py
jose-caballero: I have done slightly different, but the idea is the same.

sensors/src/jira_issue_sensor.py
jose-caballero: uber specific
fixed

sensors/src/jira_issue_sensor.py
jose-caballero: fixed

sensors/jira.request_new_project.yaml
```diff
@@ -0,0 +1,15 @@
+---
+class_name: JiraIssueSensor
+entry_point: src/jira_issue_sensor.py
+description: Search for new JIRA issues
+poll_interval: 60
```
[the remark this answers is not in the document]
jose-caballero: fixed

sensors/src/jira_issue_sensor.py
jose-caballero: done

gmatthews20 (changes requested):

PR #341, opened 2025-03-17
Jira automations skeleton

jose-caballero:
Create an skeleton for the new set of Automations to perform changes in OpenStack triggered by JIRA Issues.

In this PR, we are only adding the skeleton, but with no actual implementation for real Actions. Not yet. This is some for of placeholder for it. From now on, the implementation of the actual Automations, including the first one, can be done in incremental steps instead of all of it in a single big bang.

### DIFF
--- a/actions/jira.request_new_project.yaml
+++ b/actions/jira.request_new_project.yaml
@@ -1,0 +1,32 @@
+---
+description: Creates a new Project in OpenStack. This requires an associated JIRA ticket, for manual Project creation see project.create
+enabled: true
+entry_point: src/openstack_actions.py
+name: jira.request_new_project
+parameters:
+  lib_entry_point:
+    type: string
+    default: workflows.jira_openstack_request_new_project.request_new_project
+    immutable: true
+  requires_openstack:
+    default: true
+    immutable: true
+    type: boolean
+  cloud_account:
+    description: The clouds.yaml account to use whilst performing this action
+    required: true
+    type: string
+    default: "dev"
+    enum:
+      - "dev"
+      - "prod"
+  jira_account_name:
+    type: string
+    description: Name of JIRA Account to use. Must be configured in the pack settings.
+    required: true
+    default: "default"
+  issue_key:
+    type: string
+    description: The unique ID to identify each JIRA Issue
+    required: true
+runner_type: python-script
--- a/sensors/src/jira_issue_sensor.py
+++ b/sensors/src/jira_issue_sensor.py
@@ -1,0 +1,86 @@
+from st2reactor.sensor.base import PollingSensor
+from jira_api.jira_issue import search_issues
+from structs.jira.jira_account import JiraAccount
+import jira
+
+# pylint: disable=attribute-defined-outside-init
+
+
+class JiraIssueSensor(PollingSensor):
+    """
+    class to implement an active Sensor to query a JIRA project
+    for specific types of JIRA tickets (issues) and perform
+    the corresponding actions via a StackStorm Trigger.
+
+    Active sensors poll a remote service periodically
+    (like a cron) instead of waiting for an event input.
+    """
+
+    def setup(self):
+        """
+        set the connection object for our JIRA instance
+        """
+
+        # At sensor startup, self.config holds the pack’s config
+        if not self.config:
+            raise ValueError("No config found for sensor")
+        jira_account_name = self.config["jira_account_name"]
+        self.jira_account = JiraAccount.from_pack_config(self.config, jira_account_name)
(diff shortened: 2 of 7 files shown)


actions/jira.request_new_project.yaml
gmatthews20, line 1: One for the future, This could potentially be a use case which is better suited for a Orquesta workflow rather than a new action in `lib/workflows`

sensors/src/jira_issue_sensor.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+
+        # At sensor startup, self.config holds the pack’s config
+        if not self.config:
+            raise ValueError("No config found for sensor")
+        jira_account_name = self.config["jira_account_name"]
+        self.jira_account = JiraAccount.from_pack_config(self.config, jira_account_name)
+        endpoint = self.jira_account.atlassian_endpoint
+        token = self.jira_account.api_token
+        username = self.jira_account.username
```
gmatthews20, line 31: These might be better defined in init and then create the jira client in setup()

sensors/src/jira_issue_sensor.py
```diff
@@ -0,0 +1,86 @@
...
+
+        We only search for JIRA tickets in the "Ready For Automation" state.
+        This way, once this automation starts working on them and set them
+        as "Automation In Progress", they won't be picked up again.
+        """
+        # different types of JIRA issues:
+        request_type_dict = {
+            "Request New Project": "jira.request_new_project",
+            "Add User": "jira.add_user",
+        }
```
gmatthews20, line 52: This should also probably go in init since it is constant and doesn't need redefining every time we poll

sensors/src/jira_issue_sensor.py
```diff
@@ -0,0 +1,86 @@
...
+        for request_type, trigger_name in request_type_dict.items():
+            requirements_list = [
+                'statusCategory in ("Ready For Automation")',
+                f'"Request Type" = "{request_type}"',
+            ]
+            issues_list = search_issues(
+                self.jira_account,
+                "STFCCLOUD",
```
gmatthews20, line 60: Can we set this as a variable in `sensors/jira.request_new_project.yaml` so that we can test using a different project

gmatthews20 (changes requested):

actions/jira.request_new_project.yaml
jose-caballero: I thought Anish said we don't use Orquesta, because it is hard to test. I understood as we don't use it at all.

sensors/src/jira_issue_sensor.py
jose-caballero: done

sensors/src/jira_issue_sensor.py
jose-caballero: done

khalford (dismissed):

actions/jira.request_new_project.yaml
gmatthews20: Myself and @anish-mudaraddi had a chat about it on Monday since I wanted to use an Orquesta workflow for a hypervisor drain action which was effectively just chaining together existing actions. We decided that Orquesta would be ok for simple workflows where there is no logic and no branching, e.g. chaining actions together

sensors/src/jira_issue_sensor.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def poll(self):
+        """
+        check JIRA for new tickets to process
+
+        We only search for JIRA tickets in the "Ready For Automation" state.
+        This way, once this automation starts working on them and set them
+        as "Automation In Progress", they won't be picked up again.
+        """
+        jira_project = self.config.get("jira_project", "STFCCLOUD")
```
gmatthews20, line 55: I'd don't think that we should default to the main jira queue, it'd be better to default to None or the test jira project
Also this can go in __init__

gmatthews20 (changes requested):

sensors/src/jira_issue_sensor.py
jose-caballero: fixed

actions/jira.request_new_project.yaml
jose-caballero: I see. l am interested. When you figure it out, including how to test it properly, let me know :)

khalford (approved):

actions/jira.request_new_project.yaml
jose-caballero: I will check it, thanks.

khalford (approved):

gmatthews20 (approved):

PR #343, opened 2025-03-19
Deal with empty fields in the form

jose-caballero:
When the user does not fill the boxes with some data, the attribute
    getattr(answers, "<field number>").adf.content
is just an empty list. Therefore, we cannot just try to get the value of the first index
    getattr(answers, "<field number>").adf.content[0]
as it does not exist.

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/jira_api/issue_types/new_project_request.py
+++ b/lib/jira_api/issue_types/new_project_request.py
@@ -15,14 +15,24 @@ class NewProjectRequestIssue(IssueBase):  # pylint: disable=too-few-public-metho
         answers = proforma_form.value.state.answers
 
         properties["project_name"] = getattr(answers, "3").text
-        properties["users"] = getattr(answers, "29").adf.content[0].content[0].text
+        potential_users_info = getattr(answers, "29").adf.content
+        if potential_users_info:
+            # the value is not an empty string
+            properties["users"] = potential_users_info[0].content[0].text
+        else:
+            properties["users"] = ""
         properties["cpus"] = int(getattr(answers, "32").text)
         properties["vms"] = int(getattr(answers, "45").text)
         properties["memory"] = int(getattr(answers, "33").text)
         properties["shared_storage"] = int(getattr(answers, "34").text)
         properties["object_storage"] = int(getattr(answers, "35").text)
         properties["block_storage"] = int(getattr(answers, "44").text)
-        properties["gpus"] = getattr(answers, "17").adf.content[0].content[0].text
+        potential_gpus_info = getattr(answers, "17").adf.content
+        if potential_gpus_info:
+            # the value is not an empty string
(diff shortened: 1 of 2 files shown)


lib/jira_api/issue_types/new_project_request.py
```diff
@@ -15,14 +15,22 @@ def _create_properties(self) -> Dict:
         answers = proforma_form.value.state.answers
 
         properties["project_name"] = getattr(answers, "3").text
-        properties["users"] = getattr(answers, "29").adf.content[0].content[0].text
+        potential_users_info = getattr(answers, "29").adf.content
+        if potential_users_info != []:
+            properties["users"] = potential_users_info[0].content[0].text
+        else:
+            properties["users"] = ""
```
khalford, line 22: An empty string evaulates to false so we can just use `not`.
```suggestion
        potential_users_info = getattr(answers, "29").adf.content
        if not potential_users_info:
            properties["users"] = potential_users_info[0].content[0].text
        else:
            properties["users"] = ""
```

khalford: Also, at some point we should look at how we are expecting users to fill in this part of the form. Right now it just returns a paragraph of text,. How will we split that? Would we use a delimiter like `,` or `;`. Not for this PR

lib/jira_api/issue_types/new_project_request.py
```diff
@@ -15,14 +15,22 @@ def _create_properties(self) -> Dict:
...
         properties["memory"] = int(getattr(answers, "33").text)
         properties["shared_storage"] = int(getattr(answers, "34").text)
         properties["object_storage"] = int(getattr(answers, "35").text)
         properties["block_storage"] = int(getattr(answers, "44").text)
-        properties["gpus"] = getattr(answers, "17").adf.content[0].content[0].text
+        potential_gpus_info = getattr(answers, "17").adf.content
+        if potential_gpus_info != []:
+            properties["gpus"] = potential_gpus_info[0].content[0].text
+        else:
+            properties["gpus"] = ""
```
khalford, line 33: Same as above.
```suggestion
        potential_gpus_info = getattr(answers, "17").adf.content
        if not potential_gpus_info:
            properties["gpus"] = potential_gpus_info[0].content[0].text
        else:
            properties["gpus"] = ""
```

khalford (changes requested):
Only one change

Dmitry-Popovichev (approved):
LGTM

lib/jira_api/issue_types/new_project_request.py
jose-caballero: done

PR #346, opened 2025-04-01
update test server action to use 22.04 ubuntu image

anish-mudaraddi:
20.04 is deprecated as its eol


jacob-ward (approved):
Any reason you've selected 22.04 and not 24.04?
Not a problem, just curious

DavidFair (approved):

PR #352, opened 2025-05-23
Adding Cloud ChatOps Sensor

khalford:
### Description:

Adds a sensor to trigger the ChatOps weekly reminders.

The sensor uses the [schedule](https://schedule.readthedocs.io/en/stable/) library to run events on specific days and times. When the schedule is triggered it makes a HTTP request to the ChatOps application with a token we provide in the pack config.

### Special Notes:

Adds extra variables to the pack config.

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/src/openstack_actions.py
+++ b/actions/src/openstack_actions.py
@@ -61,4 +61,11 @@ class OpenstackActions(Action):
                 self.config, kwargs["alertmanager_account_name"]
             )
             del kwargs["alertmanager_account_name"]
+
+        if "chatops_token" in kwargs:
+            kwargs["token"] = self.config["chatops_sensor"]["token"]
+            kwargs["endpoint"] = self.config["chatops_sensor"]["endpoint"]
+            kwargs["channel"] = self.config["chatops_sensor"]["channel"]
+            del kwargs["chatops_token"]
+
         return kwargs
--- a/actions/src/openstack_actions.py
+++ b/actions/src/openstack_actions.py
@@ -62,10 +62,11 @@ class OpenstackActions(Action):
             )
             del kwargs["alertmanager_account_name"]
 
-        if "reminder_type" in kwargs:
+        if "chatops_reminder_type" in kwargs:
             kwargs["token"] = self.config["chatops_sensor"]["token"]
             kwargs["endpoint"] = self.config["chatops_sensor"]["endpoint"]
             kwargs["channel"] = self.config["chatops_sensor"]["channel"]
-            del kwargs["reminder_type"]
+            kwargs["reminder_type"] = kwargs["chatops_reminder_type"]
+            del kwargs["chatops_reminder_type"]
 
         return kwargs
--- a/actions/src/openstack_actions.py
+++ b/actions/src/openstack_actions.py
@@ -62,10 +62,10 @@ class OpenstackActions(Action):
             )
             del kwargs["alertmanager_account_name"]
 
-        if "chatops_token" in kwargs:
+        if "reminder_type" in kwargs:
             kwargs["token"] = self.config["chatops_sensor"]["token"]
             kwargs["endpoint"] = self.config["chatops_sensor"]["endpoint"]
             kwargs["channel"] = self.config["chatops_sensor"]["channel"]
-            del kwargs["chatops_token"]
+            del kwargs["reminder_type"]
 
         return kwargs
--- a/lib/workflows/chatops_pr_reminder.py
+++ b/lib/workflows/chatops_pr_reminder.py
@@ -12,6 +12,6 @@ def pr_reminder(endpoint: str, token: str, channel: str, reminder_type: str) -> 
     """
     headers = {"Content-Type": "application/json", "Authorization": f"token {token}"}
     body = {"reminder_type": reminder_type, "channel": channel}
-    response = requests.post(endpoint, json=body, headers=headers)
+    response = requests.post(endpoint, json=body, headers=headers, timeout=120)
     response.raise_for_status()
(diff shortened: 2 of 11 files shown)


sensors/src/chatops_pr_reminder_sensor.py
```diff
@@ -0,0 +1,61 @@
...
+        self._logger.debug("ChatOpsPRReminderSensor making request...")
+        request = requests.post(
+            url=self.endpoint,
+            json={"reminder_type": request_type, "channel": self.channel},
+            headers={"Authorization": f"token {self.token}"}
+        )
+        if request.status_code != 200:
+            self._logger.error(f"ChatOpsPRReminderSensor request failed with {request.status_code} {request.text}")
+        else:
+            self._logger.debug("ChatOpsPRReminderSensor request successful.")
```
gmatthews20, line 43: From a stackstorm pov this would be an action triggered by a tigger created by this sensor and matching rule, you could look at using the built in `core.http` action which performs requests if you don't want to add a new action

khalford: I have looked at the core.http action but unfortunately rules cannot access the pack config. I will create an action

actions/src/openstack_actions.py
```diff
@@ -61,4 +61,11 @@ def parse_configs(self, **kwargs):
                 self.config, kwargs["alertmanager_account_name"]
             )
             del kwargs["alertmanager_account_name"]
+
+        if "chatops_token" in kwargs:
+            kwargs["token"] = self.config["chatops_sensor"]["token"]
+            kwargs["endpoint"] = self.config["chatops_sensor"]["endpoint"]
+            kwargs["channel"] = self.config["chatops_sensor"]["channel"]
+            del kwargs["chatops_token"]
```
gmatthews20, line 69: You could set this up like the other accounts we handle, as it is at the moment I don't think the values set in the action parameters will change which "token" is picked up from the config. If this is intended maybe drop them from the action parameters

rules/chatops.pr_reminder.global.yaml
```diff
@@ -0,0 +1,19 @@
+---
+name: chatops_pr_reminder_global
+pack: stackstorm_openstack
+description: Send a global reminder request every Monday and Wednesday at 9AM to the ChatOps application
```
gmatthews20, line 4: Is this 9AM? Cron is 8AM UTC?

khalford: For now it is. As we are in BST which is an hour ahead of UTC.

rules/chatops.pr_reminder.personal.yaml
```diff
@@ -0,0 +1,19 @@
+---
+name: chatops_pr_reminder_personal
+pack: stackstorm_openstack
+description: Send a personal reminder request every Monday at 9AM to the ChatOps application
```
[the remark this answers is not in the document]
khalford: For now it is. As we are in BST which is an hour ahead of UTC.

actions/src/openstack_actions.py
khalford: It is intended. I will remove them

rules/chatops.pr_reminder.global.yaml
gmatthews20: Yes what's the plan when BST ends?

tests/actions/test_openstack_actions.py
```diff
@@ -136,3 +136,25 @@ def test_parse_configs_with_no_accounts(self):
...
+
+    def test_parse_configs_with_chatops(self):
+        """
+        tests that parse_configs parses chatops config properly if reminder_type is provided
+        """
+        # Set the config properly
+        self.action.config = {
+            "chatops_sensor": {
+                "token": "mock_token",
+                "endpoint": "mock_endpoint",  # fix typo here
```
gmatthews20, line 148: fix typo?

gmatthews20 (approved):

tests/actions/test_openstack_actions.py
khalford: I have no idea where that comment came from. It wasn't me and I don't see a typo

lib/workflows/chatops_pr_reminder.py
```diff
@@ -0,0 +1,17 @@
+from typing import Dict
+import requests
+
+
+def pr_reminder(endpoint: str, token: str, channel: str, reminder_type: str) -> Dict:
```
AlexCK-STFC, line 5: You `del(kwargs[reminder_type])` in `openstack_actions.py`, but it's an argument to your action file.

Has this been tested outside of your unittest (which explicitly passes reminder_type)?

I assume you've tested on dev openstack, but I was under the impression (from cloud_account and uses_openstack in openstack_actions) that deleting the kwarg ther ewould make it no longer accessible here

rules/chatops.pr_reminder.global.yaml
AlexCK-STFC: Can you just change it from UTC to i.e. Europe/London which will fix this?

rules/chatops.pr_reminder.personal.yaml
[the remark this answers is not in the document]
AlexCK-STFC: See above

AlexCK-STFC (changes requested):

AlexCK-STFC (dismissed):

gmatthews20 (approved):

PR #354, opened 2025-06-09
ENH: Consolidate project creation workflows, add JASMIN project creation

jcornall:
### Description:

This pull request consolidates existing project creation workflows (`create_external_project.py`, `create_internal_project.py`, `create_project.py`) into one workflow (`create_project.py`). 

The modified `create_project.py` workflow can accept JASMIN-specific inputs related to project domain, user domain and network, and will validate these inputs if selected. 

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:


lib/workflows/create_jasmin_project.py
```diff
@@ -0,0 +1,179 @@
...
+    project = create_project(
+        conn,
+        ProjectDetails(
+            name=project_name,
+            email=project_email,
+            description=project_description,
+            immutable=project_immutable,
+            parent_id=parent_id,
+            is_enabled=True,
+        ),
```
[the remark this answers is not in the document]
gmatthews20: Following a conversation with Ops team project domain has been dropped for the following reason:

All projects are created in default domain. Projects cannot be created in jasmin/stfc domain since these are maps to a LDAP service so creating a project attempts to create a record in the LDAP service. Project Parent Id is use for accounting purposes

anish-mudaraddi (dismissed):

lib/structs/project_details.py
```diff
@@ -8,7 +8,6 @@ class ProjectDetails:
     name: str
     email: str
     description: str = ""
-    project_domain: str = None
```
DavidFair, line 11: I thought `project_domain` is used to determine which LDAP server they are authenticating against?
Without this it will use the STFC LDAP by default (unless I misunderstand how keystone was wired up)

DavidFair: Answered in the commit msg for https://github.com/stfc/st2-cloud-pack/pull/354/commits/99202c635c004bd5be52e8f8a7c66c6c4f4ab2e5

lib/workflows/create_project.py
```diff
@@ -47,38 +80,201 @@ def create_project(
...
-            subnet_name=f"{project_name}-subnet",
-            router_name=f"{project_name}-router",
-            number_of_floating_ips=number_of_floating_ips,
-            number_of_security_group_rules=number_of_security_group_rules,
-            project_immutable=project_immutable,
-            parent_id=parent_id,
-            admin_user_list=admin_user_list,
-            stfc_user_list=user_list,
+    elif network_type == Networks.INTERNAL:
+        setup_internal_networking(conn, project)
```
DavidFair, line 114: Should defensively guard against any new unknown types since we don't have `match` in this version of Python:
```python
else:
  raise ValueError(f"Unknown network type requested: {network_type}")
```

lib/workflows/create_project.py
```diff
@@ -47,38 +80,201 @@ def create_project(
...
+
+    # create default security group rules
+    if external_network == Networks.EXTERNAL:
+        create_external_security_group_rules(
+            conn, project_identifier=project.id, security_group_identifier="default"
+        )
+    elif external_network == Networks.JASMIN:
+        create_internal_security_group_rules(
+            conn, project_identifier=project.id, security_group_identifier="default"
         )
```
DavidFair, line 248: I'm a bit nervous about this, it feels like a potential foot-gun where if we add a new internal sec group rule it implicitly will affect JASMIN because of an implementation detail

My gut says it's worth duplicating the method to make a JASMIN specific one that we/they can explicitly audit and verify even if it duplicates code

DavidFair (changes requested):

lib/workflows/create_project.py
gmatthews20: Is this not covered by the check in the enum in from string?

PR #355, opened 2025-06-19
Add action for checking for routers with gateways on the internal net…

gmatthews20:
### Description:

Add action which checks for routers with an external gateway address starting with 172.16

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/enums/user_domains.py
+++ b/lib/enums/user_domains.py
@@ -12,6 +12,7 @@ class UserDomains(_AutoName):
     DEFAULT = auto()
     STFC = auto()
     OPENID = auto()  # irisiam - now openid since Stein
+    JASMIN = auto()
 
     @staticmethod
     def from_string(val: str):
--- a/lib/openstack_api/openstack_project.py
+++ b/lib/openstack_api/openstack_project.py
@@ -18,7 +18,7 @@ def create_project(
     :param project_details: A datastructure containing all the project details
     :return: A clouds project, or None if it was unsuccessful
     """
-    # domain_id can be blank, we will create a project in the default domain
+
     if not project_details.name:
         raise MissingMandatoryParamError("The project name is missing")
     if not project_details.email:
@@ -26,12 +26,19 @@ def create_project(
     email_regex = r"^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$"
     if not re.match(email_regex, project_details.email):
         raise ValueError("The project contact email is invalid")
+    if not project_details.project_domain:
+        raise MissingMandatoryParamError("The domain for the project is missing")
 
     create_project_kwargs = {
         "name": project_details.name,
         "description": project_details.description,
         "is_enabled": project_details.is_enabled,
     }
+
+    # get domain id for the domain to use
+    domain_details = conn.identity.find_domain(project_details.project_domain)
+    create_project_kwargs["domain_id"] = domain_details.id
+
     if project_details.parent_id:
         create_project_kwargs["parent_id"] = project_details.parent_id
 
--- a/lib/structs/project_details.py
+++ b/lib/structs/project_details.py
@@ -2,11 +2,13 @@ from dataclasses import dataclass
 from typing import Optional
 
 
+# pylint: disable=too-many-instance-attributes
 @dataclass
 class ProjectDetails:
     name: str
     email: str
     description: str = ""
+    project_domain: str = None
     is_enabled: Optional[bool] = None
     openstack_id: Optional[str] = None
     immutable: Optional[bool] = None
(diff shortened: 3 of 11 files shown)


config.schema.yaml
```diff
@@ -155,6 +155,18 @@ alertmanager_accounts:
...
+  description: "Openstack Router sensor specific settings."
+  type: "object"
+  required: false
+  additionalProperties: false
+  properties:
+    cloud_account:
+      type: "string"
+      description: "cloud account to use - set in clouds.ymal"
+      default: "dev"
+      required: true
```
DavidFair, line 168: Can we make this a global config option? Since I can't think of a case where we'll have a ST2 instance serving prod and dev

This avoids us having to set it per-sensor going forward too

lib/openstack_api/openstack_router.py
```diff
@@ -78,3 +81,35 @@ def create_router(conn: Connection, details: RouterDetails) -> Router:
...
+        return ips
+
+    routers_with_internal_gateway = []
+    for router in conn.network.routers():
+        logger.info("checking router %s", router.id)
+        ips = get_gateway_ips_from_router(router)
+        for ip in ips:
+            if ip.startswith("172.16"):
+                routers_with_internal_gateway.append(router)
+                logger.error("Found router with 172.16.xxx.xxx Address: %s", router.id)
```
DavidFair, line 113: Nit, this says `Address:` then packs the router UUID

Would be better to log the `Address: %s Router UUID: %s` so we've got two mechanisms to track this by

sensors/openstack_router_issue.yaml
```diff
@@ -0,0 +1,33 @@
+---
+class_name: OpenstackRouterSensor
+entry_point: src/openstack_router_sensor.py
+description: Monitor for issues with openstack routers
+poll_interval: 1800 # 30 mins
```
DavidFair, line 5: Nit: could we do every 10 minutes since this could be quite disruptive when a user does this and 29 minutes to have something call out probably means we've manually tracked it down by that point anyway

AFAIK it's a relatively lightweight check, but if it's heavy we can increase it later on?

DavidFair (changes requested):

tests/lib/openstack_api/test_openstack_router.py
```diff
@@ -178,3 +182,47 @@ def test_create_router_successful():
         is_ha=True,
     )
     assert res == mock_conn.network.create_router.return_value
+
+
+def test_check_internal_router_found():
```
meoflynn, line 187: Could you add a docstring for the unit test?

tests/lib/openstack_api/test_openstack_router.py
```diff
@@ -178,3 +182,47 @@ def test_create_router_successful():
...
+    }
+    mock_conn = MagicMock()
+    mock_conn.network.routers.return_value = [mock_router_1, mock_router_2]
+
+    res = check_for_internal_routers(mock_conn)
+
+    assert res == [mock_router_1]
+
+
+def test_check_internal_router_not_found():
```
meoflynn, line 211: Could you add a docstring for the unit test?

tests/sensors/test_openstack_router_sensor.py
```diff
@@ -0,0 +1,45 @@
+from unittest.mock import MagicMock, patch
+import pytest
+from sensors.src.openstack_router_sensor import OpenstackRouterSensor
+
+
+@pytest.fixture(name="sensor")
+def state_sensor_fixture():
```
meoflynn, line 7: Could you add a docstring for this method?

meoflynn (changes requested):

sensors/src/openstack_router_sensor.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def __init__(self, sensor_service, config=None, poll_interval=None):
+        super().__init__(
+            sensor_service=sensor_service, config=config, poll_interval=poll_interval
+        )
+        self._log = self._sensor_service.get_logger(__name__)
+        self.cloud_account = self.config["sensor_cloud_account"]
+
+    def setup(self):
+        pass
```
meoflynn, line 27: Codecov is reporting this line doesn't have a unit test

tests/sensors/test_hypervisor_state_sensor.py
```diff
@@ -8,7 +8,10 @@
 def state_sensor_fixture():
```
meoflynn, line 8: Could you add a docstring to this method?

meoflynn (changes requested):

meoflynn (dismissed):

anish-mudaraddi (approved):

anish-mudaraddi (approved):

DavidFair (approved):

PR #360, opened 2025-08-21
ENH: Add Image Metadata Sensor

jcornall:
### Description:

Add a Stackstorm sensor that periodically gathers and sends the properties/metadata for all the dev cloud images. 

The sensor queries the dev cloud images and from this query compiles a payload of relevant properties/metadata, which is then dispatched. The `image.metadata_change.yaml` configuration file defines some important variables, such as the `poll_interval`, which controls the polling frequency (currently set to 24 hours to poll daily), as well as the `trigger_types`. 

The sensor injects the trigger into StackStorm when a condition is met, in this case when the `poll_interval` has elapsed.

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:


dev-0pz (dismissed):
LGTM

DavidFair (commented):
Currently this is hardcoded to go `prod` -> `dev`
Would it be possible to add some fields like `from_cloud` `to_cloud` to the config so that it's obvious to see we're initially going prod -> dev, but later on we can go dev -> prod (which we should do instead)

gmatthews20:
> Currently this is hardcoded to go `prod` -> `dev` Would it be possible to add some fields like `from_cloud` `to_cloud` to the config so that it's obvious to see we're initially going prod -> dev, but later on we can go dev -> prod (which we should do instead)

@DavidFair This currently just gathers the metadata from a single cloud instance, work to add the comparison is WIP and will follow but this is a good idea

DavidFair (approved):

AlexCK-STFC (approved):

PR #385, opened 2025-09-29
Simplify send emails code

jose-caballero:
### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/tests/lib/apis/email_api/test_emailer.py
+++ b/tests/lib/apis/email_api/test_emailer.py
@@ -1,9 +1,8 @@
 from email.mime.multipart import MIMEMultipart
 from pathlib import Path
-from unittest.mock import patch, call, MagicMock, mock_open, NonCallableMock
+from unittest.mock import MagicMock, NonCallableMock, call, mock_open, patch
 
 import pytest
-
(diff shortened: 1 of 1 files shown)


lib/apis/email_api/emailer.py
```diff
@@ -54,39 +51,45 @@ def send_emails(self, emails: List[EmailParams]):
...
+            )
 
-                server.sendmail(
-                    email_params.email_from,
-                    send_to,
-                    self.build_email(email_params).as_string(),
-                )
+            send_to = list(email_params.email_to)
+            if email_params.email_cc:
+                send_to.extend(email_params.email_cc)
```
khalford, line 72: Does this send the email to the people in the CC list and not add them as CC?

khalford (approved):

lib/apis/email_api/emailer.py
AlexCK-STFC: No, the SMTP server itself doesn't care about CC, it just wants a list of addressed to email to.
The CC is handled in the email headers in build_email

https://stackoverflow.com/a/1546435

PR #389, opened 2025-10-01
ENH: Add emailer for fixed line testing

AlexCK-STFC:
### Description:

Note: Dates are hard-coded, while flavors are customisable.

This is because:
  - The dates are unlikely to change
  - It would be a pain to make them customisable from the format of the template

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:


AlexCK-STFC:
Draft until I add unit tests

anish-mudaraddi (changes requested):
Tempted to close this and reopen if we need it later

anish-mudaraddi (approved):

PR #379, opened 2025-09-18
ENH: Add Flavor Sensor Rule to create Jira task

jcornall:
### Description:

Adds a Rule that creates a Jira task whenever the FlavorPropertiesSensor is triggered. 

Tested on Jira Project DevOps_Cloud_Test_Env queue.

The sync direction is dev >>> prod, in line with the image metadata sensor that's been merged into main - this can be changed in the flavor sensor config, within `config.schema.yaml`.

Much of this PR is dependent on #371, recommend reviewing and merging that PR before merging this one. 

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?
* [ ] Checked the latest commit runs on Dev?
* [ ] Updated the example config file(s) and README?

---

### Reviewer

Does this PR:

* [ ] Place non-StackStorm code into the `lib` directory?
* [ ] Have unit tests for the action/sensor and `lib` layers?
* [ ] Have clear and obvious action parameter names and descriptions?

PR #393, opened 2025-10-03
ENH: Add replacement util for deepdiff

AlexCK-STFC:
### Description:

Drops dependency on deepdiff by replacing functionality

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/rules/alert.flavor.properties.mismatch.yaml
+++ b/rules/alert.flavor.properties.mismatch.yaml
@@ -19,11 +19,5 @@ action:
       - Source Flavor ID: {{trigger.source_flavor_id}}\\
       - Target Flavor ID: {{trigger.target_flavor_id}}\\
       \\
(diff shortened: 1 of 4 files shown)


AlexCK-STFC:
Might cause issues for #379. Renamed "mismatch" -> "diff" for example to match the other two sensors.

We can also unify them in future

anish-mudaraddi (approved):
Your code all made sense to me.
Thanks for considering all the use-cases for the diff utility function - I initially hadn't considered them when I asked you to implement this.  

In future, it might be best to just implement what's needed and extend the functionality as we need it. In this case - just ordered nested lists/dicts usecases

I suspect it will be a while before we need all the functionality you implemented will be needed

PR #394, opened 2025-10-06
Move function find_servers_with_decom_flavor

jose-caballero:
Move the function to find servers with a given flavor name (or names) to the API directory for the query library:

      lib/apis/openstack_query_api

The code to send emails shouldn't need to know about the internal details of the query library.
Moreover when there is an API directory for that query library. All code that needs to use it should be implemented as part of the API.

In this commit we implement the following changes:

* move the function to find servers to the query library API
* remove the function to find servers from the workflow send_decom_flavor_email.py
* implement a new unittest file for the new code in the API directory
* fix some incorrect function names in the old unittest code (functions had string "user" where it should be "servers")
* fix the unittest code for the workflow

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/tests/lib/apis/openstack_query_api/test_server_queries.py
+++ b/tests/lib/apis/openstack_query_api/test_server_queries.py
@@ -41,7 +41,6 @@ def test_find_servers_with_flavor_valid(mock_flavor_query):
         "server_name",
         "addresses",
     )
-    mock_server_query_obj.to_props.assert_called()
 
     mock_server_query_obj.append_from.assert_called_once_with(
         "PROJECT_QUERY", "test-cloud-account", ["project_name"]


lib/apis/openstack_query_api/server_queries.py
```diff
@@ -0,0 +1,54 @@
...
+    cloud_account: str,
+    flavor_name_list: List[str],
+    from_projects: Optional[List[str]] = None,
+):
+    """
+    Use QueryAPI to run the query to find decom flavors
+    :param cloud_account: string represents cloud account to use
+    :param flavor_name_list: A list of flavor names to be decommissioned
+    :param from_projects: A list of project identifiers to limit search in
+    """
```
gmatthews20, line 15: I think this docstring is to specific to decom flavors

lib/apis/openstack_query_api/server_queries.py
```diff
@@ -0,0 +1,54 @@
...
+        "server_id",
+        "server_name",
+        "addresses",
+    )
+
+    if not server_query.to_props():
+        raise RuntimeError(
+            f"No servers found with flavors {', '.join(flavor_name_list)} on projects "
+            f"{','.join(from_projects) if from_projects else '[all projects]'}"
+        )
```
gmatthews20, line 49: Should we raise an exception here, there may be cases where we expect no servers to be found, e.g. checking all decom flavor VMs have been removed?

gmatthews20 (changes requested):

PR #430, opened 2026-03-17
Fix quota.yaml permission issue

dev-0pz:
### Description:

```
POST 
 400 (Bad Request)
 
Unable to schedule action &quot;stackstorm_openstack.quota.show&quot;.
 
Error: Request failed with status code 400
    at createError (
)
    at settle (
)
    at XMLHttpRequest.onloadend (
)
```
Quota.yaml doesnt allow for me to be able to use quota show since enable is set to false
### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/quota.show.yaml
+++ b/actions/quota.show.yaml
@@ -1,6 +1,6 @@
 ---
 description: Show the current quota for project
-enabled: false
+enabled: true
 entry_point: src/openstack_actions.py
 name: quota.show
 parameters:


jose-caballero (approved):
Note that the actual function may not do anything, as it doesn't return anything. It may just unconditionally print out the values. But can only check it by trying it, which requires the action to be enable :)

RJaliliSTFC (approved):

PR #436, opened 2026-04-28
Add group creation

jose-caballero:
### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/apis/openstack_api/openstack_groups.py
+++ b/lib/apis/openstack_api/openstack_groups.py
@@ -1,0 +1,22 @@
+import re
+from openstack.exceptions import ConflictException
+
+
+def create_group(conn, group_name):
+    """
+    Creates a new users group
+    :param conn: openstack connection object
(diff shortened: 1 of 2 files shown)


lib/apis/openstack_api/openstack_groups.py
```diff
@@ -0,0 +1,18 @@
...
+
+def create_group(conn, group_name):
+    """
+    Creates a new users group
+    :param conn: openstack connection object
+    :param group_name: the name of the new users group
+    :return: a group object
+    """
+    if "/" in group_name:
+        raise ValueError(f"slash character / detected in group name {group_name}")
```
khalford: I would suggest changing this to check for any characters not in the allowed group [a-zA-Z0-9] and also "-"

jose-caballero: done

PR #441, opened 2026-05-19
Add method to shutdown server

jose-caballero:
### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/apis/openstack_api/openstack_server.py
+++ b/lib/apis/openstack_api/openstack_server.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 import logging
 import time
-from typing import Optional
+from typing import Optional, List
 from openstack.connection import Connection
 from openstack.compute.v2.image import Image
 from openstack.compute.v2.server import Server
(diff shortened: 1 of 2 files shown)


lib/apis/openstack_api/openstack_server.py
```diff
@@ -203,3 +203,51 @@ def delete_server(
...
+    :rtype: None
+    """
+    server = conn.compute.find_server(server_id)
+    logger.info("Attempt to shutoff server %s", server.id)
+    if server.status == "ACTIVE":
+        logger.info("Shutting off server: %s", server.id)
+        conn.compute.stop_server(server)
+        logger.info("Waiting for server to shut off: %s", server.id)
+        conn.compute.wait_for_status(server, status="SHUTOFF")
+        logger.info("server is  shut off: %s", server.id)
```
DavidFair: Could we add a check here that the status doesn't go to error?
If it does we should `raise RuntimeError` as this shouldn't happen and probably needs manual intervention

jose-caballero: Sounds reasonable. Let me check.

DavidFair:
Also there's no way to test this code using ST2 - could you add action yamls to wrap both calls so I can test on a dev ST2 instance?

DavidFair (dismissed):
Making this a formal review

Needs unit testing (even in the PR template)
Needs to be runnable on a dev instance else we're working on hopes and prayers (again in the PR template)

Before we accept code into ST2 where it can run on a prod instance with elevated permissions automatically in the future....

lib/apis/openstack_api/openstack_server.py
jose-caballero: done

jose-caballero:
I am merging this branch into another one.

PR #458, opened 2026-06-24
Fix check for draining

jose-caballero:
### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/server.migrate.yaml
+++ b/actions/server.migrate.yaml
@@ -34,4 +34,9 @@ parameters:
     type: boolean
     required: true
     default: true
+  check_flavor:
+    description: "Check the Flavor of the Server to decide if migration is feasible. When set as True, GPUs, FPGAs, and Flavors with more than 60 CPUs will NOT be migrated."
+    type: boolean
+    required: true
+    default: true
 runner_type: python-script
--- a/lib/apis/openstack_api/openstack_server.py
+++ b/lib/apis/openstack_api/openstack_server.py
@@ -10,18 +10,19 @@ from openstack.exceptions import ResourceFailure, ResourceTimeout
 logger = logging.getLogger(__name__)
 
 
-def can_be_migrated(server: Server):
-    if server.flavor.name.startswith("g-") or server.flavor.name.startswith("f-"):
-        raise ValueError(
-            f"Attempted to move GPU or FPGA flavor, {server.flavor.name}, which is not allowed!"
-        )
+def can_be_migrated(server: Server, check_flavor: bool):
+    if check_flavor:
+        if server.flavor.name.startswith("g-") or server.flavor.name.startswith("f-"):
+            raise ValueError(
+                f"Attempted to move GPU or FPGA flavor, {server.flavor.name}, which is not allowed!"
+            )
+        if server.flavor.vcpus > 60:
+            raise ValueError(
+                f"Attempted to move flavor with greater than 60 cores, {server.flavor.name}, which is not allowed!"
+            )
     if server.status not in ["ACTIVE", "SHUTOFF"]:
         raise ValueError(
             f"Server status: {server.status}. The server must be ACTIVE or SHUTOFF to be migrated"
-        )
-    if server.flavor.vcpus > 60:
-        raise ValueError(
-            f"Attempted to move flavor with greater than 60 cores, {server.flavor.name}, which is not allowed!"
         )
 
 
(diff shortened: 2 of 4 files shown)


actions/server.migrate.yaml
```diff
@@ -34,4 +34,9 @@ parameters:
     type: boolean
     required: true
     default: true
+  check_flavor:
+    description: "Check the Flavor of the Server to decide if migration is feasible. When set as True, GPUs, FPGAs, and Flavors with more than 60 CPUs will NOT be migrated."
```
DavidFair, line 38: Nit, I would call it `live_migration`
Then the tick box (default) would do a live migration on CPU instances <60 cores

Unticking it would do a cold migration which can be done on anything but will shutdown while it happens

jose-caballero: That may be a good idea for the long term, as a separate flag. 

Note that the purpose here is slightly different. We are NOT doing live migrations during the outage week in July. 

This change is just to allow the migration of VMs with flavors for GPUs/FPGAs and/or more that 60 CPUs, which are currently blocked by st2. This PR allows a way to keep that check for those types of flavors in place, to avoid trying risky migrations by mistake, but at the same time having a mechanism to force the migration this time.

lib/apis/openstack_api/openstack_server.py
```diff
@@ -10,25 +10,27 @@
 logger = logging.getLogger(__name__)
 
 
-def can_be_migrated(server: Server):
-    if server.flavor.name.startswith("g-") or server.flavor.name.startswith("f-"):
-        raise ValueError(
-            f"Attempted to move GPU or FPGA flavor, {server.flavor.name}, which is not allowed!"
-        )
+def can_be_migrated(server: Server, check_flavor: bool):
```
anish-mudaraddi, line 13: put a default value here as `check_flavor=false` to keep backwards compatible with other workflows using this function

lib/apis/openstack_api/openstack_server.py
```diff
@@ -10,25 +10,27 @@
...
-        raise ValueError(
-            f"Attempted to move flavor with greater than 60 cores, {server.flavor.name}, which is not allowed!"
-        )
 
 
 def snapshot_and_migrate_server(
     conn: Connection,
     server_id: str,
     snapshot: bool,
+    check_flavor: bool,
```
anish-mudaraddi, line 33: do the same here `check_flavor=false`

tests/lib/apis/openstack_api/test_openstack_server.py
```diff
@@ -150,6 +153,7 @@ def test_no_snapshot_migration(mock_wait_for_migration_status, mock_snapshot_ser
         conn=mock_connection,
         server_id=mock_server_id,
         snapshot=False,
+        check_flavor=True,
```
anish-mudaraddi, line 156: don't need this anymore if you put a default value

tests/lib/apis/openstack_api/test_openstack_server.py
```diff
@@ -79,6 +80,7 @@ def test_shutoff_migration(
         server_id=mock_server_id,
         dest_host=dest_host,
         snapshot=True,
+        check_flavor=True,
```
anish-mudaraddi, line 83: dont need this if you put a default value

actions/server.migrate.yaml
DavidFair: Yeah we need to do a "cold migration" for these types of flavors
But we'll need to do this post-outage too (e.g. HW failure) and the steps will be the same

The fact it skips the checks is an implementation detail - we just want a way to force drain these types of flavors which will incur an offline

tests/lib/apis/openstack_api/test_openstack_server.py
```diff
@@ -43,6 +43,7 @@ def test_active_migration(
...
```
anish-mudaraddi, line 46: change this to `check_flavor` because you're overwriting what it set as an argument

lib/apis/openstack_api/openstack_server.py
DavidFair: I think it was the other way around, before this PR we always did the check as it was always a live migration
Now the check makes it optional for cold migrations - existing workflows will assume a live migration instead

PR #459, opened 2026-06-29
Add flag live migration

jose-caballero:
### Special Notes:

---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/apis/openstack_api/openstack_server.py
+++ b/lib/apis/openstack_api/openstack_server.py
@@ -15,20 +15,39 @@ def can_be_migrated(server: Server):
         raise ValueError(
             f"Attempted to move GPU or FPGA flavor, {server.flavor.name}, which is not allowed!"
         )
-    if server.status not in ["ACTIVE", "SHUTOFF"]:
-        raise ValueError(
-            f"Server status: {server.status}. The server must be ACTIVE or SHUTOFF to be migrated"
-        )
     if server.flavor.vcpus > 60:
         raise ValueError(
             f"Attempted to move flavor with greater than 60 cores, {server.flavor.name}, which is not allowed!"
         )
 
 
+def _cold_migration(
+    conn: Connection,
+    server: Server,
+    dest_host: Optional[str] = None,
+) -> None:
+    conn.compute.migrate_server(server=server.id, host=dest_host)
+    conn.compute.wait_for_server(server, status="VERIFY_RESIZE", wait=3600)
+    conn.compute.confirm_server_resize(server.id)
+    wait_for_migration_status(conn, server.id, "confirmed")
+
+
+def _live_migration(
+    conn: Connection,
+    server: Server,
+    dest_host: Optional[str] = None,
+) -> None:
+    conn.compute.live_migrate_server(
+        server=server.id, host=dest_host, block_migration=True
+    )
+    wait_for_migration_status(conn, server.id, "completed")
+
+
 def snapshot_and_migrate_server(
     conn: Connection,
     server_id: str,
     snapshot: bool,
+    live_migration: bool,
     dest_host: Optional[str] = None,
 ) -> None:
     """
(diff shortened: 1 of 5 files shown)


DavidFair:
Notes from testing:

To fix:
- Attempting to live_migrate a GPU will snapshot first (could take hours) then will fail as it's blocked, need to check if the migration is permitted before doing any work
~This does a live snapshot by default which requires qemu-ga. I think we should force an offline snapshot for safety to handle cases where `qemu-ga` isn't working and blocks a live migration (see below)~
- If the source is disabled and the destination has no slots the ST2 action will sit pending despite it going into error on `server migration list`. I think there's a bug on detecting a migration (not server) going into error state and this caught me out several times and needs fixing before we do this for real

To note (gotchas):
- When no space is available VM migrates back to same HV (no shutdown triggered) if it's not marked disable
~VM with qemu-guest-agent uninstalled / offline gives `quiescing instance: Error from libvirt while quiescing [Error Code 86] Guest agent is not responding: QEMU guest agent is not connected.` but will snapshot still giving potentially partially written images. If we can somehow ask OpenStack to live migrate OR if we cannot quiesce shutdown then snapshot~ :thinking:
Edit: this isn't possible, we already have `os_require_quiesce` and if the GA isn't available it will still carry on to be "crash-safe`

Test cases that worked:
- GPU -> shutdown gracefully, then restarted without intervention from myself or the "user"
- Multiple migrations back and forth
- Handles low space on source HV (1.5GB free, completed fine still)
- Snapshots all created in ST2 service account's project
- Concurrent migrations migrations work correctly
- VM that has an error for a previous migration (e.g. no space on aggregate) re-tried and completed after making space
- Qemu-GA offline migrates successfully with the caveat about snapshots above

DavidFair:
@jose-caballero could you look at:
> Attempting to live_migrate a GPU will snapshot first (could take hours) then will fail as it's blocked, need to check if the migration is permitted before doing any work

This should check for GPU or error before we do any steps - it's likely a bug in the existing implementation too



I'm currently looking at the endless wait if it fails:
> If the source is disabled and the destination has no slots the ST2 action will sit pending despite it going into error on server migration list. I think there's a bug on detecting a migration (not server) going into error state and this caught me out several times and needs fixing before we do this for real

DavidFair:
Fixed the hang - need to be careful we both fetch before editing or rebasing from this point

> This should check for GPU or VM's in invalid states/errors before we do any steps - it's likely a bug in the existing implementation too

I haven't made a start so if you beat me to it tomorrow feel free to push another commit onto this, then I think we're good to retest and maybe even :shipit:

DavidFair:
> Attempting to live_migrate a GPU will snapshot first (could take hours) then will fail as it's blocked, need to check if the migration is permitted before doing any work

#461 after this is reviewed + merged

> If the source is disabled and the destination has no slots the ST2 action will sit pending despite it going into error on server migration list. I think there's a bug on detecting a migration (not server) going into error state and this caught me out several times and needs fixing before we do this for real

This now correctly works 

```
2026-07-01 09:33:02,919 - apis.openstack_api.openstack_server - INFO - Migrating server: 4dbe774a-b036-4b1c-8323-ad167df31cbe
2026-07-01 09:33:03,363 - apis.openstack_api.openstack_server - INFO - Status of migration of server 4dbe774a-b036-4b1c-8323-ad167df31cbe: error
Traceback (most recent call last):
  File "/opt/stackstorm/st2/lib/python3.10/site-packages/python_runner/python_action_wrapper.py", line 397, in <module>
    obj.run()
---snipped----
    raise ResourceFailure(migration)
openstack.exceptions.ResourceFailure: openstack.compute.v2.migration.Migration(id=10603, uuid=528df2b2-d2c8-4fc8-b93e-89d7871fcc60, source_compute=hv-ref2023-a4000-26.nubes.rl.ac.uk, dest_compute=None, source_node=hv-ref2023-a4000-26.nubes.rl.ac.uk, dest_node=None, dest_host=None, old_instance_type_id=539, new_instance_type_id=539, instance_uuid=4dbe774a-b036-4b1c-8323-ad167df31cbe, status=error, migration_type=migration,
"
  stdout: ''
```

There's a separate bug where if the most recent migration goes into error it will re-migrate but incorrectly pick up on that previous error now we've got stricter checking. We'll need to do another PR for that but it's not a regression from this work since it would error before too....

And the migrations with enough space still work, so I'm going to mark it as tested on dev

```
    2026-07-01 09:38:19,244 - apis.openstack_api.openstack_server - INFO - Starting snapshot of server: 4dbe774a-b036-4b1c-8323-ad167df31cbe
    2026-07-01 09:38:53,886 - apis.openstack_api.openstack_server - INFO - Completed snapshot of server: 4dbe774a-b036-4b1c-8323-ad167df31cbe
    2026-07-01 09:39:03,896 - apis.openstack_api.openstack_server - INFO - Migrating server: 4dbe774a-b036-4b1c-8323-ad167df31cbe
    2026-07-01 09:39:04,416 - apis.openstack_api.openstack_server - INFO - Status of migration of server 4dbe774a-b036-4b1c-8323-ad167df31cbe: confirmed
    2026-07-01 09:39:09,443 - apis.openstack_api.openstack_server - INFO - Status of migration of server 4dbe774a-b036-4b1c-8323-ad167df31cbe: migrating
    2026-07-01 09:39:14,470 - apis.openstack_api.openstack_server - INFO - Status of migration of server 4dbe774a-b036-4b1c-8323-ad167df31cbe: post-migrating
    2026-07-01 09:39:19,498 - apis.openstack_api.openstack_server - INFO - Status of migration of server 4dbe774a-b036-4b1c-8323-ad167df31cbe: finished
    2026-07-01 09:39:19,601 - apis.openstack_api.openstack_server - INFO - Confirming resize for 4dbe774a-b036-4b1c-8323-ad167df31cbe
    2026-07-01 09:39:19,641 - apis.openstack_api.openstack_server - INFO - Migration completed of server: 4dbe774a-b036-4b1c-8323-ad167df31cbe
    '
  stdout: ''
```

DavidFair (dismissed):
Approved for Jose's work

@jose-caballero should review and approve mine with a comment (can't approve own PR) then we just need a second reviewer to merge

DavidFair (approved):
Right take 2 - approved pending a second reviewer....

anish-mudaraddi (approved):

lib/apis/openstack_api/openstack_server.py
```diff
@@ -37,24 +70,30 @@ def snapshot_and_migrate_server(
...
+    server_status = server.status.casefold()
+
+    match server_status:
+        case "shutoff":
+            _cold_migration(conn, server, dest_host)
+        case "active":
+            if live_migration:
+                can_be_migrated(server)
+                _live_migration(conn, server, dest_host)
+            else:
```
anish-mudaraddi, line 91: I'm sure you've thought about this but don't we need to shutoff the server to do a cold migration?

Does conn.compute.migrate_server() turn off the server? Perhaps an explicit shutoff command would be worthwhile in case the API changes - we should be as explicit as possible

jose-caballero: yes, conn.compute.migrate_server() turns of the VM before migrating it.

PR #461, opened 2026-07-01
BUG: Skip any migration or snapshot steps if we're in an invalid state

DavidFair:
### Description:
Skips any migration steps, as we would previously take a snapshot - force the user to wait for multiple hours then fail when we knew up-front it wouldn't work.

Instead provide rapid feedback that we can't do an error state or we can't migrate before doing any work

TODO: test this works on dev once #459 is merged

### Special Notes:
---

### Submitter:

Have you (where applicable):

* [x] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/lib/apis/openstack_api/openstack_server.py
+++ b/lib/apis/openstack_api/openstack_server.py
@@ -10,7 +10,7 @@ from openstack.exceptions import ResourceFailure, ResourceTimeout
 logger = logging.getLogger(__name__)
 
 
-def can_be_migrated(server: Server):
+def check_can_be_live_migrated(server: Server):
     if server.flavor.name.startswith("g-") or server.flavor.name.startswith("f-"):
         raise ValueError(
             f"Attempted to move GPU or FPGA flavor, {server.flavor.name}, which is not allowed!"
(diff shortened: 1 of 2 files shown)


jose-caballero (approved):

DavidFair:
Now correctly skips all steps if we attempt to do it with an error:
```
st2.actions.python.OpenstackActions: DEBUG    with Parameters: server_id: 4dbe774a-b036-4b1c-8323-ad167df31cbe
live_migration: False
cloud_account: dev
dest_host: None
snapshot: True
Traceback (most recent call last):
  File "/opt/stackstorm/st2/lib/python3.10/site-packages/python_runner/python_action_wrapper.py", line 397, in <module>
    obj.run()
  File "/opt/stackstorm/st2/lib/python3.10/site-packages/python_runner/python_action_wrapper.py", line 216, in run
    output = action.run(**self._parameters)
  File "/opt/stackstorm/packs/stackstorm_openstack/actions/src/openstack_actions.py", line 34, in run
    return action_func(**kwargs)
  File "/opt/stackstorm/packs/stackstorm_openstack/lib/apis/openstack_api/openstack_server.py", line 88, in snapshot_and_migrate_server
    raise ValueError(
ValueError: Server status: ERROR. The server must be ACTIVE or SHUTOFF to be migrated
"
  stdout: ''
```

PR #467, opened 2026-07-08
simplify emailing workflow

anish-mudaraddi:
### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:


gmatthews20:
A thought I had about emails, why don't we just use the Jira API to open a support ticket rather than sending an email and copying our support email to open a ticket?

anish-mudaraddi:
I like this idea a lot - since it saves us from having to maintain an emailing API

PR #470, opened 2026-07-10
Bump HV drain concurrency

gmatthews20:
### Description:

HV Drains limit now 10
Server migrates limit now 6 globally

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/policies/hypervisor.drain.concurrency.yaml
+++ b/policies/hypervisor.drain.concurrency.yaml
@@ -8,4 +8,6 @@ parameters:
   action: delay
   # Three HVs can be drained concurrently, this does not affect the number of
   # VMs draining per HV
-  threshold: 3
+  # Limit needed as drain disables the HV so needed to prevent the entire cloud being disabled
+  # all at once
+  threshold: 10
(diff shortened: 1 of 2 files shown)


policies/hypervisor.drain.concurrency.yaml
```diff
@@ -8,4 +8,4 @@ parameters:
   action: delay
   # Three HVs can be drained concurrently, this does not affect the number of
   # VMs draining per HV
-  threshold: 3
+  threshold: 10
```
DavidFair, line 13: Can we drop this limit completely? Since it's the migrate that's the slow step and having even 100 actions hit the API isn't going to cause any noticeable problems for the backend

gmatthews20: I think it was originally there to prevent the automation disabling the entire cloud all at once

DavidFair: Ah the comment doesn't make that clear - happy to leave it at 10/10 then :+1:

DavidFair (dismissed):

DavidFair (approved):

jose-caballero (approved):

PR #484, opened 2026-07-22
Add function to handle Servers metadata

jose-caballero:
We are going to use the metadata field to mark certain Servers that require actions.
For example, to mark the Servers that are running an EOL version of the Operating System and when the user was notified about it.

These 2 functions just perform the set and delete steps. The exact content and format of the parameters is responsibility of the workflows code calling them.

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:


anish-mudaraddi:
these functions are just calling openstacksdk's built-in method `.set_server_metadata` and `.delete_server_metadata` and not doing anything else

can't we just call these built-in methods directly in any action/API function that needs to set metadata instead of creating a whole new function that once again verifies that the server exists (slowing down the process by adding an unnecessary API call `find_server`) 

I could see this being useful if you're putting some global validation checks on the properties that is unique to us - but maybe you should just wait until there's a use-case like that before creating these functions? I'm wary of adding more code we need to maintain

jose-caballero:
@anish-mudaraddi you mentioned that these 2 functions don't add anything. 
I can see your point. I agree on waiting until we know what we really needed. 
However, at least for set_metadata( ), will it be justified if we always add a timestamp to the value? 
Something like this:

    set_metadata(conn, server_id, PARAM=VALUE)

adds a string like this 

    "PARAM=VALUE @ 2026-04-14:07:15:40

PR #487, opened 2026-07-27
Get the email address  of the User who owns a Server

jose-caballero:
Write a function to get the email address of the User who instantiated a given Server.

This information will be used to be able to send emails to those Users when something problematic has been detected on the Servers. For example, if the Server is running an EOL version of the Operating System or have been in SHUTOFF state for too long.

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:


lib/apis/openstack_api/openstack_server.py
```diff
@@ -296,3 +301,42 @@ def shutoff_server_list(conn: Connection, server_id_list: List[str]) -> None:
...
+        )
+
+    user = conn.identity.get_user(user_id)
+    if user is None:
+        raise ResourceNotFound(
+            f"User '{user_id}' referenced by server '{server_id}' was not found."
+        )
+
+    user_email = getattr(user, "email", None)
+    if not user_email:
```
khalford, line 336:
```suggestion
    if not user_id:
        raise ResourceNotFound(
            f"Server '{server_id}' does not have an associated user_id."
        )

    user = conn.identity.get_user(user_id)
    if not user:
        raise ResourceNotFound(
            f"User '{user_id}' referenced by server '{server_id}' was not found."
        )

    user_email = getattr(user, "email", None)
    if not user_email:
```

Is there any reason to use "if user is None" rather than "if not user" in this case?

khalford (changes requested):
Unit tests?

gmatthews20 (changes requested):
Is this not a duplicate of this function?
https://github.com/stfc/st2-cloud-pack/blob/main/lib/apis/openstack_query_api/user_queries.py

PR #488, opened 2026-07-27
Write code to query Wazuh

jose-caballero:
In order to be able to query Wazuh we need 2 things:

* some code to perform the query and fetch the results. We implement that code as a class
  * the init method reads the all needed credentials and related parameters from stackstorm configuration
  * we get a token always right before querying the server

* the entry point for all ACTIONS in stackstorm needs to create an instance of the new class and pass it as argument to the workflows

### Special Notes:

---

### Submitter:

Have you (where applicable):

* [ ] Added unit tests?

---

### Reviewer

Does this PR:

### DIFF
--- a/actions/src/openstack_actions.py
+++ b/actions/src/openstack_actions.py
@@ -6,14 +6,22 @@ from apis.alertmanager_api.structs.alertmanager_account import AlertManagerAccou
 from apis.email_api.structs.smtp_account import SMTPAccount
 from apis.icinga_api.structs.icinga_account import IcingaAccount
 from apis.jira_api.structs.jira_account import JiraAccount
+from apis.wazuh_api.wazuh import WazuhClient
 
 
 class OpenstackActions(Action):
-    def run(self, lib_entry_point: str, requires_openstack: bool = False, **kwargs):
+    def run(
+        self,
+        lib_entry_point: str,
+        requires_openstack: bool = False,
+        requires_wazuh: bool = False,
+        **kwargs,
+    ):
         """
         Dynamically dispatches to the function wanted
         :param lib_entry_point: path to function that handles action in lib layer
         :param requires_openstack: if action requires connection to openstack
+        :param requires_wazuh: if action requires querying the Wazuh server
         :param kwargs: all user-defined kwargs to pass to the function
         """
         module, fn_name = lib_entry_point.rsplit(".", 1)
@@ -25,6 +33,19 @@ class OpenstackActions(Action):
             "\n".join([f"{key}: {val}" for key, val in kwargs.items()]),
         )
         kwargs = self.parse_configs(**kwargs)
+
+        # in some automations we may need to query the Wazuh server
+        # however, we may not need any field in the web form related to Wazuh
+        # therefore, it will not be possible to handle the creation of
+        # the wazuh client object inside function parse_config()
+        # as there is no configuration parameter related to it that can
+        # be parsed
+        # we trigger the creation of the wazuh client object with an immutable
+        # parameter, the same way we do for OpenStack
+        if requires_wazuh:
+            wazuh = WazuhClient(self.config)
+            kwargs["wazuh"] = wazuh
+
         if not requires_openstack:
             return action_func(**kwargs)
         # setup openstack connection
--- a/lib/apis/wazuh_api/wazuh.py
+++ b/lib/apis/wazuh_api/wazuh.py
@@ -1,0 +1,145 @@
+import logging
+import re
+import urllib3
+import requests
+
+# Suppress insecure request warnings if using self-signed certificates
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+
+logger = logging.getLogger("WazuhClient")
+
+
+# pylint: disable=too-few-public-methods
+class WazuhClient:
+    def __init__(self, pack_config):
+        """
+        gets the requires values to contact a Wazuh Server
+        from the StackStorm configuration Dictionary pack_config
+
+        :param pack_config: StackStorm configuration
+        :type pack_config: dict
+        """
+        logger.info("Creating instance of WazuhClient")
+        self.username = pack_config.get("wazuh_username")
+        self.password = pack_config.get("wazuh_password")
+        self.url = pack_config.get("wazuh_url")
+        logger.info("Instance of WazuhClient created correctly")
+
+    def _get_wazuh_token(self):
+        """
+        Authenticates with the Wazuh API and returns a JWT token.
+        """
+        logger.info("Getting a token")
+        url = f"{self.url}/security/user/authenticate"
+        try:
+            # Wazuh uses Basic Auth to fetch the initial token
+            response = requests.get(
+                url, auth=(self.username, self.password), verify=False, timeout=60
+            )
+            response.raise_for_status()
+            token = response.json().get("data", {}).get("token")
+            logger.info("Token adquired correclty, returning it")
+            return token
+        except requests.exceptions.RequestException as e:
+            logger.critical("Unable to retrieve a valid token: %s", e)
+            raise e
+
+    def _query_agents(self, values=None, query=None):
+        """
+        Queries all agents in Wazuh to get a certain list of parameters
+        for a specific query or filter
+
+        :param values: the list of variables to retrieve
+        :type values: list
+        :param query: the specific query against the Wazuh server
+        :type query: str
+        :return: the information in Wazuh for all agents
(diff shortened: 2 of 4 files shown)


lib/apis/wazuh_api/wazuh.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        gets the requires values to contact a Wazuh Server
+        from the StackStorm configuration Dictionary pack_config
+
+        :param pack_config: StackStorm configuration
+        :type pack_config: dict
+        """
+        self.username = pack_config.get("wazuh_username")
+        self.password = pack_config.get("wazuh_password")
+        self.url = pack_config.get("wazuh_url")
```
gmatthews20, line 25: You should add these to this file: https://github.com/stfc/st2-cloud-pack/blob/main/config.schema.yaml

lib/apis/wazuh_api/wazuh.py
```diff
@@ -0,0 +1,76 @@
...
+            # we query Wazuh using pagination
+            # we do not know how many agents there are
+            all_agents = []
+            batch_size = 500
+            offset = 0
+            while True:
+                params = {
+                    "limit": batch_size,
+                    "offset": offset,
+                    "select": "name,os.name,os.version,status,ip,group",  # Only return fields we care about
```
gmatthews20: Not sure how many properties we can chose from. But maybe this would be passed as a parameter similar to the other query actions with `properties_to_select`?

lib/apis/wazuh_api/wazuh.py
```diff
@@ -0,0 +1,76 @@
...
+                response.raise_for_status()
+                data = response.json()
+                items = data.get("data", {}).get("affected_items", [])
+                if not items:
+                    break
+                all_agents.extend(items)
+                offset += batch_size
+            return all_agents
+        except requests.exceptions.RequestException as e:
+            print(f"Failed to fetch data: {e}")
```
gmatthews20: Is this necessary? Its just printing the original exception?

lib/apis/wazuh_api/wazuh.py
```diff
@@ -0,0 +1,76 @@
...
+        url = f"{self.url}/security/user/authenticate"
+        try:
+            # Wazuh uses Basic Auth to fetch the initial token
+            response = requests.get(
+                url, auth=(self.username, self.password), verify=False, timeout=60
+            )
+            response.raise_for_status()
+            return response.json().get("data", {}).get("token")
+        except requests.exceptions.RequestException as e:
+            print(f"Authentication failed: {e}")
```
gmatthews20: Is this necessary? Its just printing the original exception?

lib/apis/wazuh_api/wazuh.py
```diff
@@ -0,0 +1,76 @@
...
+            return None
+
+    def query_agents(self):
```
gmatthews20: unit test?

actions/src/openstack_actions.py
```diff
@@ -25,6 +32,11 @@ def run(self, lib_entry_point: str, requires_openstack: bool = False, **kwargs):
             "\n".join([f"{key}: {val}" for key, val in kwargs.items()]),
         )
         kwargs = self.parse_configs(**kwargs)
+
+        if requires_wazuh:
+            wazuh = WazuhClient(self.config)
+            kwargs["wazuh"] = wazuh
```
gmatthews20, line 47: Why is this here rather than in `parse_configs`?

actions/src/openstack_actions.py
gmatthews20, line 25: This needs updating with `requires_wazuh`

gmatthews20 (changes requested):

anish-mudaraddi (dismissed):

lib/apis/wazuh_api/wazuh.py
jose-caballero: this will be done separately. It is part of a different JIRA Ticket, assigned to Anish

lib/apis/wazuh_api/wazuh.py
jose-caballero: in this commit I am only trying to get the minimum code that can query Wazuh. More specific queries will be implement later on. 
I am currently testing something like this

```
def get_servers(os_name, os_version):
...
...
    params = {
                    "limit": 500,
                    "offset": offset,
                    "select": "name",
                    "q": f"os.platform={os_name};os.major={os_version};status=active;group!=kolla;group!=quattor",
                }
...
...
```

lib/apis/wazuh_api/wazuh.py
jose-caballero: actually, the main problem was not that, but the fact that when the exception was captured, it returned None.
Now we write a log message with CRITICAL level, and re-raise the exception

actions/src/openstack_actions.py
jose-caballero: Added inline docs to the file explaining it

actions/src/openstack_actions.py
jose-caballero: fixed

lib/apis/wazuh_api/wazuh.py
jose-caballero: same comment as the other one.

lib/apis/wazuh_api/wazuh.py
jose-caballero: maybe later, when we have the final shape of the code. This commit is just a way to start.

gmatthews20: If this is not final then why is it a PR that needs reviewing and merging now?

lib/apis/wazuh_api/wazuh.py
gmatthews20: I believe it should be part of this PR as it is a requirement for this PR to be functional. Getting the account credentials can be done separately

jose-caballero: fair. I have converted this PR to Draft and wait until Anish has done his part.

lib/apis/wazuh_api/wazuh.py
jose-caballero: I will write some unit test

lib/apis/wazuh_api/wazuh.py
```diff
@@ -0,0 +1,145 @@
...
+        logger.info(
+            "query Wazuh for Servers running OS name %s and %s version",
+            os_name,
+            os_version,
+        )
+        query = f"os.platform={os_name};os.major={os_version};status=active;group!=kolla;group!=quattor"
+        # we ensure we only get information about Servers and not Hypervisors
+        # by adding conditions
+        #   group!=kolla
+        #   group!=quattor
```
gmatthews20: Does the quattor group container VMs configured with quattor?

lib/apis/wazuh_api/wazuh.py
```diff
@@ -0,0 +1,145 @@
...
+        # to the query string
+        data = self._query_agents(values=["name"], query=query)
+
+        # extract the Server ID from the fetched data
+        # the variable "name" for OpenStack Servers looks like this:
+        #   host-192-168-231-139.openstacklocal-001682d2-79a2-41b9-af7f-c553f7d67b0d
+        # we need to extract the ID from there
```
gmatthews20, line 127: From the wazuh agent config each openstack VM should be labeled with the `openstack.uuid`. Not sure how to find the agent labels but I think this would be easier and more consistent than a regex on the agent name.

https://github.com/stfc/cloud-image-builders/blob/710988483da6c65cb812041693b2348dc7c2c40f/os_builders/roles/vm_baseline/files/wazuh/enrich-wazuh-agent.py#L69-L76

gmatthews20 (changes requested):

lib/apis/wazuh_api/wazuh.py
khalford: Agreed, I think this regex will miss edge cases

lib/apis/wazuh_api/wazuh.py
jose-caballero: good point. I have removed quattor from the query

lib/apis/wazuh_api/wazuh.py
jose-caballero: line 76 in that snippet is
```
agent_hostname = agent_hostname + "-" + openstack_metadata["uuid"]
```
so all VMs should have the UUID at the end. That's what I am parsing. Right?

